PR #1, opened 2018-10-19
Support composite keyword schema validation

smithsz:
Support use of composite keywords `allOf`, `anyOf` and `oneOf` in schema specifications.

### DIFF
--- a/test/plugins/validation/2and3/schema-ibm.js
+++ b/test/plugins/validation/2and3/schema-ibm.js
@@ -153,6 +153,131 @@ describe('validation plugin - semantic - schema-ibm - Swagger 2', () => {
     expect(res.warnings.length).toEqual(0);
   });
 
+  it('should not complain when using composite keyword "allOf"', () => {
+    const config = {
+      schemas: {
+        invalid_type_format_pair: 'error'
+      }
+    };
+
+    const spec = {
+      definitions: {
+        WordStyle: {
+          description: 'Good to have a description',
+          properties: {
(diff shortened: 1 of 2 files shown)


test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -153,6 +153,131 @@ describe('validation plugin - semantic - schema-ibm - Swagger 2', () => {
...
+                {
+                  type: 'integer',
+                  format: 'int32',
+                  description: 'Good to have a description'
+                },
+                {
+                  type: 'string',
+                  description: 'Good to have a description'
+                }
+              ]
```
mkistler: I don't believe this is a valid use of `allOf` in either OpenAPI 2 or OpenAPI 3.

The [v2 spec says](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism):
> allOf takes in an array of object definitions that are validated independently but together compose a single object.

And the [v3 spec says](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties):
> Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema.

Neither of the elements of the the `allOf` array here are object schema, so this is invalid and the validator should report the appropriate error or warning for this case.

smithsz:
Thanks Mike. It seems our specification is incorrect in its usage of these composite keywords. We'll review these.

PR #9, opened 2018-11-15
Add a "print version" option

dpopp07:
This came up while looking at an issue with Allen - there was no obvious way to see which version he had installed locally. Since most CLI tools use the `-v` option to print the version, I decided to add this.

There was previously a `-v` arg for printing validator file names but that was a debug tool I made for myself and I can't imagine any users would use that, plus we are still in `0.x` so we can make breaking changes 😏 

In this PR...
- The `-v` option is updated to print the version
- The print validator version is changed to `-p`
- The logic for turning off coloring is simplified with a ternary operator
- The README is updated with new option and also correct some outdated statements (the npm `--save` option is now deprecated)
- The appropriate option handler test is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,3 @@
+# 0.2.0
+- The options `-v` or `--version` now prints the semantic version of the validator
+- The `--print_validator_modules` short-hand option is renamed to `-p`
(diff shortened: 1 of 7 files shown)


SirSpidey:
oh sure. blame me.

PR #10, opened 2018-11-28
Add semantic release

dpopp07:
Sets up `semantic-release` for the project.

- Installed `semantic-release` as a dependency
- Added the necessary tokens as environment variables in Travis
- Wrote a `.releaserc` that is essentially the same as in the Node SDK
- Added a deploy script to the `.travis.yml` file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: node_js
 node_js:
 - '8'
-cache:
-  directories:
-  - node_modules
 deploy:
 - provider: script
   skip_cleanup: true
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
...
 - '8'
-cache:
-  directories:
-  - node_modules
-  
\ No newline at end of file
+deploy:
+- provider: script
+  skip_cleanup: true
+  script: npx semantic-release
```
AnthonyOliveri: You should specify the branch this should be run on (master). I think it will always run the deploy script when there's a push to a side branch if you don't.

AnthonyOliveri: Like this: https://github.com/watson-developer-cloud/swift-sdk/blob/master/.travis.yml#L49

PR #15, opened 2019-01-16
fix: handle null values for descriptions and schemas properties

dpopp07:
The validator was crashing on the COS spec due to some properties without values. While these were unintentional, the code should still be able to run fine.

This cleans up the areas blocking a successful run on that specific spec.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/generator": {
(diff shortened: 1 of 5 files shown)


mkistler:
Are properties with "null" schemas acceptable in OpenAPI? If so, this condition should be at least a warning, and if not, there should be an error issued.  Is that the case?

dpopp07:
That's a great question. I was just trying to get the fix out quickly, but we should let the user know if they are adding blank items in their spec. I will look into this and update the PR accordingly.

dpopp07:
@mkistler as far as I can tell, null values are _not_ allowed by the spec. So, I added a non-configurable validation that will return an error if any property in the spec has a null value.

README.md
```diff
@@ -217,6 +217,7 @@ The supported rules are described below:
 | --------------------------- | ----------------------------------------------------------------------------- | -------- |
 | invalid_type_format_pair    | Flag any schema that does not follow the [data type/format rules.][2]         | shared   |
 | snake_case_only             | Flag any property with a `name` that is not lower snake case.                 | shared   |
+| no_schema_description       | Flag any schema without a `description` field.                                | shared   |
```
mkistler: I don't understand what is going on here.  This line is _already_ in `master`, so why is GHE showing this in the diff?

dpopp07:
@mkistler This has been rebased on top of `master` and is ready for review now

PR #24, opened 2019-02-14
fix: dont error on schemas with type file for oas2 specs

dpopp07:
As we learned from [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) recently,
> An additional primitive data type "file" is used by the Parameter Object and the Response Object to set the parameter type or the response as being a file.

This PR addresses the fact that the validator was giving errors for schemas with type file.

It's worth noting that `parameters-ibm` already supported allowing type file if the condition is met that it's a `formData` parameter.

After some consideration, my solution here is to allow type "file" for all schemas. My reasoning behind this is that it would be non-trivial to check whether or not models in `definitions` are being uses for response schemas or not. Additionally, schemas pretty much only appear in parameters or responses - parameters are checked for their condition of using "file" elsewhere and responses can use "file" unconditionally. I don't see too many use cases where type "file" would be explicitly disallowed by the spec that we aren't covering.

I'm happy to revise this approach if it's found objectionable! This seemed to have the best time/value trade-off.

### DIFF
--- a/test/plugins/validation/2and3/schema-ibm.js
+++ b/test/plugins/validation/2and3/schema-ibm.js
@@ -188,6 +188,56 @@ describe('validation plugin - semantic - schema-ibm - Swagger 2', () => {
     expect(res.warnings.length).toEqual(0);
   });
 
+  it('should return an error for a response schema with non-root type file', () => {
+    const config = {
+      schemas: {
+        invalid_type_format_pair: 'error'
+      }
+    };
+
+    const spec = {
+      paths: {
+        '/pets': {
+          get: {
+            responses: {
+              '200': {
+                description: 'legal response',
+                schema: {
+                  properties: {
+                    'this_is_bad': {
+                      type: 'file',
+                      description: "non-root type of file is bad"
+                    }
+                  }
+
+                }
+              }
+            }
+          }
+        }
+      }
+    };
+
+    const res = validate({ jsSpec: spec }, config);
+    expect(res.errors.length).toEqual(1);
+    expect(res.errors[0].path).toEqual([
+      'paths',
+      '/pets',
+      'get',
+      'responses',
+      '200',
+      'schema',
+      'properties',
+      'this_is_bad',
+      'type'
+    ]);
+    expect(res.errors[0].message).toEqual(
+      'Property type+format is not well-defined.'
+    );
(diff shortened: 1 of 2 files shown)


test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -159,6 +159,35 @@ describe('validation plugin - semantic - schema-ibm - Swagger 2', () => {
...
+          }
+        }
+      }
+    };
+
+    const res = validate({ jsSpec: spec }, config);
+    expect(res.errors.length).toEqual(0);
+    expect(res.warnings.length).toEqual(0);
+  });
+
```
mkistler:
```suggestion

  it('should return an error for a response schema with non-root type file', () => {
    const config = {
      schemas: {
        invalid_type_format_pair: 'error'
      }
    };

    const spec = {
      paths: {
        '/pets': {
          get: {
            responses: {
              '200': {
                description: 'legal response',
                schema: {
                  properties: {
                    'this_is_bad': {
                      type: 'file',
                      description: "non-root type of file is bad"
                    }
                  }

                }
              }
            }
          }
        }
      }
    };

    const res = validate({ jsSpec: spec }, config);
    expect(res.errors.length).toEqual(1);
    expect(res.errors[0].path).toEqual([
      'paths',
      '/pets',
      'get',
      'responses',
      '200',
      'schema',
      'properties',
      'this_is_bad',
      'type'
    ]);
    expect(res.errors[0].message).toEqual(
      'Property type+format is not well-defined.'
    );

    expect(res.warnings.length).toEqual(0);
  });

```

dpopp07: I like that suggestion and I don't imagine it should be too difficult. I will give it a shot

dpopp07:
@mkistler I added an extra check to verify the schema is a "root" schema. I tried to think of a simple "proof" for this and I came up with the following:
- schema object is under key `schema`, OR
- schema object is in `definitions` section

all other locations for "schema objects" are in non-root locations (`items`, `properties`, etc.).

The only other location that might be considered a "root level" schema would be an object within an `allOf` array. I decided not to include that, but the code can be easily adapted for that

PR #25, opened 2019-02-18
feat: flag operations with non-form request bodies that do not specify a name

dpopp07:
Phil recommended this validation for the OpenAPI 3 transition. In OAS 3, the request body property of an operation does not have a "name" field. Having unnamed request bodies is not ideal for SDK generation so an annotation was introduced to add a name to the request body: `x-codegen-request-body-name`.

This PR validates that operations set this property and returns a `warning` if they do not. It only does this for non-form parameters, which I check for using a blacklist of mime types. Would love to get some feedback on any mime types I'm missing or if any I am using don't belong.

I also noticed that the other validation in `oas3/operations` was not configurable when it should have been, so that's been fixed as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ The default values for each rule are described below.
 | Rule                        | Default |
 | --------------------------- | --------|
 | no_request_body_content     | error   |
-| no_request_body_name        | error   |
+| no_request_body_name        | warning |
 
 ###### parameters
 | Rule                        | Default |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -307,6 +308,7 @@ The default values for each rule are described below.
 | Rule                        | Default |
 | --------------------------- | --------|
 | no_request_body_content     | error   |
+| no_request_body_name        | error   |
```
mkistler: Might want to make this a `warning`, since it only affects SDK generation.

dpopp07: Whoops, I did make it a warning in the code. Just documented it incorrectly. I'll fix that

dpopp07:
@mkistler Great point, named schemas should do the trick. I'll adjust the code so that a name annotation is not necessary in that case

dpopp07:
Additional info to check for, from @padamstx :
If there is only one content type, that content type is JSON, and the schema is not an array schema, the body will be exploded so there is no need for a name.

dpopp07:
@mkistler Comments addressed, ready for review

PR #27, opened 2019-02-20
flag descriptions sibling to `$ref` if identical to referenced description

dpopp07:
In the ongoing topic of descriptions that are siblings to $refs, we seem to have decided that they are sometimes beneficial and we've set the validation for that check to a default of `off`. That said, @padamstx thought it would be a good idea to let users know when they are using $ref sibling descriptions unnecessarily. This PR adds a validation that checks if $ref sibling descriptions are identical to the description in the schema that they reference (so, redundant).

This wasn't an easy validation to describe, so happy to change the message or the config property name if someone has other ideas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,11 +235,12 @@ The supported rules are described below:
 | invalid_non_empty_security_array | Flag any non-empty security array this is not of type OAuth2 | shared |
 
 ##### walker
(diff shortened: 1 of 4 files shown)


dpopp07:
Updated with the latest from `master`, ready for review

PR #37, opened 2019-05-17
feat: added check in the validator for info section.

SamerJaser96:
Validator will return an error if the info section is missing or filled out incorrectly.


src/plugins/validation/2and3/semantic-validators/info.js
```diff
@@ -0,0 +1,33 @@
+// Assertation 1:
+// check if info exists
+
+// Assertation 2:
+// making sure that the required version and title are defined properly
+module.exports.validate = function({ jsSpec }) {
+  const errors = [];
+  const warnings = [];
+  const no_input = !jsSpec.info;
```
dpopp07: If an api definition has `info: 'abc'`, this validation will pass. We should check that it is an object. Also, whenever I think of edge cases that could've broken my code, I add tests for them. This helps ensure code quality and helps me think of better tests the next time I'm writing them. It would be a good idea to add this test in there 👍

src/.defaultsForValidator.js
```diff
@@ -20,6 +20,10 @@
 
 const defaults = {
   'shared': {
+    'info':{
+      'no_info': 'error',
+      'mistyped_title_or_version': 'error'
+    },
```
dpopp07: Since these are explicitly required by the specification, I don't think these need to be configurable. Therefore, we don't need default values. You can remove this section

README.md
```diff
@@ -328,6 +334,13 @@ The default values for each rule are described below.
 
 ##### shared
 
+
+###### operations
+| Rule                         | Default |
+| ---------------------------- | ------- |
+| no_info                      | error   |
+| mistyped_title_or_version    | error   |
+
```
dpopp07: On second thought, I don't think we need any of this added to the readme. These are baseline spec requirements that don't really need documentation in my opinion. I'll ask Mike about this but for now, you can take this out

src/plugins/validation/2and3/semantic-validators/info.js
```diff
@@ -0,0 +1,33 @@
...
+    errors.push({
+      path: ['info'],
+      message: 'Missing info object for api defintion'
+    });
+  }
+  //Assertion 2
+  if (
+    jsSpec.info &&
+    (typeof jsSpec.info.title != 'string' ||
+      typeof jsSpec.info.version != 'string')
```
dpopp07: I have a few comments about this validation:

In the first check, you are making sure there is an info section that is an object. So, you can wrap the next validations in an `else` block or something and that way you don't have to check for `jsSpec.info` anymore.

To make this more readable, it's good to use variables with names reflecting what you are actually testing for. Also would be good to make sure the strings aren't empty. So something like
```js
const title = jsSpec.info.title;
const hasTitle = typeof title === 'string' && title.toString().trim().length > 0;

if (!hasTitle) { // ...
```

Also, I think you should split these validations up so they each have their own check and message.

src/plugins/validation/2and3/semantic-validators/info.js
```diff
@@ -0,0 +1,33 @@
...
+    mistyped_title_or_version = true;
+  }
+  if (mistyped_title_or_version) {
+    errors.push({
+      path: ['info', 'title'],
+      message:
+        'Info object missing proper definitions for title or version properties'
+    });
+  }
+  return { errors: errors, warnings: warnings };
```
dpopp07: For shorthand, you can just say

```js
return { errors, warnings }
```

test/plugins/validation/2and3/info.js
```diff
@@ -0,0 +1,41 @@
...
+describe('validation plugin - semantic - error handling for incorrect formatting', () => {
+  describe('there should be errors when the defintion is missing neccesrry fields', function() {
+    //this is for openapi object
+    it('should return an error when a parameter does not have info', () => {
+      const spec = {
+        Openapi: '3.0.0'
+      };
+
+      const res = validate({ jsSpec: spec }, config);
+      //expect(res.warnings.length).toEqual(0);
```
dpopp07: You can uncomment these now - looks like they would've caught the bug you had 😉

src/plugins/validation/2and3/semantic-validators/info.js
```diff
@@ -0,0 +1,40 @@
...
+  if (!info) {
+    errors.push({
+      path: ['info'],
+      message: 'Missing info object for api defintion'
+    });
+  } else if (typeof info != 'object') {
+    errors.push({
+      path: ['info'],
+      message: 'Info is an obect and must have properties'
+    });
```
dpopp07: This is looking good. I think it is probably overkill to have both of these as separate checks so let's combine them using something like a `hasInfo` variable:
```js
const hasInfo = info && typeof info === 'object';
```

and then use an error message like "API definition must have an \`info\` object"

src/plugins/validation/2and3/semantic-validators/info.js
```diff
@@ -0,0 +1,40 @@
...
+    const title = jsSpec.info.title;
+    const hasTitle =
+      typeof title === 'string' && title.toString().trim().length > 0;
+    const version = jsSpec.info.version;
+    const hasVersion =
+      typeof version === 'string' && version.toString().trim().length > 0;
+    if (!hasTitle) {
+      errors.push({
+        path: ['info', 'title'],
+        message: ' Info object is missng the title field'
```
dpopp07: Let's use an error message of "\`info\` object must have a string-type \`title\` field"

Most of the error messages have the format of stating the requirement rather than the problem, so I think it's good to try for that when we can. You can use the same message for the version check. Also they should not start with a space

test/plugins/validation/2and3/info.js
```diff
@@ -0,0 +1,98 @@
+const config = require('../../../../src/.defaultsForValidator').defaults.shared;
+const expect = require('expect');
+const {
+  validate
+} = require('../../../../src/plugins/validation/2and3/semantic-validators/info');
+
+describe('validation plugin - semantic - error handling for incorrect formatting', () => {
```
dpopp07: Have this describe message be `validation plugin - semantic - info`. I'm pretty sure we usually use the name of the validator (so the name of the file) for that last bit

test/plugins/validation/2and3/info.js
```diff
@@ -0,0 +1,98 @@
...
+  describe('there should be errors when the defintion is missing neccesrry fields', function() {
```
dpopp07: "Describe" blocks are used to group related tests together. Since all of your tests are dealing with the `info` object in a pretty related way, you don't need to have this describe block - the top level one is enough

test/plugins/validation/2and3/info.js
```diff
@@ -0,0 +1,98 @@
...
+      };
+
+      const res = validate({ jsSpec: spec }, config);
+      expect(res.errors.length).toEqual(1);
+      expect(res.errors[0].path).toEqual(['info']);
+      expect(res.errors[0].message).toEqual(
+        'Info is an obect and must have properties'
+      );
+    });
+  });
```
dpopp07: These tests look great. I would move this last one to be right under the first one since they are pertaining to the same check - it's good to order the tests logically like that

Issue #40, opened 2019-05-29
Add explicit control of case convention for paths and schema


aspear:
Hi!  This tool is awesome, and  it would be really nice to have explicit control over case conventions for paths, and then schema objects and enums.  As it is, the only control over this is the global boolean "snake_case_only" which is only useful for one convention (presumably IBMs).  In my company (VMware) we have many APIs that come from different products that all have their own ways of handing paths for example, one lower_camel_case, another lower_snake_case.

I propose (and have already implemented!) a couple of small backwards compatible additions:
1) in the paths section, add a new config option "paths_case_convention" that allows explict control.  The .defaultsForValidator.js changes to:
'paths': {
      'missing_path_parameter': 'error',
      'snake_case_only': 'warning',
      'paths_case_convention': ['off', 'lower_snake_case']
    },
The idea here is that a user could set snake_case_only to off, and then use paths_case_convention in a similar manner to the existing param_name_case_convention to explicitly control the selection for paths.

2) Add two additional options in the spirit of the change above for the schemas section:
'schemas': {
     ...
      'property_case_convention': [ 'off', 'lower_snake_case'],
      'enum_case_convention': [ 'off', 'lower_snake_case']
    },
This allows you to have one convention for properties of objects (e.g. lower_snake_case), and then a separate convention for enums (e.g. upper_snake_case, 'MY_ENUM_VALUE')

This  fixes #38 btw.

3) Add additional case convention values upper_snake_case and upper_dash_case (note that my company uses upper_snake_case for enums.  We don't use upper_dash_case, but while I was touching the code I figured I might as well add it so that all cases were covered)

I already have this implemented in a fork.  I am implementing tests right now.  I would be very happy to submit a pull request for it if you are open.

thanks,
Aaron Spear, VMware

dpopp07:
@aspear Thanks for the issue, so glad you like the tool! Making all of the current case convention validations configurable is something we really want to do and it's been sitting in our backlog for a while - we just haven't gotten around to it. We definitely welcome that PR, I will review it soon

PR #41, opened 2019-05-29
feat: added explicit control of case conventions in paths, properties and enums

### LINKED ISSUE #38: Allow having lower_camel_case under schema validation
for all of the camel case lovers :)

aspear:
This change adds support for a couple of new case conventions to round out case convention
support, upper_snake_case and upper_dash_case.  Additionally it includes support for explicit
and independent control of the case convention for paths (paths_case_convention), schema
properties (property_case_convention), and schema enums (enum_case_convention).  These options
are disabled by default so as not to impact current usage of snake_case_only and thus are
backwards compatible. In order to use them, a user must turn off the corresponding
'snake_case_only' option and enable these options.  Full automated tests are included.

Closes #38, #40

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/paths-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/paths-ibm.js
@@ -135,28 +135,30 @@ module.exports.validate = function({ resolvedSpec }, config) {
           });
         }
       });
-    }
-
-    // enforce path segments follow path_case_convention if provided
-    if (config.paths_case_convention) {
-      const checkStatusPath = config.paths_case_convention[0];
-      if (checkStatusPath !== 'off') {
-        const caseConvention = config.paths_case_convention[1];
-        const segments = pathName.split('/');
-        segments.forEach(segment => {
-          // the first element will be "" since pathName starts with "/"
-          // also, ignore validating the path parameters
-          if (segment === '' || segment[0] === '{') {
-            return;
-          }
-          const isCorrectCase = checkCase(segment, caseConvention);
-          if (!isCorrectCase) {
-            result[checkStatusPath].push({
-              path: `paths.${pathName}`,
-              message: `Path segments must follow case convention: ${caseConvention}`
-            });
-          }
-        });
+    } else {
+      // in the else block because usage of paths_case_convention is mutually
+      // exclusive with usage of config.snake_case_only since it is overlapping
+      // functionality
+      if (config.paths_case_convention) {
+        const checkStatusPath = config.paths_case_convention[0];
+        if (checkStatusPath !== 'off') {
+          const caseConvention = config.paths_case_convention[1];
+          const segments = pathName.split('/');
+          segments.forEach(segment => {
+            // the first element will be "" since pathName starts with "/"
+            // also, ignore validating the path parameters
+            if (segment === '' || segment[0] === '{') {
+              return;
+            }
+            const isCorrectCase = checkCase(segment, caseConvention);
+            if (!isCorrectCase) {
+              result[checkStatusPath].push({
+                path: `paths.${pathName}`,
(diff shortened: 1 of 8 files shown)


aspear:
BTW, I just realized that I forgot to update the documentation in the README.md file.  I will up the pull request with this change today.

test/plugins/validation/2and3/paths-ibm.js
```diff
@@ -271,7 +271,7 @@ describe('validation plugin - semantic - paths-ibm', function() {
     expect(res.warnings).toEqual([]);
   });
 
-  it('shoud flag a path segment that is not snake_case but should ignore path parameter', function() {
+  it('should flag a path segment that is not snake_case but should ignore path parameter', function() {
```
dpopp07: Good catch!

dpopp07: Good catch!

src/plugins/validation/2and3/semantic-validators/paths-ibm.js
```diff
@@ -134,6 +135,31 @@ module.exports.validate = function({ resolvedSpec }, config) {
           });
         }
       });
+    } else {
+      // in the else block because usage of paths_case_convention is mutually
+      // exclusive with usage of config.snake_case_only since it is overlapping
+      // functionality
+      if (config.paths_case_convention) {
```
dpopp07: This `if` statement is not really necessary. It seems weird to not check if a property is defined before referencing it, but [there is code](https://github.com/IBM/openapi-validator/blob/8152d06e184963abe101702dd69250735dfa9b90/src/cli-validator/utils/processConfiguration.js#L123-L146) that guarantees all configuration options are set by the time it reaches the individual validations. So, this will always be true.

If you prefer to leave it in, that's fine with me but I think it's a somewhat redundant check.

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -96,6 +97,34 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
       res = checkEnumValues(schema, path, config);
       errors.push(...res.error);
       warnings.push(...res.warning);
+    } else {
+      // optional support for property_case_convention and enum_case_convention
+      // in config.  In the else block because support should be mutually exclusive
+      // with config.snake_case_only since it is overlapping functionality
+      if (config.property_case_convention) {
```
dpopp07: See earlier comment

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -96,6 +97,34 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+          res = checkPropNamesCaseConvention(
+            schema,
+            path,
+            config.property_case_convention
+          );
+          errors.push(...res.error);
+          warnings.push(...res.warning);
+        }
+      }
+      if (config.enum_case_convention) {
```
dpopp07: See earlier comment

src/plugins/validation/2and3/semantic-validators/paths-ibm.js
aspear: Originally I did not have it in, but during the automated tests there were crashes because this was not defined.  My guess is that the test harness somehow bypasses this check you mentioned?

dpopp07: Ah, that is correct. The test framework runs validations individually and never touches the shared utility code. I've been slowly updating the tests to be less fragile with respect to the config object by bringing in the entire "defaults" object and using that, rather than defining a config object for each each individual test. I just hadn't gotten around to the `paths-ibm` test yet. I'll merge this as is and fix that at a later time

PR #50, opened 2019-05-31
feat: Added --config option, adding support for explicit config file

### LINKED ISSUE #49: Ability to manually specify a path to a configuration file
Right now the default behavior of lint-openapi is to search in the working directory for .validaterc and then search in parent folders.  It would be convenient for automation use cases to be able to specify a particular config file explicitly, e.g. 

 ```bash
lint-openapi --config=/path/to/my-config.json /path/to/swagger.json
```

In this use case .validaterc would be completely ignored and only the specified file would be loaded.

I have a PR ready to go with this implemented, but am having trouble with getting the tests to work.  There seems to be an issue with the config being loaded in the context of the cli tests...

### PR DESCRIPTION
**Pull request by aspear**
This feature adds a -c/--config option which is a path to a config file (.validaterc).
This allows usage of configuration that are not dependent on the .validaterc being in
the same folder (or a parent) of the current working directory.

Note that the automated test added for this doesn't run currently.  For the test I am
changing values for the defaults (a warning to error and other warning to off) so that
the reported errors in the justWarn.yaml file are changed from stock behavior.  Manually
executing the commands from the command line works as expected, but it seems that in the
context of the cli tests that the config seems to be totally ignored and the default is
all that is used.  I have not been able to figure this out yet... Any pointers as to why
this is and how to fix it are appreciated.

Closes #49

### CODE REVIEW

**Comment by aspear** [comment]
@dpopp07 do you have any idea how I can pass a config from an external file in the context of the automated tests?  I have been trying to figure out how to do this and am a bit stumped currently.  It seems that only the default config values are used in the context of the cli tests, and I don't know how to get it to actually use and load the config values from an external file.

**Comment by dpopp07** [comment]
Hey @aspear I was just looking into that. I haven't tested it but my first instinct is that this line is your problem -> `program.default_mode = true;`

Setting `default_mode` to true would override any config options that are loaded. That variable is optional, so if you remove that line, does it work? Line 257 in the test file

File: test/cli-validator/tests/optionHandling.js [hunk 1]
```diff
@@ -222,4 +222,70 @@ describe('cli tool - test option handling', function() {
...
+    const capturedText = [];
+
+    const unhookIntercept = intercept(function(txt) {
+      capturedText.push(stripAnsiFrom(txt));
+      return '';
+    });
+
+    const program = {};
+    program.args = ['./test/cli-validator/mockFiles/justWarn.yml'];
+    program.default_mode = true;
```
dpopp07: As I mentioned before, `default_mode` should not be set to `true` when you are intending to use a configuration. This line should be removed

test/cli-validator/mockFiles/justWarnConfigOverride.json
```diff
@@ -0,0 +1,10 @@
+{
+  "shared": {
+    "operations": {
+      "no_operation_id": "off"
+    },
+    "schemas": {
+      "no_schema_description": "error"
+    }
+  }
+}
```
dpopp07: Is this file going to be used? If not, will you remove it?

test/cli-validator/tests/optionHandling.js
```diff
@@ -222,4 +222,70 @@ describe('cli tool - test option handling', function() {
...
+        inErrorBlock = false;
+        inWarningBlock = true;
+      } else if (line.includes('Message')) {
+        if (inErrorBlock) {
+          errorCount++;
+        } else if (inWarningBlock) {
+          warningCount++;
+        }
+      }
+    });
```
dpopp07: This is a nifty way of counting the number of errors/warnings in the captured output - nice!

src/cli-validator/utils/processConfiguration.js
```diff
@@ -168,11 +168,19 @@ const validateConfigObject = function(configObject, chalk) {
...
+  let configFile;
+
+  if (configFileOverride) {
+    // user explicitly provided a config file, use it instead of .validaterc
+    configFile = configFileOverride;
+  } else {
+    // search up the file system for the first instance
+    // of the config file
+    configFile = await findUp('.validaterc');
+  }
```
dpopp07: Might I suggest a slightly different implementation here? My concern is the error handling if the file at the specific path doesn't exist.

Although it's not exactly the intended function, `find-up` can be used to determine existence of a file if given a full path (and it won't look anywhere else). I propose something like this:

```js
  let configObject;

  // if user explicitly provided a config file, use it instead of .validaterc
  const configFileName = configFileOverride || '.validaterc';

  // search up the file system for the first instance
  // of '.validaterc' or,
  // if a config file override is passed in, use find-up
  // to verify existence of the file
  const configFile = await findUp(configFileName);

  // if the user does not have a config file, run in default mode and warn them
  // (findUp returns null if it does not find a file)
  if (configFile === null && !defaultMode) {
    if (configFileOverride) {
      console.log(
        '\n' +
          chalk.yellow('[Warning]') +
          ` No file found at specified path: ${chalk.underline(
            `${configFileOverride}`
          )}. The validator will run in ` +
          chalk.bold.cyan('default mode.')
      );
    } else {
      console.log(
        '\n' +
          chalk.yellow('[Warning]') +
          ` No ${chalk.underline(
            '.validaterc'
          )} file found. The validator will run in ` +
          chalk.bold.cyan('default mode.')
      );
      console.log(`To configure the validator, create a .validaterc file.`);
    }
    defaultMode = true;
  }
```

That keeps the code very similar, adds an extra layer of protection against a missing file, and (arguably most importantly) makes the error messages specific to the problem. Perhaps there's a better way to write it or better comments so feel free to make changes but that's the pattern I would prefer. Let me know what you think!

aspear:
The default_mode being in there is a leftover during experimentation.  I was actually trying to see if I could set things up so that the it would always load the defaults and then load whatever you provided for a config file over the top of it.  However, I punted on this idea as I had too many variables and could not figure out how to get it to load any config file during the tests.  I will remove this line.

dpopp07:
Just saw your last comment, that makes sense. The code actually takes sort of the opposite approach - it reads in the config options and then fills in any missing options with the defaults.

aspear:
I am not sure what happened, but I just set the tests up like I expected them to work and after removing the default_mode=true, everything is now working correctly in the tests (maybe I was up too late when doing that)

test/cli-validator/tests/optionHandling.js
aspear: removed.

test/cli-validator/mockFiles/justWarnConfigOverride.json
aspear: Yes, I can remove it.  I wanted to provide only the options different than the defaults to eliminate having to maintain it if other options are added in the future.  I think this is the right behavior, but during test time it seems that this loading is not really working and that we have to provide all of the options.  In any case, I will remove this partial one for new and add it back if we fix the behavior of how the loading is done so it works during test.

src/cli-validator/utils/processConfiguration.js
aspear: I like your implementation, will change to that.

test/cli-validator/tests/optionHandling.js
aspear: does it make sense to put it somewhere shared for other tests to leverage?

test/cli-validator/mockFiles/justWarnConfigOverride.json
aspear: It turns out that this does in fact work, so I removed the justWarnConfigOverrideFull.json file instead and changed the test to use the partial one.

aspear:
Hey @dpopp07 , I was not sure if you would prefer that this change is rebased into one commit since I had to iterate on it?  (I suppose you can squash commits when merging, but perhaps that messes up the commit message format?)  I am happy to follow whatever the proper convention is.

test/cli-validator/mockFiles/justWarnConfigOverride.json
dpopp07: Cool, I thought that should work. That looks good 👍

test/cli-validator/tests/optionHandling.js
dpopp07: I was thinking the exact same thing but I think we can wait until it makes sense to add it to another test

PR #56, opened 2019-06-04
fix: issue 55 problems when filename with path is passed to --config

### LINKED ISSUE #55: Problem finding configuration file passed in explicitly
It appears that the implementation of findUp that is used to find/validate the configuration file cannot be passed path information as a part of the file name.  Doing so results in redudantly prepending the path to the returned file path.  As a result the --config option does not work as implemented unless the path passed in only has a file name and is located in the current directory.

PR with fix coming.

aspear:
It turns out that the findUp method cannot handle a full path to a file as input. When this
is passed in, it redudantly or completely incorrectly prepends the current working directory
to the front of the returned path.  Work around is to split any path information off of the
input path and pass the path portion into the method as opt.cwd.

Closes #55

### DIFF
--- a/src/cli-validator/utils/processConfiguration.js
+++ b/src/cli-validator/utils/processConfiguration.js
@@ -171,8 +171,17 @@ const validateConfigObject = function(configObject, chalk) {
 const getConfigObject = async function(defaultMode, chalk, configFileOverride) {
   let configObject;
 
-  // if user explicitly provided a config file, use it instead of .validaterc
-  const configFileName = configFileOverride || '.validaterc';
+  const findUpOpts = {};
+  let configFileName;
+
+  // You cannot pass a full path into findUp as a file name, you must split the
+  // path or else findUp redudantly prepends the path to the result.
+  if (configFileOverride) {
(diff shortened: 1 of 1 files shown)


src/cli-validator/utils/processConfiguration.js
```diff
@@ -171,8 +171,17 @@ const validateConfigObject = function(configObject, chalk) {
...
-  // if user explicitly provided a config file, use it instead of .validaterc
-  const configFileName = configFileOverride || '.validaterc';
+  const findUpOpts = {};
+  let configFileName;
+
+  // You cannot pass a full path into findUp as a file name, you must split the
+  // path or else findUp redudantly prepends the path to the result.
+  if (configFileOverride) {
+    configFileName = path.basename(configFileOverride);
+    findUpOpts.cwd = path.dirname(configFileOverride);
```
dpopp07: `findUpOpts` isn't used yet, needs to be passed into the function call:
```js
const configFile = await findUp(configFileName, findUpOpts);
```

dpopp07:
Also, it might be a good idea to add a test that covers this use case

PR #58, opened 2019-06-07
feat: validation for security definitions for OAS3

SamerJaser96:
validation for security definitions in OAS3


src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,191 @@
...
+            });
+          } else if (
+            type === OAUTH2 &&
+            flows &&
+            flows.implicit &&
+            !flows.implicit.scopes
+          ) {
+            errors.push({
+              message:
+                "'scopes' is required property type object. The available scopes for the OAuth2 security scheme.",
```
mkistler: This message is confusing.  I think it should be reworded along the lines of the other messages, e.g.:
> oauth2 authorization implicit flow must have required 'scopes' property.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,191 @@
...
+        } else if (flows && flows.implicit) {
+          const authorizationUrl = flows.implicit.authorizationUrl;
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorization implicit flow must have required 'authorizationUrl' parameter if type is `implicit` or `authorizationCode`.", /////////exepected
+              path,
+              authId: key
+            });
+          } else if (
```
mkistler: I think we want this to be an independent if check -- not in the else of the previous.  We want to catch and report all the problems in the document.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+              authId: key
+            });
+          }
+        }
+      } else if (type === HTTP) {
+        //scheme is required
+        if (!security.scheme) {
+          errors.push({
+            message:
+              'scheme must be defined for type `http` and must be a string',
```
dpopp07: I don't think it's necessary to say "must be a string" but I know the spec requires a string so if you really want to enforce it, make sure you check for it. This applies to the other checks too

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+            path,
+            authId: key
+          });
+        }
+      } else if (type == OPENID_CONNECT) {
+        const openIdConnectURL = security.openIdConnectUrl;
+        if (
+          !openIdConnectURL ||
+          typeof openIdConnectURL != 'string' ||
```
dpopp07: Use `!==` here, otherwise this is a good check 👍

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+            path,
+            authId: key
+          });
+        }
+      }
+    }
+    if (
+      flows === AUTHORIZATION_CODE ||
+      flows === PASSWORD ||
+      flows === CLIENT_CREDENTIALS
```
dpopp07: I don't think this "flows" check should be at the top level when it only applies to one type of scheme

test/plugins/validation/oas3/security-definitions-ibm.js
```diff
@@ -0,0 +1,235 @@
...
+    );
+  });
+  it('type can only be `apiKey`, `http`, `oauth2`, `openIdConnect`', function() {
+    const spec = {
+      components: {
+        securitySchemes: {
+          SecuritySchemeModel: {
+            type: 'wrong type',
+            scheme: 'basic',
+            descriptions: 'example text'
```
dpopp07: Just for clarity, this should be `description`, as `descriptions` is not an actual field in the spec

test/plugins/validation/oas3/security-definitions-ibm.js
```diff
@@ -0,0 +1,235 @@
...
+    const res = validate({ resolvedSpec: spec });
+    expect(res.errors.length).toEqual(1);
+    expect(res.warnings.length).toEqual(0);
+    expect(res.errors[0].message).toEqual(
+      '`type` must have one of the following types: `apiKey`, `oauth2`, `http`, `openIdConnect`'
+    );
+  });
+});
+
+describe('if the type is `apiKey` then it should have `query`, `header` or `cookie` as well as `name`', function() {
```
dpopp07: I see what you're doing here, but the `describe` text should provide more of a category. Something like `api key` for rules pertaining to the `apiKey` type would be sufficient. Then the `it` blocks describe what the code should and shouldn't do

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// Assertation 1: `type`` is a necessary property and has four possible values: `apiKey`, `HTTP`, `oauth2`, `openIdConnect`
+// Assertation 2: name property is required for `apiKey` type
+// Assertation 3: `in` property is required for `apiKey` type, valid values are: `query`, `header`` or `cookie`
+// Assertation 4: `scheme` property` is required for `http` type
+// Assertation 5: `flows` object is required for `oauth2`` type
+// Assertation 6: `opedIdConnectUrl` property is required for `openIdConnect` and must be a valid url
```
dpopp07: Will you clean up the back-ticks here?
- `type``
- `header``
- `oauth2``

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+          });
+        } else if (
+          !flows.implicit &&
+          !flows.authorizationCode &&
+          !flows.password &&
+          !flows.clientCredentials
+        ) {
+          errors.push({
+            message:
+              "oauth2 authorization must have required 'flows'. Valid values are 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
```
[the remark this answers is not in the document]
SamerJaser96: I used 'values' because that is how they were referred to in swagger2 security-definitions, I have changed it now to say 'parameters'

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+        if (!flows) {
+          errors.push({
+            message:
+              "oauth2 authorization must have required 'flows' parameter",
+            path
+          });
+        } else if (
+          flows === AUTHORIZATION_CODE ||
+          flows === PASSWORD ||
+          flows === CLIENT_CREDENTIALS
```
mkistler: I don't think these checks work. The value of `flows` should be an [oauth-flows-object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#oauth-flows-object), and these checks are looking for flows to have a string value.  Instead, I think you want to check if `flows` _contains_ any of those keys. So I think you want a check similar to the one below (line 97):
```
       } else if (
           flows.authorizationCode ||
           flows.password ||
           flows.clientCredentials
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+          }
+        } else if (
+          !flows.implicit &&
+          !flows.authorizationCode &&
+          !flows.password &&
+          !flows.clientCredentials
+        ) {
+          errors.push({
+            message:
+              "oauth2 authorization `flows` must have one of the following paramaters: 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
```
mkistler: "parameters" -> "properties"

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+          const authorizationUrl = flows.authorizationUrl;
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorization implicit flow must have required 'authorizationUrl' parameter if type is `implicit` or `authorizationCode`.",
+              path
+            });
+          }
+        } else if (flows.password) {
+          const tokenURL = flows.tokenURL;
```
mkistler: The `tokenURL` property actually lives _inside_ the `flows.password` object ... not directly in `flows`.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+            message:
+              "apiKey authorization must have required 'name' string property. The name of the header or query parameter to be used.",
+            path
+          });
+        }
+      }
+      // oauth2 validation
+      else if (type === OAUTH2) {
+        const flows = security.flows;
+        const tokenUrl = security.tokenUrl;
```
mkistler: I don't think the [Security Scheme Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-scheme-object) has a `tokenUrl` property.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+  const securitySchemes =
+    resolvedSpec.components && resolvedSpec.components.securitySchemes;
+
+  for (const key in securitySchemes) {
+    const path = `securitySchemes.${key}`;
+    const security = securitySchemes[key];
+    const type = security.type;
+
+    if (!type) {
+      errors.push({
```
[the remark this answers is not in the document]
dpopp07: This was conscious on my part. My reasoning is that all of these rules constitute an invalid spec so there is no need to be opinionated. Everything is marked in the spec using the keywords "REQUIRED" or "MUST". It seemed like more work / clutter in the config file than was necessary.

That said, I am open to making these configurable if you think that's a better approach.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,175 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// Assertation 1: `type` is a necessary property and has four possible values: `apiKey`, `HTTP`, `oauth2`, `openIdConnect`
+// Assertation 2: name property is required for `apiKey` type
+// Assertation 3: `in` property is required for `apiKey` type, valid values are: `query`, `header` or `cookie`
+// Assertation 4: `scheme` property` is required for `http` type
+// Assertation 5: `flows` object is required for `oauth2` type
+// Assertation 6: `opedIdConnectUrl` property is required for `openIdConnect` and must be a valid url
+
```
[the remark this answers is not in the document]
dpopp07: This also was intentional, I think. Currently, we only document "named" (aka configurable) rules in the README - many of the legacy, non-configurable rules from Swagger Editor are not documented. The reason being that they are inherent to the spec and are sort of treated as "assumed checks". I am definitely open to documenting these though

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
dpopp07: Correct. The [OAuth Flow Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#oauthFlowObject) does

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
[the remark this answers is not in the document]
mkistler: That's a reasonable explanation. It will be interesting to see how may API docs are caught by this. But I agree ... if the spec says it's required, we should throw an error.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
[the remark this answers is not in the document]
mkistler: I'm just thinking that it would be nice to point to specific section of the spec that motivated a particular check / error message. We don't need to do that in this PR, but as a service to our users I think it would be a good practice to adopt.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+    if (!type) {
+      errors.push({
+        message: 'security scheme is missing required field `type`',
+        path
+      });
+    } else if (authTypes.indexOf(type) === -1) {
+      errors.push({
+        message:
+          '`type` must have one of the following types: `apiKey`, `oauth2`, `http`, `openIdConnect`',
+        path
```
mkistler: I think the `path` should be as specific as possible, so in this case we should point directly at the `type` property within the securityScheme.
```suggestion
        path+'.type'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+      });
+    } else {
+      //apiKey validation
+      if (type === API_KEY) {
+        const authIn = security.in;
+        if (!authIn || !['query', 'header', 'cookie'].includes(authIn)) {
+          errors.push({
+            message:
+              "apiKey authorization must have required 'in' property, valid values are 'query' or 'header' or 'cookie'.",
+            path
```
mkistler:
```suggestion
            path+'.in'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+        if (!flows) {
+          errors.push({
+            message: "oauth2 authorization must have required 'flows' property",
+            path
+          });
+        } else if (
+          (flows.authorizationCode && !flows.authorizationCode.tokenUrl) ||
+          (flows.password && !flows.password.tokenUrl) ||
+          (flows.clientCredentials && !flows.clientCredentials.tokenUrl)
+        ) {
```
mkistler: I think structuring the code this way is not really what we want. I think its quite possible that an OAUTH security scheme could have a `flows` object that defines all three of `authorizationCode`, `password`, and `clientCredentials` flows. In that case, we want to validate that the `tokenUrl` is present in each one of them and issue a warning message for each case where it is not present. The current code will only issue a single message that `tokenURL` is missing without being specific to which one is in error.

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+        } else if (
+          !flows.implicit &&
+          !flows.authorizationCode &&
+          !flows.password &&
+          !flows.clientCredentials
+        ) {
+          errors.push({
+            message:
+              "oauth2 authorization `flows` must have one of the following properties: 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
+            path
```
mkistler:
```suggestion
            path+'.flows'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+          });
+        } else if (flows.implicit) {
+          const authorizationUrl = flows.implicit.authorizationUrl;
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorizationCode flow must have required 'authorizationUrl' property if type is `implicit`",
+              path
```
mkistler:
```suggestion
              path+'.flows.implicit'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+            });
+          } else if (!flows.implicit.scopes) {
+            errors.push({
+              message:
+                "oauth2 authorization implicit flow must have required 'scopes' property.",
+              path
```
mkistler:
```suggestion
              path+'.flows.implicit'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+            });
+          }
+        } else if (flows.authorizationCode) {
+          const authorizationUrl = flows.authorizationUrl;
```
mkistler:
```suggestion
          const authorizationUrl = flows.authorizationCode.authorizationUrl;
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorization implicit flow must have required 'authorizationUrl' property if type is `implicit` or `authorizationCode`.",
```
mkistler: You are checking the "authorizationCode" flow here, so 
```suggestion
                "oauth2 authorizationCode flow must have required 'authorizationUrl' property.",
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+              path
```
mkistler:
```suggestion
              path+'.flows.authorizationCode'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+            });
+          }
+        } else if (flows.password) {
+          const tokenURL = flows.password.tokenURL;
+          if (!tokenURL) {
+            errors.push({
+              message:
+                "oauth2 authorization password flow must have required 'tokenUrl' property.",
+              path
```
mkistler:
```suggestion
              path+'.flows.password'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,168 @@
...
+            });
+          }
+        } else if (flows.clientCredentials) {
+          if (!flows.clientCredentials.tokenUrl) {
+            errors.push({
+              message:
+                "oauth2 authorization clientCredentials flow must have required 'tokenUrl' property.",
+              path
```
mkistler:
```suggestion
              path+'.flows.clientCredentials'
```

dpopp07:
@mkistler Somehow I lost the ability to suggest changes some time ago and haven't seen the option since. Very frustrating because it is a great feature!

mkistler:
@dpopp07 I think it currently only works on public GithHub -- not yet available in IBM GHE.

dpopp07:
Does not work for me on public unfortunately 😕

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,182 @@
...
+          });
+        } else if (flows.implicit) {
+          const authorizationUrl = flows.implicit.authorizationUrl;
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorizationCode flow must have required 'authorizationUrl' property if type is `implicit`",
+              path: path + '.flows.implicit'
+            });
+          } else if (!flows.implicit.scopes) {
```
dpopp07: I think this may be incorrect. If the flow does not have an `authorizationUrl`, then the code will not check for the existence of `scopes`. I believe both are required and are not dependent on each other, right?

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,182 @@
...
+                "oauth2 authorization implicit flow must have required 'scopes' property.",
+              path
+            });
+          }
+        } else if (flows.authorizationCode) {
+          const authorizationUrl = flows.authorizationCode.authorizationUrl;
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorizationCode flow must have required 'authorizationUrl' property if type is `implicit` or `authorizationCode`.",
```
dpopp07: Since you are using a separate check here than the one above for `implicit`, remove mention of `implicit` from this message

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,182 @@
...
+ * limitations under the License.
+ */
+// Assertation 1: `type` is a necessary property and has four possible values: `apiKey`, `HTTP`, `oauth2`, `openIdConnect`
+// Assertation 2: name property is required for `apiKey` type
+// Assertation 3: `in` property is required for `apiKey` type, valid values are: `query`, `header` or `cookie`
+// Assertation 4: `scheme` property` is required for `http` type
+// Assertation 5: `flows` object is required for `oauth2` type
+// Assertation 6: `opedIdConnectUrl` property is required for `openIdConnect` and must be a valid url
+
+const validator = require('validator');
```
dpopp07: Will you rename this variable to `stringValidator`, just for clarity in the context of this tool being called a "validator"?

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,182 @@
...
+      });
+    } else if (authTypes.indexOf(type) === -1) {
+      errors.push({
+        message:
+          '`type` must have one of the following types: `apiKey`, `oauth2`, `http`, `openIdConnect`',
+        path: path + '.type'
+      });
+    } else {
+      //apiKey validation
+      if (type === API_KEY) {
```
dpopp07: I think you can keep using `if else (type === ...)` for the types. It's logically equivalent to what you're doing now and avoids the unnecessary nesting level

test/plugins/validation/oas3/security-definitions-ibm.js
```diff
@@ -0,0 +1,235 @@
...
+        }
+      }
+    };
+
+    const res = validate({ resolvedSpec: spec });
+    expect(res.errors.length).toEqual(1);
+    expect(res.warnings.length).toEqual(0);
+    expect(res.errors[0].message).toEqual(
+      'security scheme is missing required field `type`'
+    );
```
dpopp07: I apologize that I'm just catching this now... will you add assertions for the path returned for all of your tests? For example,
```js
expect(res.errors[0].path).toEqual('whatever.path)
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,178 @@
...
+            "oauth2 authorization `flows` must have one of the following properties: 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
+          path: path + '.flows.implicit'
+        });
+      } else if (flows.implicit) {
+        const authorizationUrl = flows.implicit.authorizationUrl;
+        if (!flows.implicit.scopes || !authorizationUrl) {
+          if (!authorizationUrl) {
+            errors.push({
+              message:
+                "oauth2 authorizationCode flow must have required 'authorizationUrl' property if type is `implicit`",
```
mkistler:
```suggestion
                "oauth2 implicit flow must have required 'authorizationUrl' property",
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,178 @@
...
+              path: path + '.flows.implicit'
+            });
+          }
+        }
+      } else if (flows.authorizationCode) {
+        const authorizationUrl = flows.authorizationCode.authorizationUrl;
+        if (!authorizationUrl) {
+          errors.push({
+            message:
+              "oauth2 authorizationCode flow must have required 'authorizationUrl' property if type is `authorizationCode`.",
```
mkistler:
```suggestion
              "oauth2 authorizationCode flow must have required 'authorizationUrl' property.",
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,178 @@
...
+      } else if (
+        !flows.implicit &&
+        !flows.authorizationCode &&
+        !flows.password &&
+        !flows.clientCredentials
+      ) {
+        errors.push({
+          message:
+            "oauth2 authorization `flows` must have one of the following properties: 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
+          path: path + '.flows.implicit'
```
mkistler:
```suggestion
          path: path + '.flows'
```

src/plugins/validation/oas3/semantic-validators/security-definitions-ibm.js
```diff
@@ -0,0 +1,178 @@
...
+        !flows.clientCredentials
+      ) {
+        errors.push({
+          message:
+            "oauth2 authorization `flows` must have one of the following properties: 'implicit', 'password', 'clientCredentials' or 'authorizationCode'",
+          path: path + '.flows'
+        });
+      } else if (flows.implicit) {
+        const authorizationUrl = flows.implicit.authorizationUrl;
+        if (!flows.implicit.scopes || !authorizationUrl) {
```
dpopp07: You don't need this `if` statement. You are checking for `if (!authorizationUrl)` and `if (!flows.implicit.scopes)` below, there is no need to wrap that in a check for either of them

test/plugins/validation/oas3/security-definitions-ibm.js
```diff
@@ -0,0 +1,255 @@
...
+    expect(res.errors.length).toEqual(1);
+    expect(res.warnings.length).toEqual(0);
+    expect(res.errors[0].path).toEqual(
+      'securitySchemes.SecuritySchemeModel.flows.implicit'
+    );
+    expect(res.errors[0].message).toEqual(
+      "oauth2 implicit flow must have required 'authorizationUrl' property"
+    );
+  });
+  it('should have `authorizationUrl` and `tokenUrl` if type is `oauth2` and flow is `authorizationCode` or `password`', function() {
```
dpopp07:
```suggestion
  it('should have `authorizationUrl` and `tokenUrl` if type is `oauth2` and flow is `authorizationCode`', function() {
```

PR #60, opened 2019-06-12
fix: raises warning for tag that hasnt been defined in tags

SamerJaser96:
if there is a tag in operation tags that isn't in the top level tags then a warning will be raised

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/operations-shared.js
+++ b/src/plugins/validation/2and3/semantic-validators/operations-shared.js
@@ -132,7 +132,7 @@ module.exports.validate = function({ resolvedSpec, isOAS3 }, config) {
             if (checkStatus !== 'off') {
               result[checkStatus].push({
                 path: `paths.${pathKey}.${opKey}.tags`,
-                message: 'tag is not defined in operations'
+                message: 'tag is not defined at the global level: ' + op.tags[i]
               });
             }
           }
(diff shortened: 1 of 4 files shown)


src/plugins/validation/2and3/semantic-validators/operations-shared.js
```diff
@@ -119,6 +119,25 @@ module.exports.validate = function({ resolvedSpec, isOAS3 }, config) {
...
+        for (let i = 0; i < resolvedSpec.tags.length; i++) {
+          resolvedTags.push(resolvedSpec.tags[i].name);
+        }
+        for (let i = 0, len = op.tags.length; i < len; i++) {
+          if (!resolvedTags.includes(op.tags[i])) {
+            const checkStatus = config.unused_tag;
+            if (checkStatus !== 'off') {
+              result[checkStatus].push({
+                path: `paths.${pathKey}.${opKey}.tags`,
+                message: 'tag is not defined in operations'
```
dpopp07: This is a case where I think it would be helpful to print the name of the unused tag in the message. Also, the tag is defined in operations but not the global level. So, the message is a little unclear. Something like this would be better:
```js
'tag is not defined at the global level: `op.tags[i]`'
```

test/plugins/validation/2and3/operations-shared.js
```diff
@@ -715,5 +715,56 @@ describe('validation plugin - semantic - operations-shared', function() {
...
+              summary: 'get everything as a string or an array',
+              responses: {
+                '200': {
+                  content: {
+                    'text/plain': {
+                      schema: {
+                        type: 'string'
+                      }
+                    },
+                    'application/json': {
```
dpopp07: Will you simplify this responses object? No reason to have that in the test for tags and an array response should return an error, it only isn't because type is an empty string for some reason.

`summary` should be changed to match

PR #61, opened 2019-06-14
fix: skips warning for non-ref response schemas for primitive or non-…

SamerJaser96:
the validation will go through for media types starting with 'application/json'

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/responses.js
+++ b/src/plugins/validation/2and3/semantic-validators/responses.js
@@ -19,11 +19,11 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
       each(obj, (response, responseKey) => {
         if (isOAS3) {
           each(response.content, (mediaType, mediaTypeKey) => {
-            const hasInlineSchema =
-              mediaType.schema &&
-              mediaTypeKey.startsWith('application/json') &&
-              !mediaType.schema.$ref;
-            if (hasInlineSchema) {
+            const hasInlineSchema = mediaType.schema && !mediaType.schema.$ref;
+            if (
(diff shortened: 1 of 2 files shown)


src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -19,7 +19,10 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
       each(obj, (response, responseKey) => {
         if (isOAS3) {
           each(response.content, (mediaType, mediaTypeKey) => {
-            const hasInlineSchema = mediaType.schema && !mediaType.schema.$ref;
+            const hasInlineSchema =
+              mediaType.schema &&
+              mediaTypeKey.startsWith('application/json') &&
+              !mediaType.schema.$ref;
```
dpopp07: So this logic is looking good to me but I would like it implemented in a slightly different way.

Whether a response "has an inline schema" or not is independent of the content type. So including that check as part of the boolean variable changes the meaning of the variable name, which is confusing. You can just make the if statement below:

```js
if (hasInlineSchema && mediaTypeKey.startsWith('application/json') {
// ...
}
```

test/plugins/validation/2and3/responses.js
```diff
@@ -268,6 +268,43 @@ describe('validation plugin - semantic - responses', function() {
         );
         expect(res.errors.length).toEqual(0);
       });
+
+      it('should not complain about schemas that dont require $ref', function() {
```
dpopp07: > should not complain about non-json response that defines an inline schema

dpopp07:
Also, CI is failing

PR #62, opened 2019-06-17
fix: use correct working dir when resolving multi-file definitions

### LINKED ISSUE #29: Problem following references when run outside of the root definition directory
Running from outside the root OpenAPI definition's directory can cause a failure to follow relative file references. For example, this fails:

```
/projects/api $ lint-openapi definition/root.yaml

[Error] There is a problem with the Swagger.
Error opening file "/projects/api/pets.yaml"
```

(Fails to find `pets.yaml` because the file is in `/projects/api/definition/`)

Even when this succeeds:
```
/projects/api $ cd definition/
/projects/api/definition $ lint-openapi root.yaml
```

dpopp07:
This fixes a bug caused by running the validator on a multi-file definition from any directory other than that of the root file.

The issue is that we read the API definition contents and store them in memory before we run the ref resolver package. When we run the resolver, it has no way to know the context of the original file so it uses the current working directory, the directory from which the command was executed.

One potential solution was to pass the filepath directly to the resolver. It can handle reading in the file and resolving references based on the location of the file as it should. The problem with this solution is that we lose all of the preprocessing we do to remove problematic characters from the document. This leads to the inability to run the validator on files that we would otherwise be able to.

So, that solution proved no good. What I did instead was programmatically change the working directory to the location of the root definition file, just for the resolving of the spec, and then change it back to the original working directory so that nothing else changes. This provides the resolver with the correct context, leaves us with the preprocessing step intact, and has no other impact on how the code runs.

Fixes #29

### DIFF
--- a/src/cli-validator/runValidator.js
+++ b/src/cli-validator/runValidator.js
@@ -197,17 +197,32 @@ const processInput = async function(program) {
       continue;
     }
 
+    // change working directory to location of root api definition
+    // this will allow the parser in `buildSwaggerObject` to resolve external refs correctly
+    const originalWorkingDirectory = process.cwd();
+    process.chdir(path.dirname(validFile));
+
     // validator requires the swagger object to follow a specific format
     let swagger;
     try {
       swagger = await buildSwaggerObject(input);
     } catch (err) {
+      // messing with the working directory can have unintended drawbacks
+      // using the try/catch to ensure that the directory is changed back
+      // to the original, even if the ref parser fails
+      // the tests fail without this line
+      process.chdir(originalWorkingDirectory);
+
       printError(chalk, 'There is a problem with the Swagger.', getError(err));
       // Uncomment the line below to see the stack trace when the validator fails
       // console.log(err.stack);
       exitCode = 1;
       continue;
     }
+
+    // return the working directory to its original location so that
+    // the rest of the program runs as expected
+    process.chdir(originalWorkingDirectory);
 
     // run validator, print the results, and determine if validator passed
     let results;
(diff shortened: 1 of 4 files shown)


test/cli-validator/tests/expectedOutput.js
```diff
@@ -226,10 +226,35 @@ describe('test expected output - OpenAPI 3', function() {
...
+
+    const allOutput = capturedText.join('');
+
+    expect(exitCode).toEqual(1);
+    expect(allOutput).toContain('errors');
+    expect(allOutput).toContain('API definition must have an `info` object');
+    expect(allOutput).toContain('warnings');
+    expect(allOutput).toContain(
+      'Operations must have a non-empty `operationId`.'
+    );
```
dpopp07: Confirming here that this test fails without my change

src/cli-validator/runValidator.js
```diff
@@ -197,18 +197,33 @@ const processInput = async function(program) {
...
       // Uncomment the line below to see the stack trace when the validator fails
       // console.log(err.stack);
       exitCode = 1;
       continue;
     }
 
+    // return the working directory to its original location so that
+    // the rest of the program runs as expected
+    process.chdir(originalWorkingDirectory);
+
```
mkistler: I think you could put this line in a `finally` clause of this `try/catch` and avoid duplicating it within the `catch`.

dpopp07: Great suggestion, thanks Mike! I knew there must be some way to not duplicate that line but I didn't think of `finally`

PR #63, opened 2019-06-17
fix: response links are not validated as responses

SamerJaser96:
walker.js now recognizes links in components and adds it to the dictionary so the validator no longer returns an error asking for the response link refs to be formatted as responses instead of following the format of '#/components/links'.

### DIFF
--- a/test/plugins/validation/2and3/walker.js
+++ b/test/plugins/validation/2and3/walker.js
@@ -600,6 +600,55 @@ describe('validation plugin - semantic - spec walker', () => {
           'description'
         ]);
       });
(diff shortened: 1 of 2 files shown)


test/plugins/validation/2and3/walker.js
```diff
@@ -600,6 +600,55 @@ describe('validation plugin - semantic - spec walker', () => {
           'description'
         ]);
       });
+      it('should return a problem for a parameters $ref in a response position', function() {
```
dpopp07: this message is incorrect

PR #66, opened 2019-06-18
ci: add test coverage reporting and tests on node 10 and 12

dpopp07:
As this package matures, it's important that we track how well our tests cover our code. This PR adds code coverage reporting using `nyc` and uses `codecov` to publish and track coverage.

I have already set up the token in Travis to talk to `codecov`. Once we start publishing coverage this way, we can display a bade with our coverage in the README.

Additionally, I decided that it would be a good idea to start testing with Node versions 10 and 12 since version 8 is being deprecated at the end of the year. The build for this package is very fast so it shouldn't be an annoyance to run the tests a few times.

Update: Coverage is 91% - not bad considering we've never looked at it before!


.travis.yml
```diff
@@ -1,11 +1,15 @@
...
 - npm run test-travis
+after_success:
+- npm run report-coverage
 deploy:
 - provider: script
   skip_cleanup: true
   script: npx semantic-release
   on:
-    branch: master
+    node: 12
```
dpopp07: The `.releaserc` file already specifies that the branch must be `master`.

package.json
```diff
@@ -22,11 +23,11 @@
     "deepmerge": "^2.1.1",
     "find-up": "^3.0.0",
     "globby": "^8.0.1",
-    "js-yaml": "^3.12.0",
+    "js-yaml": "^3.13.1",
     "json-dup-key-validator": "^1.0.2",
     "json-schema-ref-parser": "^5.1.3",
     "jsonschema": "^1.2.4",
-    "lodash": "^4.17.10",
+    "lodash": "^4.17.11",
```
mkistler: Glad you are fixing this too!

PR #67, opened 2019-06-19
fix: skips validation for non-string enum values

SamerJaser96:
Validator will skip validation of non-string enum values.

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
@@ -369,7 +369,7 @@ function checkEnumValues(schema, contextPath, config) {
         if (!isSnakecase(enumValue)) {
           result[checkStatus].push({
             path: contextPath.concat(['enum', i.toString()]),
-            message: 'Enum values must be lower snake case strings.'
+            message: 'Enum values must be lower snake case.'
           });
         }
       }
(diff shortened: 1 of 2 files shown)


src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -363,14 +363,15 @@ function checkEnumValues(schema, contextPath, config) {
...
-          path: contextPath.concat(['enum', i.toString()]),
-          message: 'Enum values must be lower snake case.'
-        });
+    if (typeof enumValue === 'string') {
+      const checkStatus = config.snake_case_only || 'off';
+      if (checkStatus.match('error|warning')) {
+        if (!isSnakecase(enumValue)) {
+          result[checkStatus].push({
+            path: contextPath.concat(['enum', i.toString()]),
+            message: 'Enum values must be lower snake case strings.'
```
dpopp07: I don't think we need this "strings" qualifier added to the end of the message. 1) They don't _have_ to be strings, 2) the term "snake case" only applies to strings. Just adding the check is enough

test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -1328,4 +1328,32 @@ describe('validation plugin - semantic - schema-ibm - OpenAPI 3', () => {
...
+    const spec = {
+      definitions: {
+        Thing: {
+          type: 'object',
+          description: 'thing',
+          properties: {
+            color: {
+              type: 'string',
+              description: 'some color',
+              enum: [1, 2, 3]
```
dpopp07: Will you update this property so it makes more sense? You describe a string property called "color", but the enums are numbers. It's a good test case, we just want the setup to be logical

dpopp07: The `type` is still "string", change it to "integer"

dpopp07: And actually, if its a "list of integers", it should probably be an array. So maybe change it to "integer"

PR #70, opened 2019-06-25
Response schemas

SamerJaser96:
validator now recognizes oneOf, allOf, and anyOf and will only issue a warning if the lists contain inline schemas instead of only refs

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/responses.js
+++ b/src/plugins/validation/2and3/semantic-validators/responses.js
@@ -10,90 +10,52 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
   result.warning = [];
 
   config = config.responses;
+
   walk(jsSpec, [], function(obj, path) {
     const contentsOfResponsesObject = path[path.length - 1] === 'responses';
     const isRef = !!obj.$ref;
+
     if (contentsOfResponsesObject && !isRef) {
       each(obj, (response, responseKey) => {
         if (isOAS3) {
           each(response.content, (mediaType, mediaTypeKey) => {
-            if (mediaType && mediaType.schema && mediaType.schema.oneOf) {
-              for (let i = 0; i < mediaType.schema.oneOf.length; i++) {
-                const hasInlineSchema =
-                  mediaType.schema &&
-                  mediaType.schema.oneOf &&
-                  !Object.keys(mediaType.schema.oneOf[i]).includes('$ref');
-                if (hasInlineSchema) {
-                  const checkStatus = config.inline_response_schema;
-                  if (checkStatus !== 'off') {
-                    result[checkStatus].push({
-                      path: [
-                        ...path,
-                        responseKey,
-                        'content',
-                        mediaTypeKey,
-                        'schema',
-                        'oneOf'
-                      ],
-                      message: INLINE_SCHEMA_MESSAGE
-                    });
-                  }
-                }
-              }
-            } else if (
-              mediaType &&
-              mediaType.schema &&
-              mediaType.schema.allOf
-            ) {
-              for (let i = 0; i < mediaType.schema.allOf.length; i++) {
-                const hasInlineSchema =
-                  mediaType.schema &&
-                  mediaType.schema.allOf &&
-                  !Object.keys(mediaType.schema.allOf[i]).includes('$ref');
-                if (hasInlineSchema) {
-                  const checkStatus = config.inline_response_schema;
-                  if (checkStatus !== 'off') {
-                    result[checkStatus].push({
-                      path: [
-                        ...path,
-                        responseKey,
-                        'content',
-                        mediaTypeKey,
-                        'schema',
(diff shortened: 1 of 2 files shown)


src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -10,18 +10,93 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
         if (isOAS3) {
           each(response.content, (mediaType, mediaTypeKey) => {
-            const hasInlineSchema = mediaType.schema && !mediaType.schema.$ref;
-            if (
-              hasInlineSchema &&
+            if (mediaType && mediaType.schema && mediaType.schema.oneOf) {
+              for (let i = 0; i < mediaType.schema.oneOf.length; i++) {
+                const hasInlineSchema =
+                  mediaType.schema &&
+                  mediaType.schema.oneOf &&
```
dpopp07: You already know that `mediaType.schema` and `mediaType.schema.oneOf` are `true`, these are redundant checks.

src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -10,18 +10,93 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+                  !Object.keys(mediaType.schema.oneOf[i]).includes('$ref');
```
dpopp07: This is the same as saying
```js
!mediaType.schema.oneOf[i].$ref
```

src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -10,18 +10,93 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+                  const checkStatus = config.inline_response_schema;
+                  if (checkStatus !== 'off') {
+                    result[checkStatus].push({
+                      path: [
+                        ...path,
+                        responseKey,
+                        'content',
+                        mediaTypeKey,
+                        'schema',
+                        'oneOf'
```
dpopp07: `i` should be included in the path, otherwise there is no way to separate between schemas within the compose schema.

src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -10,18 +10,93 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+                        'content',
+                        mediaTypeKey,
+                        'schema',
+                        'anyOf'
+                      ],
+                      message: INLINE_SCHEMA_MESSAGE
+                    });
+                  }
+                }
+              }
```
dpopp07: Will you refactor these checks so that you only need the code once? Something like
```js
const combinedSchemaTypes = ['allOf', 'oneOf', 'anyOf'];
combinedSchemaTypes.forEach(schemaType => {
  // ...
});
```

test/plugins/validation/2and3/responses.js
```diff
@@ -208,6 +245,53 @@ describe('validation plugin - semantic - responses', function() {
...
+                  200: {
+                    description: 'successful operation',
+                    content: {
+                      'application/json': {
+                        schema: {
+                          oneOf: [
+                            {
+                              type: 'object'
+                            }
+                          ]
```
dpopp07: I think it would be good to add one more element to the `oneOf` list that is a `$ref`, so you can see that it only flags the problem schemas.

test/plugins/validation/2and3/responses.js
```diff
@@ -208,6 +245,53 @@ describe('validation plugin - semantic - responses', function() {
...
+              get: {
+                summary: 'list stuff',
+                operationId: 'listStuff',
+                responses: {
+                  200: {
+                    description: 'successful operation',
+                    content: {
+                      'application/json': {
+                        schema: {
+                          oneOf: [
```
dpopp07: Use `anyOf` or `allOf` to cover more of the possibilities

src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -19,23 +19,60 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
       each(obj, (response, responseKey) => {
         if (isOAS3) {
           each(response.content, (mediaType, mediaTypeKey) => {
-            const hasInlineSchema = mediaType.schema && !mediaType.schema.$ref;
+            const combinedSchemaTypes = ['allOf', 'oneOf', 'anyOf'];
+            const hasOfProperty =
+              mediaType.schema.allOf ||
+              mediaType.schema.anyOf ||
+              mediaType.schema.oneOf;
```
dpopp07: First of all, this is very picky, but let's rename `hasOfProperty` to `hasCombinedSchema`.

Second of all, this will break if `mediaType.schema` is undefined. Just move this variable assignment to be within the if statement below (`mediaType.schema`).

PR #78, opened 2019-07-02
fix: make $ref pattern check configurable

### LINKED ISSUE #77: Unable to turn up "schema $refs must follow this format" warning to error
Am running v0.12.2 installed inside a Node 12.4 Docker container.

I'm getting a very helpful warning which looks like this:

```
  Message :   schema $refs must follow this format: *#/components/schemas*
  Path    :   paths./packages/{p_id}/clone.post.responses.200.content.application/json.schema.$ref
  Line    :   245

statistics

  Total number of errors   : 0
  Total number of warnings : 1

  warnings
   1 (100%) : schema $refs must follow this format
```

This warning has been very helpful because it helped me debug why a build with our testing team was failing. In order to get this check done early before builds go to our testing team, I'd like to turn this warning up to error - but I can't find the config for that.

From the code, it looks like this is raised by the walker, but when I set the config of walker to all errors:

```json
     "walker": {
       "no_empty_descriptions": "error",
      "has_circular_references": "error",
      "$ref_siblings": "error",
      "duplicate_sibling_description": "error"
     }
```

... then a warning is still returned.

Is there a way to turn this up to error?

dpopp07:
Makes the check for $ref patterns configurable with flag `incorrect_$ref_pattern`. I am open to changing that name. Documentation and defaults updated.

Also changes a test in `walker` that was previously overwriting the entire walker config object - not good for adding more tests!

Resolves #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,7 @@ The supported rules are described below:
 | has_circular_references       | Flag any circular references found in the Swagger spec.                      | shared |
 | $ref_siblings                 | Flag any properties that are siblings of a `$ref` property.                  | shared |
 | duplicate_sibling_description | Flag descriptions sibling to `$ref` if identical to referenced description.  | shared |
+| incorrect_$ref_pattern        | Flag internal `$ref` values that do not point to the section they should (e.g. referencing `parameters` from a `schema` field). | shared |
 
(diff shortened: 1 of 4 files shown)


dpopp07:
My only reasoning there was that it was consistent with the rule name `$ref_siblings`, but I agree it looks a little weird. I think I'll change it.

PR #80, opened 2019-07-15
Parameters referencing

SamerJaser96:
More specific checks for parameters that will check specifically for parameter objects and should not complain to properties named parameters. Will also check for parameters in a top level for swagger 2 and for parameters in path items.

### DIFF
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -1,14 +1,19 @@
 /*
     **********   DO NOT MODIFY THIS FILE   **********
+
+
     To configure the validator, edit the '.validaterc'
     file in the root folder of this project.
+
     This file contains the defaults rules for the
     validator. It is used to check for errors in
     the .validaterc file and for running the
     validator in default mode.
+
     NOTE: If you are contributing a validation to this code
     base, you should add the default rule to this object.
     In that case, please modify this file.
+
     Additionally, if a rule is being deprecated and/or replaced,
     this file must be modified (see bottom of file).
 */
@@ -30,7 +35,8 @@ const defaults = {
       'content_type_parameter': 'error',
       'accept_type_parameter': 'error',
       'authorization_parameter': 'warning',
-      'required_param_has_default': 'warning'
+      'required_param_has_default': 'warning',
+      'ref_and_inline_parameter': 'error'
     },
     'paths': {
       'missing_path_parameter': 'error',
@@ -60,9 +66,8 @@ const defaults = {
     'walker': {
       'no_empty_descriptions': 'error',
       'has_circular_references': 'warning',
-      '$ref_siblings': 'off',
-      'duplicate_sibling_description': 'warning',
-      'incorrect_ref_pattern': 'warning'
+      '$ref_siblings': 'off'
+      'duplicate_sibling_description': 'warning'
     }
   },
   'swagger2': {
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -1,19 +1,14 @@
 /*
     **********   DO NOT MODIFY THIS FILE   **********
-
-
     To configure the validator, edit the '.validaterc'
     file in the root folder of this project.
-
     This file contains the defaults rules for the
     validator. It is used to check for errors in
     the .validaterc file and for running the
     validator in default mode.
-
     NOTE: If you are contributing a validation to this code
     base, you should add the default rule to this object.
     In that case, please modify this file.
-
     Additionally, if a rule is being deprecated and/or replaced,
     this file must be modified (see bottom of file).
 */
@@ -35,8 +30,7 @@ const defaults = {
       'content_type_parameter': 'error',
       'accept_type_parameter': 'error',
       'authorization_parameter': 'warning',
-      'required_param_has_default': 'warning',
-      'ref_and_inline_parameter': 'error'
+      'required_param_has_default': 'warning'
     },
     'paths': {
       'missing_path_parameter': 'error',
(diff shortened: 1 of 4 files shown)


src/.defaultsForValidator.js
```diff
@@ -35,7 +35,8 @@ const defaults = {
       'content_type_parameter': 'error',
       'accept_type_parameter': 'error',
       'authorization_parameter': 'warning',
-      'required_param_has_default': 'warning'
+      'required_param_has_default': 'warning',
+      'ref_and_inline_parameter': 'error'
```
dpopp07: Why is this getting changed? We're not adding any new rules here

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -19,17 +19,14 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
   result.warning = [];
 
   config = config.parameters;
-
```
dpopp07: No need to remove this line

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -19,17 +19,14 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
   walk(jsSpec, [], function(obj, path) {
     // skip parameters within operations that are excluded
     if (obj['x-sdk-exclude'] === true) {
       return;
     }
-
-    const contentsOfParameterObject = path[path.length - 2] === 'parameters';
-
-    // obj is a parameter object
```
dpopp07: I'd like the keep the spacing here for cleanliness

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -19,17 +19,14 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+    const contentsOfParameterObject = params(path, isOAS3);
```
dpopp07: Change the function name to something like `isParameter`

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -157,6 +154,36 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+      const contentsOfParameterObject = sharedChecks;
+      return contentsOfParameterObject;
+    } else if (!isOAS3) {
+      //swagger 2 allows parameters to be in the top level, this checks for that
+      const contentsOfParameterObject =
+        sharedChecks || (path[0] === 'parameters' && path.length == 2);
+      return contentsOfParameterObject;
+    }
+  }
+}
```
dpopp07: This is super difficult to read. I think it would be cleaner to create a variable for each possibility, so it's clear what you are checking for.

Something like this:
```js
const isTopLevelParameter = !isOas3 && path[0] === 'parameters' && path.length === 2;

return isTopLevelParameter || ...
```

test/plugins/validation/2and3/parameters-ibm.js
```diff
@@ -299,9 +299,91 @@ describe('validation plugin - semantic - parameters-ibm', () => {
...
+              ]
+            }
+          }
+        }
+      };
+
+      const res = validate({ jsSpec: spec, isOAS3: false }, config);
+      expect(res.errors.length).toEqual(0);
+      expect(res.warnings.length).toEqual(0);
+    });
```
dpopp07: I think you can remove this test actually, I don't think it adds too much. I appreciate the robustness though

test/plugins/validation/2and3/parameters-ibm.js
```diff
@@ -510,5 +592,55 @@ describe('validation plugin - semantic - parameters-ibm', () => {
...
+              }
+            ]
+          }
+        }
+      };
+
+      const res = validate({ jsSpec: spec }, config);
+      expect(res.warnings.length).toEqual(0);
+      expect(res.errors.length).toEqual(0);
+    });
```
dpopp07: Again, this test is probably unnecessary so we can remove it.

src/.defaultsForValidator.js
SamerJaser96: This was an accident 😅

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -157,6 +157,40 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+
+  if (parameterObj) {
+    if (isOAS3) {
+      return pathCheck;
+    } else if (!isOAS3) {
+      //swagger 2 allows parameters to be in the top level, this checks for that
+      return pathCheck || isTopLevelParameter;
+    }
+  }
+}
```
dpopp07: I think this is still too complicated. Since this is tricky, I'll write some code as a suggestion:
```js
function isParameter(path, isOAS3) {
  const pathsForParameters = [
    'get',
    'put',
    'post',
    'delete',
    'options',
    'head',
    'patch',
    'trace',
    'components'
  ];
  
  const inParametersSection = path[path.length - 2] === 'parameters';

  // the above check is a necessary but not sufficient check for a parameter object
  // use the following checks to verify the object is where a parameter is supposed to be.
  // without these, a schema property named "parameters" would get validated as a parameter
  const isParameterByPath = pathsForParameters.includes(path[path.length - 3]);
  const isPathItemParameter =
    path[path.length - 4] === 'paths' && path.length === 4;
  const isTopLevelParameter =
    !isOAS3 && path[0] === 'parameters' && path.length === 2;

  return inParametersSection &&
    (isParameterByPath || isPathItemParameter || isTopLevelParameter);
}
```

src/.defaultsForValidator.js
```diff
@@ -65,7 +66,7 @@ const defaults = {
     'walker': {
       'no_empty_descriptions': 'error',
       'has_circular_references': 'warning',
-      '$ref_siblings': 'off',
+      '$ref_siblings': 'off'
```
dpopp07: This file shouldn't show any changes

PR #85, opened 2019-07-17
Request body

SamerJaser96:
Validator will return a warning for requestBody content objects with multiple schemas. Fixes #784


dpopp07:
We don't actually want to validate for this

PR #88, opened 2019-07-21
Skip case_convention checks on deprecated parameters and properties

mkistler:
This PR bypasses the case_convention checks on parameters and properties that are marked as `deprecated: true`.

I also slipped in a few minor documentation updates to the README.


src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -299,6 +299,9 @@ function checkPropNames(schema, contextPath, config) {
   forIn(schema.properties, (property, propName) => {
     if (propName.slice(0, 2) === 'x-') return;
 
+    // Skip snake_case_only checks for deprecated properties
+    if (property.deprecated === true) return;
```
SamerJaser96: would it be cleaner to have this as: if (isDeprecated) return;

PR #92, opened 2019-07-26
Pagination

SamerJaser96:
Validator will check for pagination requirements in parameters object


test/plugins/validation/2and3/parameters-ibm.js
```diff
@@ -585,5 +585,197 @@ describe('validation plugin - semantic - parameters-ibm', () => {
...
+              parameters: [
+                {
+                  start: 12,
+                  schema: {
+                    type: 'string'
+                  },
+                  description: 'good description',
+                  required: ['url', 'start']
+                }
+              ],
```
mkistler: This is not a valid spec.  A parameter must have "name" and "in".  Also, "required" is just a boolean.

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -25,8 +25,99 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
     if (obj['x-sdk-exclude'] === true) {
       return;
     }
-
     const contentsOfParameterObject = isParameter(path, isOAS3);
+    for (const head in obj.paths) {
```
mkistler: This validation is complicated enough that it probably warrants putting it into its own file.  I recommend creating a new file called "src/plugins/validation/2and3/semantic-validators/pagination-ibm.js" and put the logic there.

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -25,8 +25,99 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
+            const message =
+              'if start or cursor parameters are not present then the offset parameter should be defined as an integer and should be optional';
+            const checkStatus = config.invalid_limit_type;
+            if (checkStatus !== 'off') {
+              result[checkStatus].push({
+                path,
+                message
+              });
+            }
+          }
```
mkistler: I don't see the checks for properties in the response that are specified in the issue.  These need to be added.

PR #94, opened 2019-07-31
Pagination 

SamerJaser96:
validator now checks for pagination requirements


src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,181 @@
...
+          if (con == 'application/json') {
+            let arrayOnTop = false;
+            if (
+              (jsSpec.paths[head].get.responses[resp].content[con].schema &&
+                jsSpec.paths[head].get.responses[resp].content[con].schema
+                  .properties) ||
+              (jsSpec.paths[head].get.responses[resp].content[con].schema
+                .pagination &&
+                jsSpec.paths[head].get.responses[resp].content[con].schema
+                  .pagination.properties)
```
mkistler: Schema will generally not have `pagination` as a property.  If it is present, it would appear in `schema.properties`.

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,181 @@
...
+            ) {
+              for (const prop in jsSpec.paths[head].get.responses[resp].content[
+                con
+              ].schema.properties) {
+                if (
+                  jsSpec.paths[head].get.responses[resp].content[con].schema
+                    .properties[prop].type == 'array'
```
mkistler: You should be using `===` rather than `==` pretty much everywhere.

A good article on this: https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a
> Triple Equals is superior to double equals. Whenever possible, you should use triple equals to test equality.

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,181 @@
...
+                ) {
+                  arrayOnTop = true;
+                  break;
+                }
+              }
+              for (
+                let k = 0;
+                k < jsSpec.paths[head].get.parameters.length;
```
mkistler: This code will crash for a `get` that has no parameters.

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,181 @@
+module.exports.validate = function({ jsSpec }, config) {
+  const result = {};
+  result.error = [];
+  result.warning = [];
+
+  config = config.pagination;
+  const queryParameters = [];
```
mkistler: By defining and initializing `queryParameters` here, and not re-initializing it inside the loop, you are collecting up the query parameters for _all_ the operations in the spec into this list. That is definitely not right.

It would be better to move the definition and initialization of `queryParameters` inside the block that deals with a specific operation -- that will signal that it's value is intended to be just for that operation.

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,181 @@
...
+                  break;
+                }
+              }
+              for (
+                let k = 0;
+                k < jsSpec.paths[head].get.parameters.length;
+                k++
+              ) {
+                queryParameters.push(jsSpec.paths[head].get.parameters[k].name);
+              }
```
mkistler: If all you want is a list of the parameter names, there is a much more elegant and concise way to do that using the `map` operator.

A good article on this: https://medium.com/poka-techblog/simplify-your-javascript-use-map-reduce-and-filter-bd02c593cc2d

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                  queryParameters.includes('offset'))
+              ) {
+                const obj = jsonContent.schema;
+                // loop through parameters and make sure that the pagination related parameters follow the correct defintions
+                if (resolvedSpec.paths[head].get.parameters.length > 0) {
+                  for (
+                    let i = 0;
+                    i < resolvedSpec.paths[head].get.parameters.length;
+                    i++
+                  ) {
```
mkistler: You could make the code below much more readable by creating a local variable for the parameter you are examining in the loop body, e.g.:
```
    const param = resolvedSpec.paths[head].get.parameters[i]
```

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                        'get',
+                        'parameters',
+                        i,
+                        'name'
+                      ];
+                      const message =
+                        'limit must be of type integer and must be optional with default and maximum values';
+                      paginationWarning(message, path);
+                    }
+                    if (
```
mkistler: A comment on this line would be helpful.  Just use the text from the issue.

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+              if (
+                arrayOnTop &&
+                !jsonContent.schema.pagination &&
+                (queryParameters.includes('limit') ||
+                  queryParameters.includes('start') ||
+                  queryParameters.includes('cursor') ||
+                  queryParameters.includes('token') ||
+                  queryParameters.includes('offset'))
+              ) {
+                const obj = jsonContent.schema;
```
mkistler: It is good practice to use meaningful names for variables. You should use a better name here, something like "responseSchema".

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                // loop through parameters and make sure that the pagination related parameters follow the correct defintions
```
mkistler: In the current code structure, you will repeat the checks on the query parameter for _every_ matching response object, which is redundant work.  I think you should restructure the code so that first you determine if the operation is a "listX" -- by checking the parameters and responses -- and then if that is the case perform the validations once

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                    let i = 0;
+                    i < resolvedSpec.paths[head].get.parameters.length;
+                    i++
+                  ) {
+                    if (
+                      resolvedSpec.paths[head].get.parameters[i].name ===
+                        'limit' &&
+                      (resolvedSpec.paths[head].get.parameters[i].schema
+                        .type !== 'integer' ||
+                        resolvedSpec.paths[head].get.parameters[i].schema
```
mkistler: The `required` property is not an element of the `schema` but is a direct property of the parameter, so this check should be
```
 resolvedSpec.paths[head].get.parameters[i].required === true
```

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                          .required === true ||
+                        (!resolvedSpec.paths[head].get.parameters[i].default &&
+                          !resolvedSpec.paths[head].get.parameters[i].maximum))
```
mkistler: Both the `default` and `maximum` values for a parameter are specified within the parameter schema, so this term should be:
```
(!resolvedSpec.paths[head].get.parameters[i].schema.default &&
                          !resolvedSpec.paths[head].get.parameters[i].schema.maximum))
```

src/plugins/validation/2and3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,216 @@
...
+                  'application/json',
+                  'schema',
+                  'properties',
+                  'name'
+                ];
+                nextChecker(obj, path);
+                queryLimitChecker(obj, path, queryParameters);
+                queryOffsetChecker(obj, path, queryParameters);
+                totalCountChecker(obj, path);
+                nextTokenChecker(obj, path, queryParameters);
```
mkistler: Since this batch of five methods are always called together, I don't think there is any value is making them separate methods.  Just create one method, called something like "checkResponseProperties", and do all the checks for the response properties within that method.

mkistler:
Sorry ... I started to re-review and then realized that you had not pushed up your changes yet.

mkistler:
Replaced by PR #118

PR #96, opened 2019-08-02
fix: check for operationId naming

SamerJaser96:
will check operationIds naming conventions for certain paths and path parameters
Fixes #957


domodwyer:
Any progress on this? Happy to help!

It would be pretty helpful for us 👍

dpopp07:
@domodwyer We definitely still want this but I think we want a different approach. I am going to close this PR and start working on a new one when I find some time. Or, if you want to take a stab at it, feel free!

PR #107, opened 2019-09-27
adds discriminator validator to conform to openAPI 2 and 3 specifications

ExiliumAeternum:
The Swagger 2.0 spec and Open API 3 spec says 'The [discriminator] property name used MUST be defined at this schema'.
The validator currently does not flag violations of this requirement


src/plugins/validation/2and3/semantic-validators/discriminator.js
```diff
@@ -0,0 +1,74 @@
...
+  const schemas = jsSpec.components.schemas;
+
+  const basePath = ['components', 'schemas'];
+
+  each(schemas, (schema, schemaName) => {
+    if (has(schema, 'discriminator')) {
+      const { discriminator } = schema;
+
+      // If discriminator is not an object, error out and return
+      if (typeof discriminator === 'object') {
```
mkistler: In OpenAPI 2.0, the `discriminator` is just a string property containing the name of the discriminator property in the schema.  So what you have done here is correct for OpenAPI 3.0 but not for OpenAPI 2.0.

There are two ways to address this: 1) use alternate paths in this validator for OAS2 vs OAS3, or 2) move this validation to OAS3, and implement a separate (but very similar) validator for OAS2 (aka Swagger).

test/plugins/validation/2and3/discriminator.js
```diff
@@ -0,0 +1,170 @@
...
+    expect(res.errors[1].path).toEqual(
+      ['components', 'schemas', 'Food', 'discriminator', 'propertyName'].join(
+        '.'
+      )
+    );
+    expect(res.errors[1].message).toEqual(
+      'The discriminator property name used must be defined in this schema'
+    );
+  });
+});
```
mkistler: Nice tests!

src/plugins/validation/oas3/semantic-validators/discriminator.js
```diff
@@ -0,0 +1,75 @@
...
+
+      // If discriminator propertyName is not a string, error out and return
+      const { propertyName } = discriminator;
+      if (typeof propertyName !== 'string') {
+        errors.push({
+          path: basePath
+            .concat([schemaName, 'discriminator', 'propertyName'])
+            .join('.'),
+          message:
+            '`propertName` inside discriminator object must be of type string'
```
mkistler: Typo here: "propertName" -> "propertyName"

PR #109, opened 2019-10-04
fix: Consider as valid multiple schemes in a security requirement object

### LINKED ISSUE #108: Multiple scheme security requirement should be allowed
https://github.com/IBM/openapi-validator/blob/0c0b39c3f83f14f4213afa5eb62a5076d0f1f9d6/src/plugins/validation/2and3/semantic-validators/security-ibm.js#L60-L75

According to the [standard](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#security-requirement-object):
> Security Requirement Objects that contain multiple schemes require that all schemes MUST be satisfied for a request to be authorized. This enables support for scenarios where multiple query parameters or HTTP headers are required to convey security information.

You're checking that each Security Requirement Object contains only one scheme, and setting a _non-configurable warning_.

I suggest one of the following:
a) Remove this validation, because is non-standard.
b) Make it configurable so user can decide what to do with it.

Thanks.

PD: Are you reviewing PRs?

**UPDATE**
This should be also be fixed, because it only labels the first one as _used_.
https://github.com/IBM/openapi-validator/blob/f8cc2bd6411c4450682b6c4c10703f83f2cde19d/src/plugins/validation/2and3/semantic-validators/security-definitions-ibm.js#L77-L78

jormaechea:
Multiple security schemes are now considered as valid and flagged as used.
Added unit tests for the new code.

Fixes #108

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/security-definitions-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/security-definitions-ibm.js
@@ -74,24 +74,25 @@ module.exports.validate = function({ resolvedSpec, isOAS3 }, config) {
 
   function flagUsedDefinitions(security) {
     security.forEach(scheme => {
-      // each object in this array should only have one key - the name of the scheme
-      const name = Object.keys(scheme)[0];
+      const schemeNames = Object.keys(scheme);
 
-      // make sure this scheme was in the security definitions, then label as used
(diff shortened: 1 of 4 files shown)


jormaechea:
This is ready for review!

PR #111, opened 2019-11-17
Add check for 204 response with response body

mkistler:
This PR adds a check for a 204 response that defines a response body.  I defined this rule as an error and did not make it configurable since returning a response body with a 204 response violates the HTTP spec:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.5

The PR also contains a fix for a latent problem in the validator that was exposed by this new feature -- semantic validators with the same name as a shared (`2and3`) validator were not being run.

### DIFF
--- a/src/cli-validator/utils/validator.js
+++ b/src/cli-validator/utils/validator.js
@@ -60,14 +60,21 @@ module.exports = function validateSwagger(allSpecs, config) {
   }
 
   // run semantic validators
-  const allValidators = Object.assign(
-    {},
-    semanticValidators,
-    sharedSemanticValidators
-  );
 
-  Object.keys(allValidators).forEach(key => {
-    const problem = allValidators[key].validate(allSpecs, config);
+  Object.keys(semanticValidators).forEach(key => {
+    const problem = semanticValidators[key].validate(allSpecs, config);
+    if (problem.errors.length) {
+      validationResults.errors[key] = [...problem.errors];
+      validationResults.error = true;
+    }
+    if (problem.warnings.length) {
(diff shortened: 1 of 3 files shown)


src/cli-validator/utils/validator.js
```diff
@@ -60,14 +60,21 @@ module.exports = function validateSwagger(allSpecs, config) {
   }
 
   // run semantic validators
-  const allValidators = Object.assign(
-    {},
-    semanticValidators,
-    sharedSemanticValidators
-  );
```
dpopp07: This was a great catch!

src/cli-validator/utils/validator.js
```diff
@@ -60,14 +60,21 @@ module.exports = function validateSwagger(allSpecs, config) {
...
+    if (problem.warnings.length) {
+      validationResults.warnings[key] = [...problem.warnings];
+      validationResults.warning = true;
+    }
+  });
+
+  Object.keys(sharedSemanticValidators).forEach(key => {
+    const problem = sharedSemanticValidators[key].validate(allSpecs, config);
     if (problem.errors.length) {
       validationResults.errors[key] = [...problem.errors];
```
dpopp07: Now that this has been pointed out, I actually think this is still buggy. If an error was reported by a semantic validator with key `foo` and then also by a shared validator with key `foo`, the first errors would be overwritten, right? Maybe we need to rethink this a little more

mkistler: I see that now.  Just pushed a fix to handle this case.

PR #112, opened 2019-11-17
Add rule to check for duplicate path parameters

mkistler:
This PR adds a rule, with default configuration of 'warn', that flags identical path parameters defined on the individual operations of a path.  These parameters should be defined on the path object.

Note that only completely identical parameters are flagged -- variations in the parameter description will suppress the warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ The supported rules are described below:
 | missing_path_parameter      | For a path that contains path parameters, flag any operations that do not correctly define those parameters. | shared |
 | snake_case_only             | Flag any path segment that does not use snake case.                                                          | shared |
 | paths_case_convention       | Flag any path segment that does not follow a given case convention. snake_case_only must be 'off' to use.    | shared |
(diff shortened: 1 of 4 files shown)


mkistler:
In the case of four operations, two define identical path parameters and two others define identical path parameters but different from the first two, it was a conscious decision to emit no warning.  In this case, we _could_ move one or the other set of path parameters up to the path object, remove them from the operations that define those exact parameters, and leave the other ones, but that only   eliminates one set of parameters.  If three operations defined identical parameters and one was different, we could potentially eliminate two duplicate sets of parameters.  But I think this could open a can of worms about how to optimally reduce the number of redundant path parameters (what about 5 operations where 3 have identical params and the other two have different identical params?). 

I think the current design gives the clearest guidance with the clearest benefit.

dpopp07:
Sounds good! Just wanted to do the due diligence and ask about that. This looks ready to merge. Will you add a `feat` header to the commit message when you squash to trigger a release?

PR #118, opened 2019-12-02
Add rule to check pagination list operations conform to API Handbook

mkistler:
This PR adds a rule that checks paginated list operations for conformance to basic requirements of the API Handbook for paginated collections.  All checks are grouped under a single rule that is set to "warning" by default.

I've only implemented this rule for oas3 schemas.  The oas2 support would require significant additional logic and I considered it not worth the effort, at least for this initial PR.  Support for oas2 could be added later if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,8 @@ The supported categories are described below:
 | schemas    | Rules pertaining to [Schema Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject)       |
 | security_definitions | Rules pertaining to [Security Definition Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject) |
 | security   | Rules pertaining to [Security Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject) |
-| walker     | Rules pertaining to the entire document.                                                                                                   |
-
+| walker     | Rules pertaining to the entire document.                                                                                              |
+| pagination | Rules pertaining to [Pagination Objects](https://pages.github.ibm.com/CloudEngineering/api_handbook/collections/pagination.html)   |
 #### Rules
 
 Each category contains a group of rules. The spec that each rule applies to is marked in the third column. For the actual configuration structure, see the [default values](#default-values).
@@ -199,7 +199,13 @@ The supported rules are described below:
 | no_array_responses           | Flag any operations with a top-level array response.                                | shared   |
 | parameter_order              | Flag any operations with optional parameters before a required param.               | shared   |
 | no_request_body_content      | [Flag any operations with a `requestBody` that does not have a `content` field.][3] | oas3     |
-| no_request_body_name         | Flag any operations with a non-form `requestBody` that does not have a name set with `x-codegen-request-body-name`. | oas3 |
+| no_request_body_name         | Flag any operations with a non-form `requestBody` that does not have a name set with `x-codegen-request-body-name`. | oas3|
+
+##### pagination
+| Rule                        | Description                                                              | Spec   |
+| --------------------------- | ------------------------------------------------------------------------ | ------ |
+| pagination_style            | Flag any parameter or response object that doesnt follow pagination requirements. | oas3 |
+
 
 ##### parameters
 | Rule                        | Description                                                              | Spec   |
@@ -361,6 +367,11 @@ The default values for each rule are described below.
 | no_summary                   | warning |
 | no_array_responses           | error   |
 | parameter_order              | warning |
+
+###### pagination
+| Rule                        | Default |
+| --------------------------- | --------|
+| pagination_style            | warning |
 
 ###### parameters
 | Rule                        | Default |
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ The supported categories are described below:
 | security_definitions | Rules pertaining to [Security Definition Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityDefinitionsObject) |
 | security   | Rules pertaining to [Security Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject) |
 | walker     | Rules pertaining to the entire document.                                                                                              |
-| pagination | Rules pertaining to [Pagination Objects](https://pages.github.ibm.com/CloudEngineering/api_handbook/collections/pagination.html)   |
+| pagination | Rules pertaining to pagination   |
 #### Rules
 
 Each category contains a group of rules. The spec that each rule applies to is marked in the third column. For the actual configuration structure, see the [default values](#default-values).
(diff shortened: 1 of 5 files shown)


src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,229 @@
...
+            'application/json',
+            'schema',
+            'properties',
+            'offset'
+          ],
+          message: `The "offset" property in the response body of a paginated list operation must be of type integer and required.`
+        });
+      }
+    }
+  }
```
dpopp07: Since these two checks look identical in every way but the name of the parameter, I would suggest creating a new function to handle this and passing in both "limit" and "offset".

Something like
```js
function doTheCheck(paramName) {
  const prop = jsonResponse.schema.properties[paramName];
  if (!prop) {
    result[checkStatus].push({
      path: [
        'paths',
        path,
        'get',
        'responses',
        resp,
        'content',
        'application/json',
        'schema',
        'properties'
      ],
      message: `The response body of a paginated list operation must contain a "${paramName}" property.`
    });
  } else if (
    prop.type !== 'integer' ||
    !jsonResponse.schema.required ||
    jsonResponse.schema.required.indexOf(paramName) === -1
  ) {
    result[checkStatus].push({
      path: [
        'paths',
        path,
        'get',
        'responses',
        resp,
        'content',
        'application/json',
        'schema',
        'properties',
        paramName
      ],
      message: `The "${paramName}" property in the response body of a paginated list operation must be of type integer and required.`
    });
  }
}
```

It would need access to variables like `jsonResponse` and `result` but I think you can make it a closure inside the `validate` method to maintain their context. This would remove some code, but it's not a big deal. So, up to you.

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -28,15 +28,20 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
 
-    const resp = getJsonResponse(resolvedSpec.paths[path].get);
-    const jsonResponse = resp
-      ? resolvedSpec.paths[path].get.responses[resp].content['application/json']
-      : undefined;
+    // Find first success response code
+    const resp = Object.keys(operation.responses || {}).find(code => code.startsWith('2'));
+    // Now get the json content of that response
+    const content = resp && operation.responses[resp].content;
+    const jsonResponse = content && content['application/json'];
```
mkistler: I reworked this section to put all the code inline.  I think this makes it clearer what `resp` and `jsonResponse` contain.

@dpopp07  I would like your opinion on the style I used here -- using `&&` as a form of "optional chaining".

README.md
```diff
@@ -199,7 +199,13 @@ The supported rules are described below:
 | no_array_responses           | Flag any operations with a top-level array response.                                | shared   |
 | parameter_order              | Flag any operations with optional parameters before a required param.               | shared   |
 | no_request_body_content      | [Flag any operations with a `requestBody` that does not have a `content` field.][3] | oas3     |
-| no_request_body_name         | Flag any operations with a non-form `requestBody` that does not have a name set with `x-codegen-request-body-name`. | oas3 |
+| no_request_body_name         | Flag any operations with a non-form `requestBody` that does not have a name set with `x-codegen-request-body-name`. | oas3|
+
+##### pagination
+| Rule                        | Description                                                              | Spec   |
+| --------------------------- | ------------------------------------------------------------------------ | ------ |
+| pagination_style            | Flag any parameter or response object that doesnt follow pagination requirements. | IBM guidelines |
```
dpopp07: Sorry if I caused any confusion, but I do think this should still be `oas3`. These "Spec" fields directly correspond to the configuration file

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
mkistler: Rather than create a function for this, I just created a variable to hold the common elements of the path.  By itself this eliminates almost 30 lines of code, but keeps the flexibility of the separate logic for these two parameters.

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
dpopp07: This looks good! The "option chaining" you're referring to is fairly common and adds a nice "safety" to JS code

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
dpopp07: That sounds good to me, my spidey-senses just get tingly when I see opportunity to reduce duplicated logic 🙂

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // - If the operation has an `offset` query parameter, the response body must contain an `offset` property this is type integer and required
+
+    if (offsetParamIndex !== -1) {
+      const offsetProp = jsonResponse.schema.properties.offset;
+      if (!offsetProp) {
+        result[checkStatus].push({
+          path: propertiesPath,
+          message: `The response body of a paginated list operation must contain a "offset" property.`
```
danhudlow: I suggest `The offset parameter of a paginated list must also be reflected as an "offset" property in the response body schema.` since the message as written is not strictly true.

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -0,0 +1,192 @@
...
+        result[checkStatus].push({
+          path: ['paths', path, 'get', 'parameters', offsetParamIndex],
+          message: 'The offset parameter must be of type integer and optional.'
+        });
+      }
+    }
+
+    // - if the operation has a `start`, `cursor`, or `page_token` query parameter, it must be type string and optional
+
+    const startParamNames = ['start', 'cursor', 'page_token'];
```
danhudlow: I'm surprised to see the flexibility here—the handbook says such a parameter MUST be called `start`.

danhudlow:
Other things that seem closely related but are not verified (typing these from memory, so please feel free to consult the handbook and correct me):
- `first.href` should always be present in the response
- `next.href` should be used for the next page URL
- If tokens are provided in the response, they should be in `first.token` and `next.token` (Hmm, do we require echoing the current token in `token`? If not, maybe we should.)
- The array property used for returned resources is named for the last segment of the request path.

Feel free to consider these out of scope, but they might inform an update to the name of this particular rule—maybe `pagination_control_style`?

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
mkistler: Agreed that the message as is was not strictly true, so I will change it to:
> A paginated list operation with an "offset" parameter must include an "offset" property in the response body schema.

Which I think captures the spirit of your suggested change.

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
mkistler: I suppose I'm being an easy grader here 😊

PR #123, opened 2020-01-27
fix: swagger2 validator should only report error if there's body parameter but no consume

SaltedCaramelCoffee:
Currently, swagger2 validator reports error on all PUT or POST operations that has no consume field. This is technically incorrect because a POST operation is not required to have a request body and if there is no request body, there does not need to be a consumes field.

The fix now checks for body parameters on the path level and also operation level, and will only report error when there is body parameter, but no consume field.

### DIFF
--- a/src/plugins/validation/swagger2/semantic-validators/operations-ibm.js
+++ b/src/plugins/validation/swagger2/semantic-validators/operations-ibm.js
@@ -77,7 +77,7 @@ module.exports.validate = function({ jsSpec }, config) {
             result[checkStatus].push({
               path: `paths.${pathKey}.${opKey}.consumes`,
               message:
-                'PUT and POST operations with body parameter must have a non-empty `consumes` field.'
+                'PUT and POST operations with a body parameter must have a non-empty `consumes` field.'
             });
           }
         }
--- a/src/plugins/validation/swagger2/semantic-validators/operations-ibm.js
+++ b/src/plugins/validation/swagger2/semantic-validators/operations-ibm.js
@@ -48,19 +48,23 @@ module.exports.validate = function({ jsSpec }, config) {
 
         // Check for body parameter in path
         let hasBodyParamInPath = false;
-        path.parameters.forEach(parameter => {
-          if (parameter.in === 'body') {
-            hasBodyParamInPath = true;
-          }
-        });
+        if (path.parameters) {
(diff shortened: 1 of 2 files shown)


src/plugins/validation/swagger2/semantic-validators/operations-ibm.js
```diff
@@ -46,14 +46,38 @@ module.exports.validate = function({ jsSpec }, config) {
...
+          (hasBodyParamInOps || hasBodyParamInPath)
+        ) {
           const checkStatus = config.no_consumes_for_put_or_post;
 
           if (checkStatus !== 'off') {
             result[checkStatus].push({
               path: `paths.${pathKey}.${opKey}.consumes`,
               message:
-                'PUT and POST operations must have a non-empty `consumes` field.'
+                'PUT and POST operations with body parameter must have a non-empty `consumes` field.'
```
dpopp07: Just a small wording suggestion. Might want to do this on line 2 as well

```suggestion
                'PUT and POST operations with a body parameter must have a non-empty `consumes` field.'
```

test/plugins/validation/swagger2/operations-ibm.js
```diff
@@ -84,7 +84,7 @@ describe('validation plugin - semantic - operations-ibm - swagger2', function()
     expect(res.errors.length).toEqual(1);
     expect(res.errors[0].path).toEqual('paths./CoolPath.post.consumes');
     expect(res.errors[0].message).toEqual(
-      'PUT and POST operations must have a non-empty `consumes` field.'
+      'PUT and POST operations with body parameter must have a non-empty `consumes` field.'
```
dpopp07: Will you also add a test exercising your changes? As in, a test where there is no consumes and no body parameter and show that there are no errors

test/plugins/validation/swagger2/operations-ibm.js
```diff
@@ -84,11 +84,121 @@ describe('validation plugin - semantic - operations-ibm - swagger2', function()
...
+        no_consumes_for_put_or_post: 'error'
+      }
+    };
+
+    const spec = {
+      paths: {
+        '/CoolPath': {
+          parameters: [
+            {
+              name: 'BadParameter',
```
dpopp07: Poor parameter - there's something wrong with the operation, not him 😉

PR #124, opened 2020-01-29
feat: operationId should conform to naming convention

SaltedCaramelCoffee:
[documentation for operation Id naming convention
](http://watson-developer-cloud.github.io/api-guidelines/swagger-coding-style#operationid)

We should add support to the validator to check for conformance to these conventions and report any non-conformance as a warning

This PR takes the following approach to check for operationId naming convention: 

- Only consider paths `p` for which `p` and `p/{param}` exist, `p` has a GET and POST path, and `p/{param}` has a GET, a DELETE, and a POST or PUT or PATCH.
- For paths that satisfy this constraint, we verify that
  - the operationId for GET on `p` starts with "list"
  - the operationId for POST on `p` starts with "create" or "add"
  - the operationId for GET on `p/{param}` starts with "get"
  - the operationId for DELETE on `p/{param}` starts with "delete"
  - the operationId for POST/PUT/PATCH on `p/{param}` starts with "update"

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/operation-ids.js
+++ b/src/plugins/validation/2and3/semantic-validators/operation-ids.js
@@ -54,14 +54,14 @@ module.exports.validate = function({ resolvedSpec }) {
     return !!prev;
   };
 
-  const operationIdPassedConventionCheck = (opKey, operationId, pathParam) => {
+  const operationIdPassedConventionCheck = (opKey, operationId, hasPathParam) => {
     // Only consider paths for which
     // - paths with no path param has a GET and POST path
     // - paths with path param has a GET, a DELETE, and a POST or PUT or PATCH.
 
     let checkPassed = true;
 
-    if (!pathParam) {
+    if (!hasPathParam) {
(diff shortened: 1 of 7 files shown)


src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -50,15 +54,74 @@ module.exports.validate = function({ resolvedSpec }) {
     return !!prev;
   };
 
+  const operationIdPassedConventionCheck = (opKey, operationId, pathParam) => {
```
dpopp07:
```suggestion
  const operationIdPassedConventionCheck = (opKey, operationId, hasPathParam) => {
```

This is a little picky but since JS is loosely typed, it's nice when variable names hint at the type of the variable. `pathParam` is a boolean, but sounds like it's a param object. `hasPathParam` makes it clear that the variable is a boolean

dpopp07:
I approved it but now noticing that Travis builds are failing - make sure you get those sorted before merging!

PR #128, opened 2020-02-04
Add --fix flag to autofix errors and warnings reported

rmkeezer:
This adds the functionality of attempting to auto-fix the API document based on the warnings and errors received in the validator report.

This is experimental and currently tested with only JSON API docs.


src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,212 @@
...
+        if (!Array.isArray(path)) {
+          path = path.split('.');
+        }
+        if (path[path.length - 1].includes('[')) {
+          path.push(path[path.length - 1].split(']')[0].split('[')[1]);
+          path[path.length - 2] = path[path.length - 2].split(['['])[0];
+        }
+        const { obj, parent, ppParent } = getObjectFromPath(path);
+        if (message.includes('Schema must have a non-empty description')) {
+          obj['description'] = path[path.length - 1];
```
mkistler: I'm uncomfortable with this fix.  The spirit of a `description` is to _add_ information, and this appears to just repeat information that is already present.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,212 @@
...
+        } else if (
+          message.includes(
+            'Schema properties must have a description with content in it'
+          )
+        ) {
+          parent['description'] = path[path.length - 2];
```
mkistler: Same concern as above here.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,212 @@
...
+            'Rely on the `securitySchemas` and `security` fields to specify authorization methods.'
+          )
+        ) {
+          if (!('security' in originalJSON)) {
+            originalJSON['security'] = [{ IAM: [] }];
+            originalJSON['components']['securitySchemes'] = {
+              IAM: {
+                type: 'apiKey',
+                name: 'Authorization',
+                description: 'Your IBM Cloud IAM access token.',
```
mkistler: It would be great if you could use the description from the `Authorization` header here (if there is one).

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,212 @@
...
+        pathSegments[i] = snakeCase.snakeCase(segment);
+      }
+    });
+    parent[pathSegments.join('/')] = parent[path[path.length - 1]];
+    delete parent[path[path.length - 1]];
+  });
+
+  console.log('Writing Fixes to new file');
+  const fixedJSON = JSON.stringify(originalJSON, null, 4);
+  fs.writeFileSync('test1_new.json', fixedJSON);
```
mkistler: You need to set the file name correctly here.

mkistler:
Oh ... and we'll need to update the doc to explain this option.

src/cli-validator/index.js
```diff
@@ -43,6 +43,10 @@ program
   .option(
     '-e, --errors_only',
     'only print the errors, ignore the warnings'
+  )
+  .option(
+    '-f, --fix',
+    'automatically fix errors and write api doc to a new file {filename}_fixed.json'
```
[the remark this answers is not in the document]
rmkeezer: I agree, I made it write to a separate file as a convenience for testing when trying new unreliable fixes but now I'll make it do the fixes in-place as the fixes are more reliable.

src/cli-validator/utils/fixProblems.js
rmkeezer: Good point, it doesn't really add any information so I'll leave it out.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
+const each = require('lodash/each');
```
mkistler: @dpopp07 Should we have copyright stmts in the validator files?

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
...
+  const types = errorsOnly ? ['errors'] : ['errors', 'warnings'];
+
+  const originalJSON = JSON.parse(originalFile);
+
+  const paramsToDelete = {};
+  const paramsToPrepend = {};
+  const newPathNames = {};
+  const definitionsToDelete = {};
+
+  const markForDeletion = path => {
```
mkistler: Since these are effectively method declarations, I think you should add a comment above each that briefly describes what they do.  For example, I think you expect `path` to the be the path to a parameter -- it would wrong to think that this method could mark an arbitrary element for deletion.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
...
+  types.forEach(type => {
+    each(results[type], problems => {
+      problems.forEach(problem => {
+        const message = problem.message.split(':')[0];
+        let path = problem.path;
+        // path needs to be an array to get the line number
+        if (!Array.isArray(path)) {
+          path = path.split('.');
+        }
+        if (path[path.length - 1].includes('[')) {
```
mkistler: A comment here would be really helpful!

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
...
+      paramsToPrepend[pathstr].findIndex(compFunc) != -1
+    ) {
+      paramsToPrepend[pathstr].splice(
+        paramsToPrepend[pathstr].findIndex(compFunc),
+        1
+      );
+    }
+  };
+
+  const getObjectFromPath = path => {
```
mkistler: Comment here please, particularly to describe the return value.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
...
+            originalJSON['security'] = [{ IAM: [] }];
+            originalJSON['components']['securitySchemes'] = {
+              IAM: {
+                type: 'apiKey',
+                name: 'Authorization',
+                description: obj['description']
+                  ? obj['description']
+                  : 'Your IBM Cloud IAM access token.',
+                in: 'header'
+              }
```
mkistler: This fix is very specific to IBM Cloud services.  I think some additional logic is needed here.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,229 @@
...
+            };
+          }
+          markForDeletion(path);
+          unmarkIfMarkedForMoving(path, parent);
+        } else if (
+          message.includes('operationIds must follow case convention')
+        ) {
+          parent[path[path.length - 1]] = snakeCase.snakeCase(
+            parent[path[path.length - 1]]
```
mkistler: The operationId case convention is actually configurable, so this might not be the right fix.  We may need some additional logic here.

src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -112,7 +112,7 @@ module.exports.validate = function({ resolvedSpec }) {
         // If PATCH operation doesn't exist for path, POST operationId should start with "update"
         if (
           !allPathOperations.includes('patch') &&
-          !operationId.match(/^(update[a-zA-Z0-9_]+/m)
+          !operationId.match(/^update[a-zA-Z0-9_]+/m)
```
mkistler: Oh good fix!

src/cli-validator/utils/fixProblems.js
dpopp07: Yes, probably. I don't think many, if any, have them though.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,259 @@
...
+        } else if (
+          message.includes(
+            'Schema properties must have a description with content in it'
+          )
+        ) {
+          // parent['description'] = path[path.length - 2];
+        } else if (
+          message.includes('Definition was declared but never used in document')
+        ) {
+          definitionsToDelete[path.join(',')] = true;
```
mkistler: If you made `definitionsToDelete` an array, you could simply push the path (leaving it as an array) onto `definitionsToDelete` and avoid the `join` here and corresponding `split` below when you process the deletes.

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,259 @@
...
+  const types = errorsOnly ? ['errors'] : ['errors', 'warnings'];
+
+  const originalJSON = JSON.parse(originalFile);
+
+  const paramsToDelete = {};
+  const paramsToPrepend = {};
+  const newPathNames = {};
+  const definitionsToDelete = {};
+
+  // Mark a parameter for deletion given a path separated by commas
```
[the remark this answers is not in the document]
rmkeezer: Whoops that is correct I misread my own code thanks!

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,259 @@
...
+          } else if (opIdCase === 'lower_camel_case') {
+            parent[path[path.length - 1]] = camelCase(
+              parent[path[path.length - 1]]
+            );
+          }
+        } else if (
+          message.includes(
+            'Required parameters should appear before optional parameters.'
+          )
+        ) {
```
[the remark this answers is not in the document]
rmkeezer: It does get moved to be the first parameter, but isn't that fine if the first parameters are all required and the last parameters are all optional? Does the ordering matter inside the group of required parameters or the group of optional parameters?

src/cli-validator/utils/fixProblems.js
rmkeezer: Good idea, done

src/cli-validator/utils/fixProblems.js
```diff
@@ -0,0 +1,259 @@
...
+        if (!Array.isArray(path)) {
+          path = path.split('.');
+        }
+
+        // In the case that a path includes brackets ['path', 'to', 'array[index]'],
+        // remove the brackets and add the index inside as a new element ['path', 'to', 'array', 'index']
+        if (path[path.length - 1].includes('[')) {
+          path.push(path[path.length - 1].split(']')[0].split('[')[1]);
+          path[path.length - 2] = path[path.length - 2].split(['['])[0];
+        }
```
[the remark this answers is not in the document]
rmkeezer: Good catch, I rewrote the logic to work like it should

src/cli-validator/utils/fixProblems.js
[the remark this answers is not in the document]
mkistler: It's "fine" in the sense that it satisfies the rule, but it is a more invasive change than necessary and is more likely to be objectionable to the user.  But more importantly, we want the parameter list to correspond (as closely as we can) to the order of the parameters in the SDKs, and the generator will move it to above the first optional parameter.

src/cli-validator/index.js
```diff
@@ -43,6 +43,14 @@ program
   .option(
     '-e, --errors_only',
     'only print the errors, ignore the warnings'
+  )
+  .option(
+    '-f, --fix',
+    'automatically fix errors and make in-place changes to the API doc file'
+  )
+  .option(
+    '--fix_newfile',
```
danhudlow: Perhaps `--fix-new-file`?

rmkeezer: Yeah I agree the syntax is awkward, I keep accidently putting --fix-newfile instead of --fix_newfile

efussi:
This is very cool, I installed this PR locally and gave it a try. One thing I noted:
- The help reads
   > --fix_new_file                 automatically fix errors and write api doc to a new file {filename}_fixed.json
- but the generated file was named **{filename}_new.json**

PR #130, opened 2020-02-04
fix: print error message and the usage menu for an unknown command line argument

presto21:
When an unsupported option is given, a single-line error is printed. Now, the usage menu is printed along with the single-line error.

Change: 

- unsupportedOptionError.js modifies the commander code
- modified index.js file to include the unsupportedOptionError.js file.

Commander code was:
```
Command.prototype.unknownOption = function(flag) {
if (this._allowUnknownOption) return;
console.error();
console.error(" error: unknown option %s'", flag);
console.error();
process.exit(1);
};
```

Now:
```
program.unknownOption = function(flag) {
if (this._allowUnknownOption) return;
console.error();
console.error(" error: unknown option %s'", flag);
console.error();
// this is the extra code added to the function
this.outputHelp();
process.exit(1);
};
```

### DIFF
--- a/src/cli-validator/index.js
+++ b/src/cli-validator/index.js
@@ -6,7 +6,7 @@
 require('./utils/checkVersion')('8.9.0');
 require('./utils/updateNotifier');
 
-const program = require('commander');
+const program = require('./utils/modified-commander');
 const cliValidator = require('./runValidator');
 const version = require('../../package.json').version;
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1415,6 +1415,38 @@
         }
       }
     },
+    "@sinonjs/commons": {
+      "version": "1.7.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.7.0.tgz",
+      "integrity": "sha512-qbk9AP+cZUsKdW1GJsBpxPKFmCJ0T8swwzVje3qFd+AkQb74Q/tiuzrdfFg8AD2g5HH/XbE/I8Uc1KYHVYWfhg==",
+      "requires": {
+        "type-detect": "4.0.8"
+      }
+    },
+    "@sinonjs/formatio": {
+      "version": "4.0.1",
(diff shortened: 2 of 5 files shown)


src/cli-validator/index.js
```diff
@@ -6,7 +6,8 @@
 require('./utils/checkVersion')('8.9.0');
 require('./utils/updateNotifier');
 
-const program = require('commander');
+// include the modified commander module
+const program = require('./utils/unsupportedOptionError.js');
```
dpopp07: The only change we've made is for unsupported options, but conceptually that's not what we're importing. I think it's important to know that we're bring in the commander module, so I suggest changing your module name to something like `modified-commander`.

```suggestion
const program = require('./utils/modified-commander.js');
```

src/cli-validator/utils/unsupportedOptionError.js
```diff
@@ -0,0 +1,23 @@
+var program = require('commander');
```
dpopp07: You should never use `var` in modern JS. Use `const` for everything that won't change and `let` for any variable you will modify

```suggestion
const program = require('commander');
```

src/cli-validator/index.js
```diff
@@ -6,7 +6,7 @@
 require('./utils/checkVersion')('8.9.0');
 require('./utils/updateNotifier');
 
-const program = require('commander');
+const program = require('./utils/modified-commander.js');
```
dpopp07: The `.js` isn't necessary here, just a Node convention to not use it
```suggestion
const program = require('./utils/modified-commander');
```

src/cli-validator/utils/unsupportedOptionError.js
```diff
@@ -0,0 +1,23 @@
...
+  console.error("  error: unknown option `%s'", flag);
+  console.error();
+
+  // this is the extra code added to the function
+  this.outputHelp();
+
+  process.exit(1);
+};
+
+module.exports = program;
```
dpopp07: It would appear that you forgot to remove this file 😏

Issue #131, opened 2020-02-05
operationIds validation incorrectly handling sub-resources


bvanhav:
**Issue Type:** bug

**Validator Version:** v0.17.0

**Issue:** When I have a sub-resource in my spec (example path `/resource/{resource-id}/sub-resource`) the validator does not validate the operationId correctly. In this case I would expect a POST /resource/{resource-id}/sub-resource to allow an operationId of `create_resource_sub_resource` (or similar), but the validator raises a warning unless the operationId starts with `update` (which is technically not correct as this is a create of a child-resource).

There is a similar problem with listing sub-resources (validator requires operationId to start with `get` and won't allow `list`).

The following yaml should duplicate the problem:
```
components:
  parameters:
    resource_id:
      description: The resource identifier
      in: path
      name: resource_id
      required: true
      schema:
        type: string
paths:
  /resource/{resource_id}/child_resource:
    post:
      description: Create a child resource.
      operationId: create_resources_child_resource
      parameters:
      - $ref: '#/components/parameters/resource_id'
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/ChildResource'
        description: The child resource template
        required: true
      responses:
        "201":
          description: Created
      summary: Add child resource to a resource
      tags:
      - Child Resource
components:
  schemas:
    ChildResource:
      properties:
        name:
          description: Name of child resource
          type: string
          example: My Child Resource
      required:
      - name
      description: The template for the child resource
      type: object
      title: ChildResource
```

dpopp07:
@bvanhav Thanks for bringing this up, someone on our team will work on it soon.

SaltedCaramelCoffee:
This issue is fixed in https://github.com/IBM/openapi-validator/pull/133, closing.

PR #133, opened 2020-02-06
fix: allow create as a verb for post operations with path param

SaltedCaramelCoffee:
This is a PR in response to https://github.com/IBM/openapi-validator/issues/131

We should allow `create` as a verb for `POST` requests with path parameters. Validator will no longer flag this as a violation.

Also updated naming convention warning messages to include details about what the right operationId verb should be. The drawback is that now the `-s` summary function for openapi validator will not group all naming convention warnings together, since the summary grouping is done by grouping warnings or errors with the same message together.


src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -64,11 +64,13 @@ module.exports.validate = function({ resolvedSpec }) {
     // - paths with path param has a GET, a DELETE, and a POST or PUT or PATCH.
```
mkistler: The operationdId convention implemented here is not correct.  The rule should be:
- If the path does not _end_ in a path parameter:
  - `POST` should be "create" or "add"
  - "GET" should be "list"
- If the path _ends_ in a path parameter:
  - if the path has a `PATCH` operation
     - `PATCH` should be "update"
  - else if the the path has a "POST" operation
      - `POST` should be "update"
  - `PUT` should be "replace"
  - `DELETE` should be `delete`

So two important things to note:
1. You are only concerned with a path parameter _at the end_ of the path
1. You can't check operations in isolation -- you have to check all operations on the path together.

src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -57,47 +59,87 @@ module.exports.validate = function({ resolvedSpec }) {
...
+        // operationId for POST should start with "add" or "create" only if PATCH operation exist for this path
+        if (
+          allPathOperations.includes('patch') &&
+          !operationId.match(/^(add|create)[a-zA-Z0-9_]+/m)
+        ) {
+          checkPassed = false;
+          verbs.push('add');
+          verbs.push('create');
+
+          // If PATCH operation doesn't exist for path, POST operationId should start with "update", "add" or "update"
```
mkistler: If there is no `PATCH` operation, then we should only allow "update" for `POST`.

PR #139, opened 2020-02-13
feat: add warning for binary string in "application/json" or parameter

barrett-schonefeld:
Changes:
- added a findOctetSequencePaths function to handle the logic of finding schema type: string, format: binary for cases of nested arrays, objects, nested arrays of type object, objects with properties that are nested arrays, and objects with properties that are objects, and the simple case where a schema uses type: string, format: binary directly. This function takes a resolved schema and returns a list of paths to octet sequences (empty list if none found). The function accepts the path both as an array and a string and returns the path in the same format received.
- added logic to handle application/json request bodies that use schema type: string, format: binary
- added logic to handle application/json response bodies of type: string, format: binary
- added logic to handle parameters of type: string, format: binary
- removed 'binary' as a valid format for type: string parameters. parameters of type: string, format: binary will result in "type+format not well-defined" error
- added "json_or_param_binary_string" as to .defaultsForValidator as a warning in the shared.schemas section
- added "json_or_param_binary_string" configuration option to the README.md rules and defaults sections in the schemas section
- changed findOctetSequencePaths to accept the path both as a string and an array.

Refactoring in responses.js:
- moved the check for 204-response with content to be part of the other success code checks
- added a function to get all response codes, all status codes, and all success codes

Tests:
- added tests to ensure warnings are issued for request bodies, response bodies, and parameters with schema, type: string, format: binary
- added complex tests to exercise combinations of nested arrays and objects that contain schema type: string, format: binary (complex tests done on response bodies)

### DIFF
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -96,6 +96,9 @@ const defaults = {
       'no_response_codes': 'error',
       'no_success_response_codes': 'warning',
       'no_response_body': 'warning'
+    },
+    'schemas': {
+      'json_or_param_binary_string': 'warning'
     }
   }
 };
--- a/src/plugins/validation/oas3/semantic-validators/operations.js
+++ b/src/plugins/validation/oas3/semantic-validators/operations.js
@@ -4,15 +4,21 @@
 // Assertation 2. Operations with non-form request bodies should set the `x-codegen-request-body-name`
 // annotation (for code generation purposes)
 
+// Assertation 3. Request bodies with application/json content should not use schema
+// type: string, format: binary.
+
 const pick = require('lodash/pick');
 const each = require('lodash/each');
 const { hasRefProperty } = require('../../../utils');
+const findOctetSequencePaths = require('../../../utils/findOctetSequencePaths')
+  .findOctetSequencePaths;
 
 module.exports.validate = function({ resolvedSpec, jsSpec }, config) {
   const result = {};
   result.error = [];
   result.warning = [];
 
+  const configSchemas = config.schemas;
   config = config.operations;
 
   const REQUEST_BODY_NAME = 'x-codegen-request-body-name';
@@ -80,6 +86,28 @@ module.exports.validate = function({ resolvedSpec, jsSpec }, config) {
                 path: `paths.${pathName}.${opName}`,
                 message
               });
+            }
+          }
+
+          // Assertation 3
+          const binaryStringStatus = configSchemas.json_or_param_binary_string;
+          if (binaryStringStatus !== 'off') {
+            for (const mimeType of requestBodyMimeTypes) {
+              if (mimeType === 'application/json') {
+                const schemaPath = `paths.${pathName}.${opName}.requestBody.content.${mimeType}.schema`;
+                const octetSequencePaths = findOctetSequencePaths(
+                  requestBodyContent[mimeType].schema,
+                  schemaPath
+                );
+                const message =
+                  'JSON request/response bodies should not contain binary (type: string, format: binary) values.';
+                for (const p of octetSequencePaths) {
+                  result[binaryStringStatus].push({
+                    path: p,
+                    message
+                  });
+                }
+              }
             }
           }
         }
--- a/src/plugins/validation/oas3/semantic-validators/parameters.js
+++ b/src/plugins/validation/oas3/semantic-validators/parameters.js
@@ -7,13 +7,20 @@
 
 // https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameterObject
 
+// Assertation 3:
+// A paramater should not use schema type: string, format: binary because there is now well-
+// defined way to encode an octet sequence in a URL.
+
 const { isParameterObject, walk } = require('../../../utils');
+const findOctetSequencePaths = require('../../../utils/findOctetSequencePaths')
+  .findOctetSequencePaths;
 
 module.exports.validate = function({ jsSpec }, config) {
   const result = {};
   result.error = [];
   result.warning = [];
 
+  const configSchemas = config.schemas;
   config = config.parameters;
 
   walk(jsSpec, [], function(obj, path) {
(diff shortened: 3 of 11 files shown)


src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -84,7 +93,49 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
+          const octetSequencePaths = octetSequences(
+            responseBodyContent[mimeType].schema,
+            schemaPath
+          );
+          const message =
+            'No well-defined way to represent bare octets (type: string, format: binary) as JSON.';
+          for (const p of octetSequencePaths) {
+            result[binaryStringStatus].push({
+              path: p.split('.'),
+              message: message
```
mkistler: There seems to be contention in how to represent the path here.  We need to join it to call `octetSequencePaths`, but then split it again to set result.  Can't we just use one format consistently?

test/plugins/validation/oas3/operations.js
```diff
@@ -191,6 +250,16 @@ describe('validation plugin - semantic - operations - oas3', function() {
...
+    const config = {
+      operations: {
+        no_request_body_content: 'error',
+        no_request_body_name: 'warning'
+      },
+      schemas: {
+        binary_string: 'warning'
+      }
+    };
+
```
mkistler: It seems odd that we would have to add this config for existing tests.  Why is this?

test/plugins/validation/oas3/parameters.js
```diff
@@ -143,6 +179,18 @@ describe('validation plugin - semantic - parameters - oas3', function() {
...
+        no_in_property: 'error',
+        invalid_in_property: 'error',
+        missing_schema_or_content: 'error',
+        has_schema_and_content: 'error'
+      },
+      schemas: {
+        binary_string: 'warning'
+      }
+    };
+
```
mkistler: Again ... why are these needed when they weren't needed previously?

src/plugins/validation/oas3/semantic-validators/responses.js
barrett-schonefeld: Some of the semantic validators use arrays and others use strings with `.` as the delimiter. For this PR I just matched the convention of each validator, but I would like to standardize the paths to use arrays in the next PR.

src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -84,7 +93,49 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
+}
+
+function isStatusCode(code) {
   const allowedFirstDigits = ['1', '2', '3', '4', '5'];
   return code.length === 3 && allowedFirstDigits.includes(code.slice(0, 1));
 }
+
+function validateResponseBody(
+  responseObj,
+  allResponseCodes,
```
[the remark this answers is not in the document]
barrett-schonefeld: Yes it is.

test/plugins/validation/oas3/operations.js
barrett-schonefeld: Since this binary string rule spans multiple semantic validators, we have to get the rule from the shared schema rules with `const configSchemas = config.schemas`.

The parameters, responses, and operations validators previously only used the OAS3-specific rules, but the way it is implemented now requires the shared schema rules to be present. Later it is used:

`const binaryStringStatus = configSchemas.binary_string;
  if (binaryStringStatus !== 'off') {`

I could wrap this with:

`if (configSchemas) {
    const binaryStringStatus = configSchemas.binary_string;`

test/plugins/validation/oas3/parameters.js
barrett-schonefeld: (See above)

src/plugins/validation/oas3/semantic-validators/parameters.js
```diff
@@ -66,8 +73,39 @@ module.exports.validate = function({ jsSpec }, config) {
...
+            'Parameters should not contain binary (type: string, format: binary) values.';
+          result[binaryStringStatus].push({
+            path: p.split('.'),
+            message: message
+          });
+        }
+      }
     }
   });
-
```
dpopp07: No need to remove this whitespace

test/plugins/validation/oas3/operations.js
```diff
@@ -2,10 +2,19 @@ const expect = require('expect');
 const {
   validate
 } = require('../../../../src/plugins/validation/oas3/semantic-validators/operations');
-const config = require('../../../../src/.defaultsForValidator').defaults.oas3;
```
dpopp07: Wait.. why have you removed this line? This is meant to handle the config so that you don't have to define it for every independent test. It should work for your use case

test/plugins/validation/oas3/operations.js
dpopp07: I now see that this answers my comment above. I think we might want to re-consider the approach here. Possibly define separate rules for each section to give more fine-grained control. We can discuss this with @mkistler

src/plugins/validation/oas3/semantic-validators/operations.js
```diff
@@ -78,10 +84,32 @@ module.exports.validate = function({ resolvedSpec, jsSpec }, config) {
...
+                const octetSequencePaths = octetSequences(
+                  requestBodyContent[mimeType].schema,
+                  schemaPath
+                );
+                const message =
+                  'No well-defined way to represent bare octets (type: string, format: binary) as JSON.';
+                for (const p of octetSequencePaths) {
+                  result[binaryStringStatus].push({
+                    path: p,
+                    message: message
```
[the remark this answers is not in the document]
dpopp07: I only commented on a few, but all of these instances need to be fixed

src/plugins/validation/oas3/semantic-validators/operations.js
```diff
@@ -78,10 +84,32 @@ module.exports.validate = function({ resolvedSpec, jsSpec }, config) {
                 'Operations with non-form request bodies should set a name with the x-codegen-request-body-name annotation.';
               result[checkStatus].push({
                 path: `paths.${pathName}.${opName}`,
-                message
+                message: message
```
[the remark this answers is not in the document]
barrett-schonefeld: Other logging in the module defines the key, so I added for consistency.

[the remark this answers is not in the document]
dpopp07: I see. In that case, I would recommend you change the rest of them to match what this was before

test/plugins/validation/oas3/parameters.js
```diff
@@ -224,7 +284,219 @@ describe('validation plugin - semantic - parameters - oas3', function() {
...
+      schemas: {
+        json_or_param_binary_string: 'warning'
+      }
+    };
+
+    const spec = {
+      components: {
+        parameters: {
+          GoodParam: {
+            schema: {}
```
dpopp07: I would actually recommend making this test about the schema field not even being there. That's often a good thing to verify in JS code. Just make `GoodParam` an empty object.

src/.defaultsForValidator.js
```diff
@@ -64,7 +64,8 @@ const defaults = {
       'description_mentions_json': 'warning',
       'array_of_arrays': 'warning',
       'property_case_convention': [ 'error', 'lower_snake_case'],
-      'enum_case_convention': [ 'error', 'lower_snake_case']
+      'enum_case_convention': [ 'error', 'lower_snake_case'],
+      'binary_string': 'warning'
```
dpopp07: Don't forget - this needs to match what is in the README/code

test/plugins/validation/oas3/operations.js
barrett-schonefeld: Added a 'schemas' section to the OAS3 rules and put the json_or_param_binary_string config option there.

Then removed the individual configuration objects in each test and replaced it by importing the default OAS3 rules.

test/plugins/validation/oas3/parameters.js
```diff
@@ -1,4 +1,5 @@
 const expect = require('expect');
+
```
dpopp07: Don't need to add this whitespace

PR #140, opened 2020-02-14
feat: catch properties with the same name but different type.

presto21:
Properties with the same name but different type are allowed in an API definition. However, that would not constitute good, consistent design within an API. Therefore, the validator should catch properties that follow this pattern and produce a warning. 

Change:
- Modified schema-ibm.js to find properties with the same name but different type in the API definition.
- added an API definition and a test to exercise the new behavior.

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
@@ -100,6 +100,10 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
     }
   });
 
+  // Collect the properties info. Pass it to checkProperties to catch
+  // properties that have the same name but an inconsistent type.
+  const propertiesToCompare = {};
+
   schemas.forEach(({ schema, path }) => {
     generateFormatErrors(schema, path, config, isOAS3, messages);
 
@@ -136,6 +140,15 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
           );
         }
       }
+    }
+    if (config.inconsistent_property_type !== 'off') {
+      checkProperties(
+        schema,
+        propertiesToCompare,
+        path,
+        config.inconsistent_property_type,
+        messages
+      );
     }
   });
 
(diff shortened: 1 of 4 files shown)


src/cli-validator/utils/duplicateProperties.js
```diff
@@ -0,0 +1,123 @@
...
+  const arr = [];
+
+  // Parsing through the splits and store key and values
+  // that are present after we encounter "properties"
+  for (let i = 0; i < splits.length; i++) {
+    if (splits[i].includes('properties')) {
+      let tempKey = splits[i];
+      tempKey = tempKey.replace(':', '');
+      tempKey = tempKey.replace(/"|'/g, '');
+      tempKey = tempKey.trim();
```
dpopp07: I feel like this stuff is unnecessary - we should be able to parse through a JSON object rather than a string

test/cli-validator/tests/expectedOutput.js
```diff
@@ -263,4 +263,44 @@ describe('test expected output - OpenAPI 3', function() {
...
+    );
+    expect(capturedText[4].trim()).toEqual('Line    :   124');
+    expect(capturedText[6].trim()).toEqual(
+      'Message :   Properties name and name have the same name but different types'
+    );
+    expect(capturedText[7].trim()).toEqual(
+      'Path    :   components.schemas.kid.properties.name'
+    );
+    expect(capturedText[8].trim()).toEqual('Line    :   179');
+  });
```
dpopp07: Since your feature is a validation, it should not be tested here. It should have its own unit test. See the other plugin tests for examples

test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -1417,4 +1420,44 @@ describe('validation plugin - semantic - schema-ibm - OpenAPI 3', () => {
...
+    );
+    expect(capturedText[4].trim()).toEqual('Line    :   124');
+    expect(capturedText[6].trim()).toEqual(
+      'Message :   Properties name and name have the same name but different types'
+    );
+    expect(capturedText[7].trim()).toEqual(
+      'Path    :   components.schemas.kid.properties.name'
+    );
+    expect(capturedText[8].trim()).toEqual('Line    :   179');
+  });
```
dpopp07: Since you've written the code as a validator, you need to use the existing testing framework for validators. The process of capturing text is only necessary for testing CLI-tool-specific behaviors. Use the other tests for examples of what this should look like

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -411,3 +415,29 @@ function isRootSchema(path) {
...
+    for (const [key, value] of entries) {
+      // skips properties that use $ref
+      if (value.type != undefined) {
+        if (propDict.hasOwnProperty(key)) {
+          if (propDict[key] != value.type) {
+            path += `.properties.${key}`;
+            path = path.replace(/,/g, '.');
+            messages.addMessage(
+              path,
+              `Properties ${key} and ${key} have the same name but different types`,
```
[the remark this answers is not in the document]
dpopp07: I know I gave you that message - I am just now thinking we could potentially make it more helpful. If we can't think of anything, this is fine 👍

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -411,3 +415,29 @@ function isRootSchema(path) {
...
+              'warning'
```
dpopp07: I just realized that I forgot to comment on this - this check should be configurable

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -511,3 +524,44 @@ function isRootSchema(path) {
...
+  contextPath,
+  configOption,
+  messages
+) {
+  if (!schema.properties) return;
+
+  for (const [key, value] of Object.entries(schema.properties)) {
+    // skips properties that use $ref
+    if (value.type) {
+      if (Object.prototype.hasOwnProperty.call(propertiesToCompare, key)) {
```
dpopp07: This isn't necessary in this scenario. You simplify it like this:

```suggestion
      if (propertiesToCompare.key) {
```

PR #143, opened 2020-02-19
feat: error for oneOf, anyOf, allOf schema that does not use array

barrett-schonefeld:
Changes:
- added default error for oneOf, anyOf, or allOf schema that is not use top-level array

Tests:
- added tests when oneOf, anyOf, or allOf schemas are nested ("oneOf allOf," for example), when oneOf, anyOf, or allOf schema is used as a property in an obect, and when oneOf, anyOf, or allOf schema used as items in array
- updated schema-ibm test to import the default config values and made necessary changes to test to account for unexpected errors and warnings resulting from having all default values present
- added test to ensure errors not issued when compose models use an array

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,7 +430,6 @@ The default values for each rule are described below.
 | array_of_arrays             | warning |
 | property_case_convention    | error, lower_snake_case |
 | enum_case_convention        | error, lower_snake_case |
-| json_or_param_binary_string | warning |
 
 ###### walker
 | Rule                          | Default |
--- a/README.md
+++ b/README.md
@@ -430,6 +430,7 @@ The default values for each rule are described below.
 | array_of_arrays             | warning |
 | property_case_convention    | error, lower_snake_case |
 | enum_case_convention        | error, lower_snake_case |
+| json_or_param_binary_string | warning |
 
 ###### walker
 | Rule                          | Default |
--- a/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
@@ -50,14 +50,33 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
             This includes responses, request bodies, parameters (with content rather than schema),
             both at the operation level and within the top-level "components" object
     */
+
     const modelLocations = ['definitions', 'schemas', 'properties'];
     if (
       current === 'schema' ||
       current === 'items' ||
-      modelLocations.indexOf(path[path.length - 2]) > -1
+      modelLocations.includes(path[path.length - 2])
     ) {
       const pushLeafSchemas = (obj, path) => {
-        if (obj.allOf && Array.isArray(obj.allOf)) {
+        if (obj.allOf && !Array.isArray(obj.allOf)) {
+          messages.addMessage(
+            path.concat(['allOf']),
+            'allOf value should be an array',
+            'error'
+          );
+        } else if (isOAS3 && obj.oneOf && !Array.isArray(obj.oneOf)) {
+          messages.addMessage(
+            path.concat(['oneOf']),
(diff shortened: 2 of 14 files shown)


test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -1005,8 +882,152 @@ describe('validation plugin - semantic - schema-ibm - OpenAPI 3', () => {
...
+    expect(res.errors[2].message).toEqual('anyOf should be an array');
+    expect(res.errors[2].path).toEqual([
+      'components',
+      'schemas',
+      'any_of_array',
+      'items',
+      'anyOf'
+    ]);
+  });
+
```
dpopp07: Will you add one positive test? It can just be one compose model which correct usage in one of the files

README.md
```diff
@@ -251,6 +251,7 @@ The supported rules are described below:
 | no_property_description     | Flag any schema that contains a 'property' without a `description` field.     | shared   |
 | description_mentions_json   | Flag any schema with a 'property' description that mentions the word 'JSON'.  | shared   |
 | array_of_arrays             | Flag any schema with a 'property' of type `array` with items of type `array`. | shared   |
+| compose_model_field_not_array | Flag any compose model schema that does not use an array.                   | shared   |
```
mkistler: If we keep this configurable, I think we want a clearer name and description.  "compose model" is a term from the SDK generator -- not used in OpenAPI or JSON Schema.  So better to use something like "allOf_value_not_array" and "Flag any `allOf` schema attribute whose value is not an array".

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -50,14 +50,33 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
...
       current === 'items' ||
-      modelLocations.indexOf(path[path.length - 2]) > -1
+      modelLocations.includes(path[path.length - 2])
     ) {
       const pushLeafSchemas = (obj, path) => {
-        if (obj.allOf && Array.isArray(obj.allOf)) {
+        if (obj.allOf && !Array.isArray(obj.allOf)) {
+          messages.addMessage(
+            path.concat(['allOf']),
+            'allOf should be an array',
```
mkistler:
```suggestion
            'allOf value should be an array',
```

test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -984,12 +867,6 @@ describe('validation plugin - semantic - schema-ibm - OpenAPI 3', () => {
   });
 
   it('should process allOf, oneOf, anyOf schemas correctly', () => {
-    const config = {
-      schemas: {
-        invalid_type_format_pair: 'error'
-      }
-    };
-
```
mkistler: All the changes above here appear to be unrelated to this feature.

barrett-schonefeld: Yes, why I was introducing tests to this module, I thought it would be a good idea to clean up the individual declarations of config and replace them with one import. I changed the config name to customConfig for the cases where the test uses a custom value.

README.md
barrett-schonefeld: I chose to use compose model because this PR checks that `allOf`, `anyOf, and `oneOf` use arrays. In my next changes, I am going to remove the configuration option, but I want to make sure you agree with the choice to include `anyOf` and `oneOf`.

test/plugins/validation/2and3/schema-ibm.js
dpopp07: ^ I support these changes

README.md
mkistler: Yes -- we should check all three.  I made my suggestion before I saw the code below.

PR #144, opened 2020-02-19
feat: flag definitions that are defined but never used

barrett-schonefeld:
Changes:
- added checks for parameters, responses, and other definitions in the components section that are not used in the API definition
- updated comments in refs.js
- restructured refs.js as needed to support checking components.parameters and components.responses

Tests:
- added tests to that ensure that a warning is issued when a response or parameter definition is not used
- added tests to ensure a warning is not issued when a response or parameter definition is used

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/refs.js
+++ b/src/plugins/validation/2and3/semantic-validators/refs.js
@@ -15,12 +15,7 @@ module.exports.validate = function({ jsSpec, specStr, isOAS3 }) {
     return messages;
   }
 
-  const basePath = isOAS3 ? ['components', 'schemas'] : ['definitions'];
-
-  // Assertation 1
-  // This is a "creative" way to approach the problem of collecting used $refs,
-  // but other solutions required walking the jsSpec recursively to detect $refs,
-  // which can be quite slow.
+  // Collects all refs in the API definition
   const refRegex = /\$ref.*["'](.*)["']/g;
   let match = refRegex.exec(specStr);
   const refs = [];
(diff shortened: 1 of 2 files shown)


src/plugins/validation/2and3/semantic-validators/refs.js
```diff
@@ -15,12 +15,8 @@ module.exports.validate = function({ jsSpec, specStr, isOAS3 }) {
     return messages;
   }
 
-  const basePath = isOAS3 ? ['components', 'schemas'] : ['definitions'];
-
   // Assertation 1
-  // This is a "creative" way to approach the problem of collecting used $refs,
-  // but other solutions required walking the jsSpec recursively to detect $refs,
-  // which can be quite slow.
```
dpopp07: Also remove this comment? Also, I don't think the line `// Assertation 1` was supposed to be right above this - will you remove it?

src/plugins/validation/2and3/semantic-validators/refs.js
```diff
@@ -29,21 +25,52 @@ module.exports.validate = function({ jsSpec, specStr, isOAS3 }) {
...
-    startsWith(v, `#/${basePath.join('/')}`)
+    startsWith(v, `#/${definitionSectionName}`)
   );
 
-  const definitions = isOAS3 ? jsSpec.components.schemas : jsSpec.definitions;
-  each(definitions, (def, defName) => {
-    if (definitionsRefs.indexOf(`#/${basePath.join('/')}/${defName}`) === -1) {
+  // checks if the definitions are used, and if not, record a warning
+  if (isOAS3) {
+    const definitionTypeList = ['schemas', 'parameters', 'responses'];
```
dpopp07: I like the way you're doing this. Since it's so simple to check the different locations - we might as well look at all of them! Here is the [full list](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#components-object)

src/plugins/validation/2and3/semantic-validators/refs.js
```diff
@@ -29,21 +25,52 @@ module.exports.validate = function({ jsSpec, specStr, isOAS3 }) {
...
+        ['definitions'],
+        messages
+      );
+    }
+  }
+
+  return messages;
+};
+
+function definitionNotUsedLogger(
```
dpopp07: I know you like this word, but I'm going to request we don't call this a "logger" - only because at some point I want to add a _real_ logger to this project 🙂 

Maybe use something like `checkIfDefinitionNotUsed`

src/plugins/validation/2and3/semantic-validators/refs.js
barrett-schonefeld: I've seen that examples are often excluded from validation. walk.js, for example, does not walk down keys named 'example' or 'examples.'

Should I include or exclude examples from the full list?

src/plugins/validation/2and3/semantic-validators/refs.js
barrett-schonefeld: Hahah! Had writer's block initially and meant to change it

src/plugins/validation/2and3/semantic-validators/refs.js
dpopp07: That's a good question. I think still include it. No, we can't validate examples, but we can still let a user know if they've defined one they aren't using

PR #145, opened 2020-02-21
feat: list valid types and formats in error messages

### LINKED ISSUE #117: Request for a higher order error for bad "type"
Am running v0.15.2 installed inside a Node 13.2.0 alpine Docker container.

Given the following openapi v3 spec:

```yaml
openapi: 3.0.0
info:
  description: Test types
  version: '0'
  title: Thing service serves things
servers:
  - url: 'https://example.com/test'
paths:
  '/thing':
    get:
      summary: Get thing
      operationId: get_thing
      responses:
        '200':
          description: OK.
          content:
            application/json:
              schema:
                type: object
                properties:
                  thing:
                    type: thing
```

There is a problem on the last line `type: thing`.

## Current behaviour

When editing this file in the latest Swagger editor at http://editor.swagger.io/ the following error is shown:

Screenshot from 2019-11-26 12-44-55

And the validator *does* return an error, which is good. This is the error returned:

```
errors

  Message :   Property type+format is not well-defined.
  Path    :   paths./things.get.responses.200.content.application/json.schema.properties.things.type
  Line    :   23
```

There are two problems with this current behaviour of the validator:

1. In this case, `format` has not been defined, so it's a confusing error message. My guess is that there is no value we could add for `format` that would fix this error.

2. The real error which is that the `type` of `thing` is invalid is not being returned, so the validator is not helping the user "get back on track".

## Desired behaviour

Would it be possible to generate a new higher order error? In this case something that matches what the swagger editor returns?

```
errors

  Message :   Property type is not equal to one of the allowed values.
  Path    :   paths./things.get.responses.200.content.application/json.schema.properties.things.type
  Line    :   23
```

I think that would be helpful because it seems to me that the root cause of the problem is that the `type` is not one of the allowed values: "array, boolean, integer, number, object, string", rather than there being an issue with `format`, which is unspecified in this example.

Having a new type error would mean more control available in the config file because `type+format` errors could be tuned down for files that are using custom `format` such as "uuid", without stomping on important `type` errors like this one.

barrett-schonefeld:
Changes:
- changed function name formatValid to typeFormatErrors
- passed additional arguments (messages and checkStatus) to be able to record errors within the function
- negated the boolean conditions used to determine if the type+format is valid and (instead of using the boolean logic to return the validity of the schema) used the negated logic to record errors
- added descriptive messages to help resolve the type+format errors
- removed the duplicate type+format validation function in parameters-ibm.js
- added parameters to the list of schemas to validate
- moved validation for swagger2 file parameters into 'file' section of typeFormatErrors. I believe this is the only validation that was done in parameters-ibm that is not handled by typeFormatErrors.

Tests:
- added a test to ensure file params without in: formData produces an error
- added a negative test to ensure no error produced file params that use in: form data
- added tests to ensure an error produced when a format paired with type: array or type: object
- added negative test to ensure no error produced when object properties and array items use valid type, format pairs
- added test to ensure an error produced when schema defines a format but not a type

Resolves #117

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
@@ -50,8 +50,12 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
             This includes responses, request bodies, parameters (with content rather than schema),
             both at the operation level and within the top-level "components" object
     */
-
-    const modelLocations = ['definitions', 'schemas', 'properties'];
+    const modelLocations = [
+      'definitions',
+      'schemas',
+      'properties',
+      'parameters'
+    ];
     if (
       current === 'schema' ||
       current === 'items' ||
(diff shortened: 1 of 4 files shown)


src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -139,60 +144,135 @@ function generateFormatErrors(schema, contextPath, config, isOAS3, messages) {
...
+          checkStatus
+        );
+      }
       break;
     default:
-      valid = false;
+      // invalid type
+      messages.addMessage(
+        path.concat(['type']),
+        'Invalid type. Valid types are: integer, number, string, boolean, object, array.',
```
barrett-schonefeld: Note: I did not include the `file` type in this list. I could make the error message conditional (using isOAS3) or add `file` to the list with the comment, "(swagger2 only)." However, it seems the `file` type is not well-defined, so I was skeptical of adding it to this list.

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -139,60 +144,135 @@ function generateFormatErrors(schema, contextPath, config, isOAS3, messages) {
...
+          'File type only valid for swagger2 and must be used as root schema.',
+          checkStatus
+        );
+      } else if (obj.in && obj.in !== 'formData') {
+        messages.addMessage(
+          path.concat(['type']),
+          'File type parameter must use in: formData.',
+          checkStatus
+        );
+      }
```
dpopp07: This may not be a big deal since it's likely a small use case, but "format" shouldn't be specified for type "file" Since you check for that for some types above, it might make sense to do that here. Not a requirement

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
dpopp07: Thanks for pointing this out, that's a fair concern. I think the way to go is a conditional list using `isOAS3` that includes "file" at the end

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
barrett-schonefeld: Yes, I will add this.

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -50,11 +50,16 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
             This includes responses, request bodies, parameters (with content rather than schema),
             both at the operation level and within the top-level "components" object
     */
-    const modelLocations = ['definitions', 'schemas', 'properties'];
+    const modelLocations = [
+      'definitions',
+      'schemas',
+      'properties',
+      'parameters'
```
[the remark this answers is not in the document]
barrett-schonefeld: Yes, that is exactly right. This module assumed that responsibility when the parameter validation was removed from parameters-ibm.js.

test/plugins/validation/2and3/schema-ibm.js
```diff
@@ -874,12 +1118,11 @@ describe('validation plugin - semantic - schema-ibm - OpenAPI 3', () => {
...
     expect(res.errors.length).toEqual(2);
     expect(res.errors[0].message).toEqual(
-      'Property type+format is not well-defined.'
+      'Invalid type. Valid types are: integer, number, string, boolean, object, array.'
     );
     expect(res.errors[1].message).toEqual(
-      'Property type+format is not well-defined.'
+      'Schema of type string should use one of the following formats: byte, binary, date, date-time, password.'
     );
-    expect(res.warnings.length).toEqual(0);
```
dpopp07: Why remove this line?

barrett-schonefeld: On some of my previous PRs, we talked about removing the additional checks that were not directly related to the test. When I've seen this in other tests, I've removed the checks to prevent unexpected test failures in the future due to adding a validation rule to a module.

dpopp07: I can't argue with that. Carry on 🙂

PR #146, opened 2020-02-24
feat: catch inline parameters that appear on multiple operations

presto21:
To encourage "DRY" style in API documents, check for inline parameters that appear on multiple operations. Such parameters should be defined in the Parameters section and then referenced from any operation that includes them.

Change:

- added a function to parameters-ibm.js to find inline parameter that appear on multiple operations.
- added an API definition and a test to exercise the new behavior

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
@@ -8,6 +8,10 @@
 // Header parameters must not define a content-type or an accept-type.
 // http://watson-developer-cloud.github.io/api-guidelines/swagger-coding-style#do-not-explicitly-define-a-content-type-header-parameter
 
+// Assertation 4:
+// Parameters that are defined inline to an operation but appear on multiple operations should be defined
+// in the Parameters section (once) and then referenced from any operation that includes them.
+
 const pick = require('lodash/pick');
 const includes = require('lodash/includes');
 const { checkCase, isParameterObject, walk } = require('../../../utils');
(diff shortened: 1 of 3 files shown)


src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -182,3 +192,17 @@ function formatValid(obj, isOAS3) {
   }
   return false;
 }
+
+function checkInlineParameters(obj, isRef, parameterNames, messages, path) {
+  if (!isRef) {
+    if (!parameterNames.includes(obj.name)) {
```
dpopp07: I don't think that only checking the name is sufficient for recommending a referenced parameter. If these parameters have the same name but are different (which is maybe a different issue), they cannot be consolidated into a referenced parameter. This check should assert a stricter level of equality before determining there is an issue.

test/plugins/validation/2and3/parameters-ibm.js
```diff
@@ -609,5 +612,37 @@ describe('validation plugin - semantic - parameters-ibm', () => {
...
+
+      expect(capturedText[1].trim()).toEqual('warnings');
+      expect(capturedText[2].trim()).toEqual(
+        'Message :   Inline parameters that appear on multiple operations should be defined in the Parameters section and then referenced.'
+      );
+      expect(capturedText[3].trim()).toEqual(
+        'Path    :   components.parameters.Id'
+      );
+      expect(capturedText[4].trim()).toEqual('Line    :   90');
+    });
```
dpopp07: This test should not be running the whole program or capturing any output. It should just be a unit test for the single `parameters-ibm` validator. You can use the other tests as examples

PR #147, opened 2020-02-24
feat: command line option to print config option associated with each message

### LINKED ISSUE #44: Add optional reporting of failing validation
lint-openapi has a "--print_validator_modules" which prints the name of the module that contains the failing validation,  but this information is of limited value because the information is an implementation detail that is invisible to a user of the tool from the command line.  It would be nice to have another option that resulted in printing the actual name of the config option that resulted in the validation failure so that users can understand what config option controls that result.

Today I see this for example: 

```bash
$ lint-openapi ./api-validator-openapi-2x.yml
warnings
  Message :   Response schemas should be defined with a named ref.
  Path    :   paths./validation-jobs/{job_id}/junit.get.responses.200.schema
  Line    :   176
```
It would be ideal to have something like this:
```bash
$ lint-openapi  --print_validator_name ./api-validator-openapi-2x.yml
warnings
  Message  :  Response schemas should be defined with a named ref.
  Validator  :  inline_response_schema
  Path         :  paths./validation-jobs/{job_id}/junit.get.responses.200.schema
  Line          :  176
```
(sorry about the alignment above, dang markdown and non-fixed-width font)

It is true that the messages are directly correlated to the message, but the only way to do this now that I see is building a map of messages to the validator config names, which seems like a fragile and non scalable way to do it.

I am happy to help with implementation by submitting a PR.

barrett-schonefeld:
Changes:
- updated addMessage in the MessageCarrier to accept an optional rule parameter (default value is 'builtin')
- updated all tests that check for equality of error/warning objects rather than the individual message/path values. Necessary because these tests did not expect a 'rule' field in this object.
- updated each semantic validator to pass the config option name (or if no config option, default value of 'builtin' is used)
- added the --print_rule_names (or -r) command-line option to allow users to print the rule names
- updated the function that prints the results to print the rule if this print_rule_names option is used

Tests:
- added a test that captures the output and ensures there is a rule for each error/warning

Example Output:
Screen Shot 2020-02-24 at 4 15 34 PM

Resolves #44

### DIFF
--- a/src/cli-validator/index.js
+++ b/src/cli-validator/index.js
@@ -22,6 +22,10 @@ program
     'print the validators that catch each error/warning (helpful for development)'
   )
   .option(
+    '-r, --print_rule_names',
+    'print the configuration option names associated with each error and warning'
+  )
+  .option(
     '-n, --no_colors',
     'turn off output coloring'
   )
--- a/src/cli-validator/runValidator.js
+++ b/src/cli-validator/runValidator.js
@@ -32,6 +32,7 @@ const processInput = async function(program) {
 
   // interpret the options
   const printValidators = !!program.print_validator_modules;
+  const printRuleNames = !!program.print_rule_names;
   const reportingStats = !!program.report_statistics;
 
   const turnOffColoring = !!program.no_colors;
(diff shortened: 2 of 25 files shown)


mkistler:
Closing this PR as it has been incorporated into #222.

Issue #148, opened 2020-02-26
Configurable operationId naming convention


fenuks:
#124 adds linting naming conventions for `operationId`, but apparently, there is no option to disable it, or change which verbs are in convention, e.g. in my case I would like to use `search` verb instead of `list` when GET endpoint has parameters, since in my API, it's always a search operation.

openapi-validator version: 0.24.0

dpopp07:
@fenuks There should certainly be a way to disable it. If that is not currently implemented, we will fix that.

As for custom verbs, that would require more effort and we may not be able to prioritize that at the moment. We can look into it though.

fenuks:
Thank you, of course it's understandable that custom verbs is not something that is highly needed, especially that current ones work for you.

As for disabling this rule, I believe there is no possibility now, there is no mention in documentation about it, beside that I quickly skimmed through code, and to my understanding, there is no switch to turn checking off.

dpopp07:
I confirmed that you are correct - there is no way to configure this option. This needs to be fixed soon. @barrett-schonefeld @presto21 will one of you take on this issue?

PR #149, opened 2020-02-27
fix: config option for operationId naming convention

presto21:
Naming conventions for operationId should be configurable. 

Change:
- added config option to README.md and .defaultsForValidator.js
- added config logic to operation-ids.js 
- added a test case to exercise the new behavior

Related: [Issue 148](https://github.com/IBM/openapi-validator/issues/148)

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ The supported rules are described below:
 | no_summary                   | Flag any operations that do not have a `summary` field.                             | shared   |
 | no_array_responses           | Flag any operations with a top-level array response.                                | shared   |
 | parameter_order              | Flag any operations with optional parameters before a required param.               | shared   |
+| operation_id_naming_convention | Flag any `operationId` that does not follow consistent naming convention.         | shared   |
 | no_request_body_content      | [Flag any operations with a `requestBody` that does not have a `content` field.][3] | oas3     |
 | no_request_body_name         | Flag any operations with a non-form `requestBody` that does not have a name set with `x-codegen-request-body-name`. | oas3|
 
@@ -379,6 +380,7 @@ The default values for each rule are described below.
 | no_summary                   | warning |
 | no_array_responses           | error   |
 | parameter_order              | warning |
+| operation_id_naming_convention | warning |
 
 ###### pagination
 | Rule                        | Default |
--- a/src/plugins/validation/2and3/semantic-validators/operation-ids.js
+++ b/src/plugins/validation/2and3/semantic-validators/operation-ids.js
@@ -8,8 +8,10 @@ const merge = require('lodash/merge');
 const each = require('lodash/each');
 const MessageCarrier = require('../../../utils/messageCarrier');
 
-module.exports.validate = function({ resolvedSpec }) {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -379,6 +379,7 @@ The default values for each rule are described below.
 | no_summary                   | warning |
 | no_array_responses           | error   |
 | parameter_order              | warning |
+| operationId_naming_convention| warning |
```
dpopp07: Rule names are meant to be snake case, we should not use mixed case

```suggestion
| operation_id_naming_convention | warning |
```

src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -153,15 +155,18 @@ module.exports.validate = function({ resolvedSpec }) {
...
+          if (checkPassed === false) {
+            messages.addMessage(
+              op.path + '.operationId',
+              `operationIds should follow consistent naming convention. operationId verb should be ${verbs}`.replace(
+                ',',
+                ' or '
+              ),
+              checkStatus
+            );
+          }
```
dpopp07: A couple of comments - the config handler will automatically fill in all rule values with "error", "warning", or "off", so there's no need for the `|| 'off'`.

Also, much of this logic can be simplified by just passing the key directly to the message carrier - it has built-in checking for whether something is an error, warning, or off:

```suggestion
        if (checkPassed === false) {
          messages.addMessage(
            op.path + '.operationId',
            `operationIds should follow consistent naming convention. operationId verb should be ${verbs}`.replace(
              ',',
              ' or '
            ),
            config.operation_id_naming_convention
          );
        }
```

test/plugins/validation/2and3/operation-ids.js
```diff
@@ -164,14 +171,18 @@ describe('validation plugin - semantic - operation-ids', function() {
...
     expect(res.errors.length).toEqual(0);
     expect(res.warnings.length).toEqual(12);
     expect(res.warnings[0].path).toEqual('paths./books.get.operationId');
     expect(res.warnings[0].message).toEqual(
       'operationIds should follow consistent naming convention. operationId verb should be list'
     );
+
+    const customRes = validate({ resolvedSpec }, customConfig);
+    expect(customRes.errors.length).toEqual(0);
+    expect(customRes.warnings.length).toEqual(0);
```
dpopp07: I appreciate the thought here, but these added checks are unnecessary. We can assume the message carrier logic works as it has it's own tests.

Also, just for future reference, it is best practice to test different behaviors in entirely different tests (different `it` functions), rather than adding checks for multiple behaviors within the same test.

README.md
```diff
@@ -379,6 +379,7 @@ The default values for each rule are described below.
 | no_summary                   | warning |
 | no_array_responses           | error   |
 | parameter_order              | warning |
+| operation_id_naming_convention | warning |
```
dpopp07: Sorry I missed this before - this isn't the only place in the README that the rule should be documented. See the other rules for examples

Issue #110, opened 2019-10-21
[Feature Request] Docker Image to run openapi-validator


dkirrane:
Is there a Docker Image to run openapi-validator? (Similar to `openapitools/openapi-generator-cli`)

jamescooke:
Hi @dkirrane - I'm not officially connected to the project, but I've got a public docker container on gitlab here: https://gitlab.com/jamescooke/ibm-openapi-validator-docker

I was *just* looking at moving that container to docker hub, but my guess would be that it would be best if it was shipped there under the [IBM verified publisher account](https://hub.docker.com/publishers/ibmcorp).

Am happy to contribute / help out if any of the team here wishes. The `Dockerfile` I've written could be pinched and re-used if helpful: https://gitlab.com/jamescooke/ibm-openapi-validator-docker/blob/master/Dockerfile

dpopp07:
@jamescooke if you have a public container already, I would be more than happy for you to move it to Docker Hub and just have a community Docker image for the validator.

dpopp07:
@jamescooke Is the image in Docker Hub? If so, I'll close this issue

jamescooke:
@dpopp07 Not yet, sorry :disappointed: 

While writing this, I spotted that @dkirrane has created a public image here: https://hub.docker.com/r/dkirrane/ibm-openapi-validator - maybe that's a good enough reason to close this issue?

dpopp07:
Ah yes, that sounds good to me.

@dkirrane as the OP if you're after a different solution, feel free to re-open. I'm going to close for now

jamescooke:
An updated, semi-auto-building, slim container is now up on Docker hub: https://hub.docker.com/r/jamescooke/openapi-validator

I've only just spotted that it's possible to compile a binary with `npm run pkg` - so it might be possible to make future improvements to this container in terms of reduced size, speed.

PR #150, opened 2020-03-04
Add Docker image as installation option in README

jamescooke:
Related to #110 - adding info to the README on running the validator with Docker.


jamescooke:
> Sorry I hadn't gotten to it yet

No problem at all - just helpin'.

PR #154, opened 2020-03-10
feat: adds a default warning for 422 and 302 responses

barrett-schonefeld:
Reason:
- IBM Handbook says 400 response should be used instead of 422
- IBM Handbook prefers 303 and 307 responses over 302

Changes:
- added warning for 422 response codes
- added warning for 302 response codes
- added ibm_status_code_guidelines config option with default warning

Tests
- added a test to ensure warning issued for 422 response
- added a test to ensure warning issued for 302 response

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,7 @@ The supported rules are described below:
 | no_response_codes         | Flag any response object that has no valid response codes.   | oas3 |
 | no_success_response_codes | Flag any response object that has no success response codes. | oas3 |
 | no_response_body          | Flag any non-204 success responses without a response body.  | oas3 |
+| ibm_status_code_guidelines| Flag any violations of status code conventions per IBM API Handbook  | oas3 |
 
 ##### schemas
 | Rule                        | Description                                                                   | Spec     |
@@ -373,6 +374,7 @@ The default values for each rule are described below.
 | no_response_codes         | error   |
 | no_success_response_codes | warning |
 | no_response_body          | warning |
+| ibm_status_code_guidelines| warning |
 
 ##### schemas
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -252,6 +252,7 @@ The supported rules are described below:
 | no_response_codes         | Flag any response object that has no valid response codes.   | oas3 |
 | no_success_response_codes | Flag any response object that has no success response codes. | oas3 |
 | no_response_body          | Flag any non-204 success responses without a response body.  | oas3 |
+| ibm_status_code_guidelines| Flag any violations of status code conventions per IBM API Handbook  | oas3 |
```
dpopp07: I like the idea of having these rules wrapped under a general key to avoid too much verbosity between similar rules (especially for rules specific to IBM). That said, it's a slight deviation from previous patterns so I'd like to see what @mkistler thinks as well

barrett-schonefeld: I wish I could take credit, but I stole the idea from @mkistler!

barrett-schonefeld:
I changed the structure a bit. Since these 302 and 422 warnings are similar, I put the checks in one for-loop rather than in the top level if-else statement.

Now the module says, "If no status codes, error. Else, check for discouraged status codes and validate success codes."

src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -49,29 +49,47 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
+              config.ibm_status_code_guidelines
             );
-          } else if (successCode === '204' && obj[successCode].content) {
+          } else if (statusCode === '302') {
             messages.addMessage(
-              path.concat(['204', 'content']),
-              `A 204 response MUST NOT include a message-body.`,
-              'error'
+              path.concat(['302']),
+              'Should use status codes 303 and 307 instead of 302.',
```
dpopp07: Picky grammatical suggestion - not possible to define 303 _and_ 307 as status codes 🙂 

```suggestion
              'Should use status codes 303 or 307 instead of 302.',
```

PR #162, opened 2020-04-16
feat: Added check for case collsions of property names

rmkeezer:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1653

When there are two property names of different conventions but equivalent names, an error with be reported. (e.g. `propName` and `prop_name` exist)

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/schema-ibm.js
@@ -115,7 +115,7 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
 
       checkEnumValues(schema, path, config, messages);
     } else {
-      // optional support for property_case_convention and enum_case_convention
+      // optional support for property_case_convention, property_case_collision, and enum_case_convention
       // in config.  In the else block because support should be mutually exclusive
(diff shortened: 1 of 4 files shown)


src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -415,6 +426,57 @@ function checkPropNames(schema, contextPath, config, messages) {
...
+        if (isCaseCollision) {
+          messages.addMessage(
+            contextPath.concat(['properties', propName1]),
+            `Property names are identical except for the naming conventions: ${propName1} ~ ${propName2}`,
+            checkStatus
+          );
+        }
+      }
+    });
+  });
```
dpopp07, line 477: I think we can use a simpler algorithm here. This is a little tough to read. Here is some pseudo-code as a starting point:
```
create an array to hold used names
for name in properties:
  if property shouldn't be processed, skip (deprecated, etc.)
  make name caseless
  if array includes name, error
  else, add name to array
```

I don't think you need to worry about duplicate properties, those should be caught elsewhere. I also think it's okay to only list one of the property names in the error message.

dpopp07:
Looks good, I think you're good to merge 👍

PR #176, opened 2020-07-17
Allow additional verbs for PUT and DELETE

hudlow:


src/plugins/validation/2and3/semantic-validators/operation-ids.js
```diff
@@ -71,25 +76,42 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
+        }
       }
-      // operationId for DELETE should starts with "delete"
+      // operationId for DELETE should start with "delete", "remove", or "unset"
       else if (opKey === 'delete') {
         verbs.push('delete');
+        
+        // "remove" should complement "add"
+        if (
+          allPathOperations.includes('post') && allPathOperationIds.post.startsWith('add') ||
```
danhudlow, line 103: Oops, this line doesn't make sense — the `POST` would be on another path.

PR #186, opened 2020-08-10
feat: Validate missing response description

rmkeezer:
Throw error when a description is missing for any response schema since failing to do so will cause a compilier error.


src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -75,6 +75,13 @@ module.exports.validate = function({ resolvedSpec }, config) {
           );
         } else {
           for (const statusCode of successCodes) {
+            if (!obj[statusCode].description) {
+              messages.addMessage(
+                path.concat([statusCode]),
+                `A ${statusCode} response should include a description.`,
+                'error'
+              );
+            }
```
dpopp07, line 84: This looks good. I actually think we should move this check up to line 55 since this is required of all responses, not just success responses.

test/plugins/validation/oas3/responses.test.js
```diff
@@ -83,6 +83,35 @@ describe('validation plugin - semantic - responses - oas3', function() {
...
+              }
+            }
+          }
+        }
+      }
+    };
+
+    const res = validate({ resolvedSpec: spec }, config);
+    expect(res.warnings.length).toEqual(0);
+    expect(res.errors.length).toEqual(1);
```
dpopp07, line 112: Need to add an `expect` statement for the error message as well

src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -52,6 +52,13 @@ module.exports.validate = function({ resolvedSpec }, config) {
       } else {
         // default warnings for discouraged status code per IBM API Handbook
```
dpopp07, line 53: Okay I have to be a little obnoxious here but for the sake of clarity, this comment should be moved to be just before `if (statusCode === '422') {`. Those are the checks the comment describes, so it is a little confusing to leave here as if it applies to everything.

src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -52,6 +52,13 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
         for (const statusCode of statusCodes) {
+          if (!obj[statusCode].description) {
+            messages.addMessage(
+              path.concat([statusCode]),
+              `A ${statusCode} response should include a description.`,
```
dpopp07, line 58: I didn't look at this close enough last time - because this check is an unconditional requirement and applies to every response, this message is slightly misleading. I think it should be changed to

```suggestion
              'All responses must include a description.',
```

PR #187, opened 2020-08-14
fix: Required properties warning now checked in child refs

rmkeezer:
Required property definitions are now checked for in objects that have allOfs, anyOfs, and refs. Also if they are not defined it is only a warning since "required" properties are not required to be defined anywhere.

I added an extra util function for reducing an object (resolving it's refs) since I did not see an easy way of getting an object in the jsSpec using only the ref path. Let me know if there is an easier way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,7 @@ The supported rules are described below:
 | property_case_collision     | Flag any property with a `name` that is identical to another property's `name` except for the naming convention | shared |
 | enum_case_convention        | Flag any enum with a `value` that does not follow a given case convention. snake_case_only must be 'off' to use.    | shared |
 | json_or_param_binary_string | Flag parameters or application/json request/response bodies with schema type: string, format: binary. | oas3 |
+| undefined_required_properties| Flag any schema with undefined required properties                       | shared   |
 
 ##### security_definitions
 | Rule                        | Description                                                                           | Spec   |
@@ -383,6 +384,7 @@ The default values for each rule are described below.
 | Rule                        | Default |
 | --------------------------- | ------- |
 | json_or_param_binary_string | warning |
+| undefined_required_properties    | warning |
 
 ##### shared
 
--- a/src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
+++ b/src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
@@ -10,8 +10,43 @@
 
 const { walk } = require('../../../utils');
 const MessageCarrier = require('../../../utils/messageCarrier');
+const at = require('lodash/at');
 
-module.exports.validate = function({ jsSpec }) {
+const reduceObj = function(jsSpec, obj) {
+  if (obj['$ref']) {
+    const objPath = obj['$ref'].split('/');
+    objPath.shift();
+    return reduceObj(jsSpec, at(jsSpec, [objPath])[0]);
+  }
+  return obj;
+};
+
+const checkReqProp = function(jsSpec, obj, requiredProp) {
+  obj = reduceObj(jsSpec, obj);
+  if (obj.properties && obj.properties[requiredProp]) {
+    return true;
+  } else if (Array.isArray(obj.anyOf) || Array.isArray(obj.oneOf)) {
+    const childList = obj.anyOf || obj.oneOf;
+    let reqPropDefined = true;
+    childList.forEach((childObj) => {
+      if (!checkReqProp(jsSpec, childObj, requiredProp)) {
+        reqPropDefined = false;
+      }
+    });
+    return reqPropDefined;
+  } else if (Array.isArray(obj.allOf)) {
+    let reqPropDefined = false;
+    obj.allOf.forEach((childObj) => {
+      if (checkReqProp(jsSpec, childObj, requiredProp)) {
+        reqPropDefined = true;
+      }
+    });
+    return reqPropDefined;
+  }
+  return false;
+}
+
+module.exports.validate = function({ jsSpec }, config) {
   const messages = new MessageCarrier();
 
   walk(jsSpec, [], function(obj, path) {
@@ -36,13 +71,14 @@ module.exports.validate = function({ jsSpec }) {
       }
 
       // Assertation 2
+      const undefinedRequiredProperties = config.schemas.undefined_required_properties;
       if (Array.isArray(obj.required)) {
         obj.required.forEach((requiredProp, i) => {
-          if (!obj.properties || !obj.properties[requiredProp]) {
+          if (!checkReqProp(jsSpec, obj, requiredProp)) {
             messages.addMessage(
               path.concat([`required[${i}]`]).join('.'),
-              "Schema properties specified as 'required' must be defined",
-              'error'
+              "Schema properties specified as 'required' should be defined",
+              undefinedRequiredProperties
             );
           }
         });
--- a/test/plugins/validation/2and3/items-required-for-array-objects.test.js
+++ b/test/plugins/validation/2and3/items-required-for-array-objects.test.js
@@ -2,6 +2,7 @@ const expect = require('expect');
 const {
   validate
 } = require('../../../../src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects');
+const config = require('../../../../src/.defaultsForValidator').defaults.shared;
 
 describe('validation plugin - semantic - items required for array objects - Swagger 2', () => {
   it('should return an error when an array header object omits an `items` property', () => {
@@ -34,7 +35,7 @@ describe('validation plugin - semantic - items required for array objects - Swag
       }
     };
 
-    const res = validate({ jsSpec: spec });
+    const res = validate({ jsSpec: spec }, config);
     expect(res.errors.length).toEqual(1);
     expect(res.errors[0].path).toEqual([
       'paths',
@@ -84,7 +85,7 @@ describe('validation plugin - semantic - items required for array objects - Swag
       }
     };
 
-    const res = validate({ jsSpec: spec });
+    const res = validate({ jsSpec: spec }, config);
     expect(res.errors.length).toEqual(0);
     expect(res.warnings.length).toEqual(0);
   });
@@ -98,7 +99,7 @@ describe('validation plugin - semantic - items required for array objects - Swag
       }
     };
 
-    const res = validate({ jsSpec: spec });
+    const res = validate({ jsSpec: spec }, config);
     expect(res.errors.length).toEqual(1);
     expect(res.errors[0].path).toEqual('definitions.TestObject');
     expect(res.errors[0].message).toEqual(
@@ -121,7 +122,7 @@ describe('validation plugin - semantic - items required for array objects - Swag
       }
     };
 
-    const res = validate({ jsSpec: spec });
+    const res = validate({ jsSpec: spec }, config);
     expect(res.errors.length).toEqual(1);
     expect(res.errors[0].path).toEqual(
       'definitions.TestObject.properties.arrayProperty'
@@ -162,7 +163,7 @@ describe('validation plugin - semantic - items required for array objects - Open
       }
     };
 
-    const res = validate({ jsSpec: spec });
+    const res = validate({ jsSpec: spec }, config);
     expect(res.errors.length).toEqual(1);
     expect(res.errors[0].path).toEqual(
       'paths./pets.get.responses.200.headers.X-MyHeader.schema'
@@ -173,7 +174,7 @@ describe('validation plugin - semantic - items required for array objects - Open
     expect(res.warnings.length).toEqual(0);
   });
 
-  it('should return an error when a model does not define a required property', () => {
+  it('should return a warning when a model does not define a required property', () => {
     const spec = {
       components: {
         schemas: {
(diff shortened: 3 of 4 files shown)


src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -39,11 +40,40 @@ module.exports.validate = function({ jsSpec }) {
       if (Array.isArray(obj.required)) {
         obj.required.forEach((requiredProp, i) => {
           if (!obj.properties || !obj.properties[requiredProp]) {
-            messages.addMessage(
-              path.concat([`required[${i}]`]).join('.'),
-              "Schema properties specified as 'required' must be defined",
-              'error'
-            );
+            // For allOf, ensure atleast one of the child objects has the required property
+            if (obj.allOf) {
```
dpopp07, line 44: Might be a good idea to verify that this is an array. I know it is supposed to be according to the spec, but it would add some type safety since the code will crash on `forEach` otherwise.

test/plugins/validation/2and3/items-required-for-array-objects.test.js
```diff
@@ -173,32 +173,46 @@ describe('validation plugin - semantic - items required for array objects - Open
...
+    expect(res.warnings.length).toEqual(1);
+    expect(res.warnings[0].path).toEqual(
       'components.schemas.TestObject.required[0]'
     );
-    expect(res.errors[0].message).toEqual(
-      "Schema properties specified as 'required' must be defined"
+    expect(res.warnings[0].message).toEqual(
+      "Schema properties specified as 'required' should be defined"
     );
-    expect(res.warnings.length).toEqual(0);
```
dpopp07, line 201: We need to have tests for all four scenarios: `allOf`, `oneOf`, `anyOf`, or none of the above.

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -39,11 +40,40 @@ module.exports.validate = function({ jsSpec }) {
...
+              var props = {};
+              obj.allOf.forEach((childObj) => {
+                childObj = reduceObj(jsSpec, childObj);
+                props = Object.assign(props, childObj.properties);
+              });
+              if (!props[requiredProp]) {
+                messages.addMessage(
+                  path.concat([`required[${i}]`]).join('.'),
+                  "Schema properties specified as 'required' should be defined",
+                  'warning'
```
dpopp07, line 54: I'm thinking we should make the a configurable check, instead of hard coding a "warning", and maybe setting the default value to error. What do you think?

src/plugins/utils/reduceObj.js
```diff
@@ -0,0 +1,17 @@
...
+    obj['allOf'].forEach((childObj) => {
+      newObj = Object.assign(newObj, reduceObj(jsSpec, childObj));
+    });
+    return newObj;
+  }
+  return obj;
+};
+
+module.exports = reduceObj;
+
```
dpopp07, line 17: I believe there is an easier way. In other validators, we use `lodash/at` to resolve references. See [this line](https://github.com/IBM/openapi-validator/blob/d1ed5c0eef42f95aaac6befc80ae02a9fdb5bb04/src/plugins/validation/2and3/semantic-validators/walker-ibm.js#L36) for an example. You should be able to do this with the `jsSpec`. You may need some additional logic but you can just add a simple helper function in `items-required-for-array-objects.js` rather than creating a new utility file.

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
rmkeezer: It's a good idea to make it configurable, but i think setting the default value to "warning" would be better since the original problem statement, issue #1920, wanted to reduce importance of this issue from error to warning. @mkistler what do you think?

src/plugins/utils/reduceObj.js
rmkeezer: That's what I was looking for, great! Although it didn't actually reduce my logic much, I still have to convert the path from `"#/path/to/ref"` to `['path', 'to', 'ref']` and then reduce any $refs or allOfs

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
dpopp07: Oh good call - I missed that issue. Warning sounds good to me 👍

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -36,14 +52,44 @@ module.exports.validate = function({ jsSpec }) {
...
+              }
+            // For anyOf/oneOf, ensure all of the child objects have the required property
+            } else if (Array.isArray(obj.anyOf) || Array.isArray(obj.oneOf)) {
+              var childList = obj.anyOf || obj.oneOf;
+              childList.forEach((childObj) => {
+                childObj = reduceObj(jsSpec, childObj);
+                if (childObj.properties && !childObj.properties[requiredProp]) {
+                  messages.addMessage(
+                    path.concat([`required[${i}]`]).join('.'),
+                    "Schema properties specified as 'required' should be defined",
```
dpopp07, line 81: My concern here is that, if an API definition defines a `oneOf` with three schemas, and two of them are missing the definition, they will get two identical messages without distinguishing which ones are problematic.

Perhaps we should use a message like "Schema should define required property <property name>" and point the path at the schema itself

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -10,8 +10,24 @@
 
 const { walk } = require('../../../utils');
 const MessageCarrier = require('../../../utils/messageCarrier');
+const at = require('lodash/at');
 
-module.exports.validate = function({ jsSpec }) {
+const reduceObj = function(jsSpec, obj) {
+  if (obj['$ref']) {
+    var objPath = obj['$ref'].split('/');
```
dpopp07, line 17: I meant to comment on this earlier, but we shouldn't use `var` anywhere. These should all be `const`.

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -10,8 +10,24 @@
...
+    objPath.shift();
+    return reduceObj(jsSpec, at(jsSpec, [objPath])[0]);
+  } else if (obj['allOf']) {
```
dpopp07, line 20: What if the object is a `oneOf` or `anyOf`? I'm not sure what you would want to do here but I'm not sure it would work as intended if that happened.

test/plugins/validation/2and3/items-required-for-array-objects.test.js
```diff
@@ -190,15 +191,141 @@ describe('validation plugin - semantic - items required for array objects - Open
...
+    expect(res.warnings.length).toEqual(1);
+    expect(res.warnings[0].path).toEqual(
+      'components.schemas.TestObject.required[0]'
+    );
+    expect(res.warnings[0].message).toEqual(
+      "Schema properties specified as 'required' should be defined"
+    );
+  });
+
+  it('should return an warning when a model with an anyOf does not define a required property in all of its children', () => {
```
dpopp07, line 247:
```suggestion
  it('should return a warning when a model with an anyOf does not define a required property in all of its children', () => {
```

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -36,14 +52,44 @@ module.exports.validate = function({ jsSpec }) {
...
+      const checkRequiredProps = config.operations.check_required_properties;
       if (Array.isArray(obj.required)) {
         obj.required.forEach((requiredProp, i) => {
           if (!obj.properties || !obj.properties[requiredProp]) {
-            messages.addMessage(
-              path.concat([`required[${i}]`]).join('.'),
-              "Schema properties specified as 'required' must be defined",
-              'error'
-            );
+            // For allOf, ensure atleast one of the child objects has the required property
```
dpopp07, line 59:
```suggestion
            // For allOf, ensure at least one of the child objects has the required property
```

README.md
```diff
@@ -397,6 +397,7 @@ The default values for each rule are described below.
 | no_array_responses           | error   |
 | parameter_order              | warning |
 | operation_id_naming_convention | warning |
+| check_required_properties    | warning |
```
dpopp07, line 400: The naming convention for rules is usually to describe what the issue is. For example, `no_array_responses` leads to `error`.

I think we should name this something like `undefined_required_properties`

dpopp07: Also there's one more place in the README that needs to be filled in - the description for each rule

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
dpopp07: I think it might also be good to have a test for the scenario of an `anyOf` here to make sure its covered.

Thinking about it now, I wonder if using the `resolvedSpec` would make all of this go away haha

rmkeezer: I took a look at using resolvedSpec, but it seems to resolve just $refs and not allOf/oneOf/anyOf unless i'm using it wrong. In this case I dont think oneOfs or anyOfs can be feasibly "reduced" here since for allOf i can just combine all the properties of the children but for oneOfs and anyOfs, each child must be checked that it defines the required property.

dpopp07: That's correct, it would resolve all $refs, including the refs inside of allOf/oneOf/anyOf, but no it wouldn't reconcile all of the sub-schemas. It still might be helpful.

I get that's the trouble with oneOf/anyOf, but I'm not sure the solution should be to ignore them. Maybe this function should be modified to return an array of schemas and in the case of single schemas or allOf, you just return one element. Then `forEach` can be called on the result. Just an idea. I guess this would only happen if sub-schemas to an allOf/oneOf/anyOf were _themselves_ a allOf/oneOf/anyOf, at which point the schema is just obnoxiously complicated... but I can't help but feel we should be complete here, even for the extreme scenario

test/plugins/validation/2and3/items-required-for-array-objects.test.js
```diff
@@ -190,17 +191,293 @@ describe('validation plugin - semantic - items required for array objects - Open
...
+        }
+      }
+    };
+
+    const res = validate({ jsSpec: spec }, config);
+    expect(res.errors.length).toEqual(0);
+    expect(res.warnings.length).toEqual(0);
+  });
+
+  it('should not return a warning when a anyOf model with a nested allOf defines a required property in atleast one of its children', () => {
```
dpopp07, line 339:
```suggestion
  it('should not return a warning when a anyOf model with a nested allOf defines a required property in at least one of its children', () => {
```

test/plugins/validation/2and3/items-required-for-array-objects.test.js
```diff
@@ -190,17 +191,293 @@ describe('validation plugin - semantic - items required for array objects - Open
...
+        }
+      }
+    };
+
+    const res = validate({ jsSpec: spec }, config);
+    expect(res.errors.length).toEqual(0);
+    expect(res.warnings.length).toEqual(0);
+  });
+
+  it('should not return a warning when a model with a nested allOf defines a required property in atleast one of its children', () => {
```
dpopp07, line 389:
```suggestion
  it('should not return a warning when a model with a nested allOf defines a required property in at least one of its children', () => {
```

test/plugins/validation/2and3/items-required-for-array-objects.test.js
```diff
@@ -190,17 +191,293 @@ describe('validation plugin - semantic - items required for array objects - Open
...
+        }
+      }
+    };
+
+    const res = validate({ jsSpec: spec }, config);
+    expect(res.errors.length).toEqual(0);
     expect(res.warnings.length).toEqual(0);
   });
 
+  it('should return a warning when a model with an allOf does not define a required property in atleast one of its children', () => {
```
dpopp07, line 439:
```suggestion
  it('should return a warning when a model with an allOf does not define a required property in at least one of its children', () => {
```

README.md
```diff
@@ -213,6 +213,7 @@ The supported rules are described below:
 | no_array_responses           | Flag any operations with a top-level array response.                                | shared   |
 | parameter_order              | Flag any operations with optional parameters before a required param.               | shared   |
 | operation_id_naming_convention | Flag any `operationId` that does not follow naming convention.                    | shared   |
+| undefined_required_properties| Flag any operations with undefined required properties                              | shared   |
```
dpopp07, line 216: We aren't quite flagging "operations" so I think this description could be more accurate - we're flagging properties of a schema.

This comment is making me realize this configuration option belongs under the "schemas" category rather than the "operations" category - my apologies for not catching this earlier

src/plugins/validation/2and3/semantic-validators/items-required-for-array-objects.js
```diff
@@ -10,8 +10,47 @@
...
+    return reqPropDefined;
+  } else if (Array.isArray(obj.allOf)) {
+    let reqPropDefined = false;
+    obj.allOf.forEach((childObj) => {
+      if (childObj.properties && childObj.properties[requiredProp]) {
+        reqPropDefined = true;
+      } else if (checkReqProp(jsSpec, childObj, requiredProp)) {
+        reqPropDefined = true;
+      }
+    });
```
dpopp07, line 47: Just like above, I think this can be simplified:

```suggestion
    obj.allOf.forEach((childObj) => {
      if (checkReqProp(jsSpec, childObj, requiredProp)) {
        reqPropDefined = true;
      }
    });
```

test/plugins/validation/2and3/items-required-for-array-objects.test.js
rmkeezer: Good catch, I fixed this and must have accidentally reverted it haha

PR #191, opened 2020-08-28
Added support to validate if examples are set in the schemas as well as support for custom validation 

ChrisPhillips-cminion:
I have added two key features in this PR

1. Added support to validate that schemas include an example
2. Added support to allow users to specify custom validation pattern matching,

All tests pass

Read Me Updated

Tests have not been added for


ChrisPhillips-cminion:
Closed as ive done some clean up

PR #194, opened 2020-08-31
fix: dont validate schema properties called 'responses' as responses

### LINKED ISSUE #188: Property name in model schema is triggering the response validation
We have a schema that contains `responses` property e.g.
```
    "NodeStatsInformation":
      "description": "Example NodeStatsInformation response."
      "value":
...
        "couch_replicator":
...
          "responses":
            "failure":
              ...
```
I believe the `responses` property above is triggering the tool to run `responses` validation.  We get this error:
```
  Message :   Each `responses` object MUST have at least one response code.
  Path    :   paths./_node/{node_name}/_stats.get.responses.200.content.application/json.schema.properties.couch_replicator.properties.responses
  Line    :   9779
```

dpopp07:
Previously, the only way we checked for a response object was by checking for the string "responses" in the path at a certain location. This would cause schema properties named `responses` to be validated as if it were a response. This would cause confusing warnings/errors.

This PR introduces robust logic to prove that the object in question is in fact, a response.

Additionally, I fixed some linting issues and added a linting step to the Travis build so that these things won't go uncaught in the future.

Resolves #188

### DIFF
--- a/src/plugins/utils/is-response.js
+++ b/src/plugins/utils/is-response.js
@@ -1,0 +1,35 @@
+module.exports = (path, isOAS3 = true) => {
+  const operations = [
+    'get',
+    'put',
+    'post',
+    'delete',
+    'options',
+    'head',
+    'patch',
+    'trace'
+  ];
+
+  const pathLength = path.length;
+
+  // a necessary but not sufficient check for a response object
+  // without these, a schema property named "responses"
+  // may be validated as a response
+  const isResponsesProperty = path[pathLength - 1] === 'responses';
+
+  // three scenarios:
+  // 1) inside of an operation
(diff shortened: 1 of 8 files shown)


src/plugins/utils/is-response.js
```diff
@@ -0,0 +1,35 @@
...
+    'patch',
+    'trace'
+  ];
+
+  const pathLength = path.length;
+
+  // a necessary but not sufficient check for a response object
+  // without these, a schema property named "responses"
+  // may be validated as a response
+  const inResponsesSection = path[pathLength - 1] === 'responses';
```
mkistler, line 18: I think this name is misleading.  It would be better as `isResponsesProperty` or `isNamedResponses`.

PR #195, opened 2020-09-09
feat: initial Spectral integration

jorge-ibm:
Initial integration of Spectral into the IBM Validator.
Changes:
- Spectral is now ran as part of the validator execution. Any results found through spectral are then merged with the results found through the IBM Validator.
- A default Spectral ruleset ``.defaultsForSpectral.json`` was created and includes all the [Spectral rules](https://meta.stoplight.io/docs/spectral/docs/reference/openapi-rules.md) for Swagger/OAS3 that are recommended by default, and don't overlap with any of the Validator's rules. Any future rules can be easily added from this file.
- Added logic for Spectral to use a ``.spectral.json`` file if it's found in the cwd to give users the ability to specify their own custom Spectral ruleset.
- Added several unit tests to ensure none of the enabled Spectral rules overlap with our validator's rules.
- Fixed minor typo in an existing rule: ``securitySchemas`` -> ``securitySchemes``

Given that the results from Spectral are just "fed in" to the validator, and the current validator logic handles the logic from there, I did not find any issues running any of the validator's cmd line options with this new integration.


src/spectral/rulesets/.defaultsForSpectral.json
```diff
@@ -0,0 +1,27 @@
+{
```
jorge-ibm, line 1: If we want to start with a smaller ruleset, or no rules at all, we can also do that to minimize the blast radius of this change. Either option would still give users the ability to write their own rules.

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,44 @@
...
+
+  if (results) {
+    for (const validationResult of results) {
+      if (validationResult) {
+        const code = validationResult['code'];
+        const severity = validationResult['severity'];
+        const message = validationResult['message'];
+        const path = validationResult['path'];
+
+        if (code && severity != null && message && path) {
```
jorge-ibm, line 14: @dpopp07 - my Node is rusty, but I hope simply ANDing each variable is the equivalent of ``!= null && != undefined``

src/plugins/validation/2and3/semantic-validators/parameters-ibm.js
```diff
@@ -92,7 +92,7 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
         let messageAuth =
           'Parameters must not explicitly define `Authorization`.';
         messageAuth = isOAS3
-          ? `${messageAuth} Rely on the \`securitySchemas\` and \`security\` fields to specify authorization methods.`
+          ? `${messageAuth} Rely on the \`securitySchemes\` and \`security\` fields to specify authorization methods.`
```
jorge-ibm, line 95: Minor typo Phil pointed out. This should be [securitySchemes](https://swagger.io/docs/specification/authentication/) in OAS3

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,44 @@
...
+          if (code == 'parser') {
```
dpopp07, line 15: Typically, you always want to use strict operators. This applies to the following checks as well

```suggestion
          if (code === 'parser') {
```

src/spectral/utils/spectral-validator.js
dpopp07: That is correct, although I would probably want to check more than just `null` for `severity`. Maybe something like `typeof severity === 'number'`

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,44 @@
...
+            // Spectral doesn't allow disabling parser rules, so don't include them
+            // in the output for now
+            continue;
+          }
+          // Our validator only supports warning/error level, so only include
+          // those validation results (for now)
+          if (severity == '1') {
```
dpopp07, line 22: I thought you said `severity` would be an integer? If it's a string, you can treat it the same in the if statement above

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,44 @@
...
+            //warning
+            messages.addMessage(path, message, 'warning');
+          } else if (severity == '0') {
+            //error
+            messages.addMessage(path, message, 'error');
+          }
+        } else {
+          console.error(
+            'There was an error while parsing the spectral results: ',
+            JSON.stringify(validationResult)
```
dpopp07, line 32: This isn't necessarily an error, right? It just means the object didn't look like we expected it to. Maybe we should wrap this in a check for "debug"

test/spectral/tests/disabled-rules.test.js
```diff
@@ -0,0 +1,659 @@
...
+    // set up mock user input
+    const program = {};
+    program.args = ['./test/spectral/mockFiles/swagger/disabled-rules.yml'];
+    program.default_mode = true;
+
+    const exitCode = await commandLineValidator(program);
+    const capturedText = getCapturedText(consoleSpy.mock.calls);
+    const allOutput = capturedText.join('');
+
+    expect(exitCode).toEqual(1);
```
dpopp07, line 25: If no rules are triggered, the exit code should be 0

test/spectral/tests/enabled-rules.test.js
```diff
@@ -0,0 +1,522 @@
...
+    program.print_validator_modules = true;
+
+    const exitCode = await commandLineValidator(program);
+    const capturedText = getCapturedText(consoleSpy.mock.calls);
+    const allOutput = capturedText.join('');
+    // Spectral should be the only validator module in the output
+    const count = (allOutput.match(/Validator:/g) || []).length;
+
+    expect(exitCode).toEqual(1);
+    expect(count).toEqual(2);
```
dpopp07, line 29: I think I just don't understand this check but I expected this "count" to be 1

dpopp07 (changes requested):
Leaving an initial review of some things to clean up in the code while others are looking at the PR. It looks great so far but I'm going to keep looking at it

src/spectral/rulesets/.defaultsForSpectral.json
```diff
@@ -0,0 +1,27 @@
...
+	"extends": [["spectral:oas", "off"]],
+	"formats": ["oas2", "oas3"],
+	"functionsDir": "../functions",
+	"rules": {
+		"no-eval-in-markdown": true,
+		"no-script-tags-in-markdown": true,
+		"openapi-tags": true,
+		"operation-description": true,
```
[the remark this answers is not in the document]
jorge-ibm: They _shouldn't_.  I took care to try to test each rule I was enabling and making sure there wasn't duplicate output. Though if there's a rule we think is a duplicate, then it's possible I might've overlooked it.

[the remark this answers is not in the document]
dpopp07: Cool, that sounds good. I just wanted to double check

test/spectral/tests/disabled-rules.test.js
jorge-ibm: Well this whole test suite tests that Spectral rules are disabled, so regular validator rules can still be triggered. In this test case, some non-spectral rules triggered errors in validation. I thought of creating a separate api file for each spectral rule I was testing, to make sure other validations weren't being triggered but I figured that would be overkill.

test/spectral/tests/enabled-rules.test.js
jorge-ibm: Well I'm hoping to just test that Spectral is the only validator module that is included in the output (to test for possible duplicate rules). I guess the issue is, we print each validator module in errors/warnings:
```
errors

Validator: spectral
  Message :   `number_of_coins` property type should be integer
  Path    :   definitions.ModelWithBadExample.example.number_of_coins
  Line    :   177

warnings

Validator: spectral
```
So maybe I can think of a better way to test this..

test/spectral/tests/disabled-rules.test.js
dpopp07: That's reasonable - will you just add a comment explaining that? It was confusing to see a "positive test" with an exit code of 0

src/spectral/utils/spectral-validator.js
jorge-ibm: It is, but since I'm using loose equality here (``==``) node matched the type. I'll change these conditions to strict equality.

src/cli-validator/runValidator.js
```diff
@@ -15,6 +15,11 @@ const print = require('./utils/printResults');
 const printJson = require('./utils/printJsonResults');
 const printError = require('./utils/printError');
 const preprocessFile = require('./utils/preprocessFile');
+const { Spectral, isOpenApiv2, isOpenApiv3 } = require('@stoplight/spectral');
+
+// default spectral ruleset file
+const defaultSpectralRuleset =
+  process.cwd() + '/src/spectral/rulesets/.defaultsForSpectral.json';
```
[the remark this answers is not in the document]
jorge-ibm: Interesting.. thanks for the tip! I'll make that change

jorge-ibm:
I realized that I missed updating the code to also include Spectral in api invocations of the validator. I'll submit some changes soon to make sure that's accounted for.

src/cli-validator/runValidator.js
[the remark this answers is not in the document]
dpopp07: Good catch, I didn't think that suggestion though all the way

jorge-ibm:
@mkistler @dpopp07 this is ready for re-review

src/spectral/rulesets/.defaultsForSpectral.json
mkistler, line 1: Was there some reason to avoid using yaml?

src/cli-validator/utils/processConfiguration.js
```diff
@@ -367,9 +367,31 @@ const validateConfigOption = function(userOption, defaultOption) {
   return result;
 };
 
+const getSpectralRuleset = async function(defaultRuleset) {
+  const ruleSetFileName = '.spectral.json';
```
mkistler, line 371: I think we'd improve adoption if we supported yaml in addition or even instead of json.  Spectral strongly favors yaml for its configuration file -- all it's examples are shown in yaml.

mkistler (approved):
This looks good but I recommend we allow users to provide their Spectral config in yaml format.

src/spectral/rulesets/.defaultsForSpectral.json
jorge-ibm: None at all, just my personal preference. I always found it easier to configure a json file vs a yaml. I just took this phrase from their documentation literally:
_A ruleset is a JSON or YAML file (often the file will be called .spectral.yaml), and there are two main parts_
I am completely not against changing our default ruleset to yaml! I'll go ahead and make that change

src/cli-validator/utils/processConfiguration.js
jorge-ibm: I'll add support for both a custom yaml and json ruleset file.

jorge-ibm:
@mkistler I just pushed a commit to change the default ruleset to yml, and allow for custom ruleset files to be either yml or json.

src/cli-validator/utils/processConfiguration.js
```diff
@@ -367,9 +367,39 @@ const validateConfigOption = function(userOption, defaultOption) {
...
+    '.spectral.json'
+  ];
+  let ruleSetFile;
+
+  // search up the file system for the first ruleset file found
+  try {
+    for (const file of ruleSetFilesToFind) {
+      if (!ruleSetFile) {
+        ruleSetFile = await findUp(file);
+      }
```
dpopp07, line 384: It looks like the files will be searched in prioritized order. I think this is a good approach, but maybe it should be documented

src/cli-validator/utils/processConfiguration.js
```diff
@@ -367,9 +367,39 @@ const validateConfigOption = function(userOption, defaultOption) {
...
+    }
+  } catch (err) {
```
dpopp07, line 386: This is an extremely minor comment but since you aren't using the error at all, I think it might be better style to just use `catch ()`

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,75 @@
...
+            messages.addMessage(path, message, 'error');
+          }
+        } else {
+          if (debug) {
+            console.log(
+              'There was an error while parsing the spectral results: ',
+              JSON.stringify(validationResult)
+            );
+          }
+          continue;
```
dpopp07, line 41: Correct me if I'm wrong but this `continue` statement looks superfluous - there's no more code after it

test/spectral/tests/spectral-validator.test.js
```diff
@@ -0,0 +1,102 @@
...
+    const debug = false;
+    const parsedSpectralResults = spectralValidator.parseResults(
+      mockResult,
+      debug
+    );
+    const capturedText = getCapturedText(consoleSpy.mock.calls);
+    const allOutput = capturedText.join('');
+
+    expect(parsedSpectralResults.errors.length).toEqual(1);
+    expect(parsedSpectralResults.warnings.length).toEqual(0);
```
dpopp07, line 36: I don't think you need this for every test but for at least one test, it would be good to make expectations about the message and path to verify they are parsed correctly

dpopp07 (commented):
Looking good but I have a few more comments to be addressed before merging

src/cli-validator/utils/processConfiguration.js
dpopp07: Come to think of it, it doesn't look like there are any README changes. Is there anything the user would need to know to take advantage of this new feature? We should probably add a section about it - even if it just tells the user what's available and links the appropriate Spectral documentation

src/lib/index.js
```diff
@@ -3,18 +3,41 @@ const config = require('../cli-validator/utils/processConfiguration');
...
   const swagger = await buildSwaggerObject(input);
-  const results = validator(swagger, configObject);
+
+  try {
+    if (typeof input !== 'object') {
+      // If input is not an object, assume it's a file
+      const readFile = util.promisify(fs.readFile);
+      input = await readFile(input, 'utf8');
+      input = preprocessFile(input);
+    }
```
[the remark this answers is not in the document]
jorge-ibm: I guess I'm confused about the documentation for the api [here](https://github.com/ibm/openapi-validator#node-module), is ``openApiDoc`` just an object in that example and not a file ?

src/cli-validator/utils/processConfiguration.js
jorge-ibm: I'm fine with adding some documentation on at least how to specify your own custom ruleset file. Though I had the idea that this initial implementation would be "hidden" from the user.

dpopp07: I guess this was intended to be a "Beta" feature for the time being, so that makes sense to me. Disregard this comment

src/lib/index.js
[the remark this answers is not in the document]
dpopp07: That's interesting that it worked fine - I think it's not "technically" supported though. So I don't think we need explicit code here for it

src/cli-validator/utils/processConfiguration.js
jorge-ibm: I had to change this to ``catch { }``

src/spectral/utils/spectral-validator.js
jorge-ibm: Yeah I don't think that's doing anything. I'll remove it.

src/spectral/utils/spectral-validator.js
```diff
@@ -0,0 +1,74 @@
...
+            // in the output (for now)
+            continue;
+          }
+          // Our validator only supports warning/error level, so only include
+          // those validation results (for now)
+          if (severity === 1) {
+            //warning
+            messages.addMessage(path, message, 'warning');
+          } else if (severity === 0) {
+            //error
```
dpopp07, line 31: Very picky style comment, comments should have a space after the "//" here and on line 28. It doesn't really matter but I'm surprised eslint didn't complain

dpopp07 (approved):
Looks good! 👍

PR #198, opened 2020-09-22
fix(validator): Spurious errors when using `application/merge-patch+json` media type

### LINKED ISSUE #197: Spurious errors when using `application/merge-patch+json` media type
Our `PATCH` methods use JSON merge patch semantics -- i.e., `application/merge-patch+json` rather than `application/json`.  However, attempting to use that type results in the following error from the validator:

```
  Message :   Operations with non-form request bodies should set a name with the x-codegen-request-body-name annotation.
  Path    :   paths./dedicated_host/groups/{id}.patch
  Line    :   385
```

This appears to be due to `src/plugins/validation/oas3/semantic-validators/operations.js` having too narrow an understanding of JSON format.  Per https://tools.ietf.org/html/rfc6838#section-4.2.8 any media type ending in `+json` needs to be regarded as being in JSON format.

meem:
Fixes #197.

### DIFF
--- a/src/plugins/validation/oas3/semantic-validators/operations.js
+++ b/src/plugins/validation/oas3/semantic-validators/operations.js
@@ -47,7 +47,8 @@ module.exports.validate = function({ resolvedSpec, jsSpec }, config) {
           // request body has content
           const firstMimeType = requestBodyMimeTypes[0]; // code generation uses the first mime type
           const oneContentType = requestBodyMimeTypes.length === 1;
-          const isJson = firstMimeType === 'application/json' ||
+          const isJson =
+	    firstMimeType === 'application/json' ||
             firstMimeType.endsWith('+json');
 
           const hasArraySchema =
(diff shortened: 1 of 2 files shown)


dpopp07 (changes requested):
The code looks good to me 👍 Do you mind adding a test to capture this behavior?

I also would like to hear what @mkistler thinks about the change

meem:
@dpopp07 Thanks for the review, and good call on adding a test case.  I added a specific one for this scenario, though I think we may need a broader rework of `operations.test.js` to account for both `application/json` and `+json` throughout.

mkistler (approved):
Looks good! 👍 

Thanks for the contribution @meem !

dpopp07:
@meem thanks for adding the test - I think this looks good for now. If you would like to add further tests, feel free to open a subsequent PR. Thanks!

dpopp07 (approved):
Looks good 👍  Thanks for the PR!

PR #201, opened 2020-10-01
fix: spectral file path resolving

### LINKED ISSUE #199: Relative path is not resolved correctly
With version 0.30.0, compared to 0.29.4, openapi-validator regressed in way how relative paths are resolved.

I have following folder structure:
```sh
src/
    api1.yml
    api2.yml
    shared/
        common.yml
```
`api1.yml` and `api2.yml` files reference some definitions from `shared/common.yml`, e.g. responses `shared/common.yml#/components/responses/NoContent`.

If I run validator from other folder than `src/` then e.g. `lint-openapi src/api1.yml`, then error is reported.

```sh
errors
  Message :   ENOENT: no such file or directory, open 'shared/common.yml'
  Path    :   paths./mymethod.post.responses.201.$ref
  Line    :   42
```
It seems as if relative path is now resolved in respect to current working directory, instead of relatively to specification file being processed.

Earlier versions work fine.

jorge-ibm:
Fixes an issue where Spectral wasn't correctly resolving file references when the file(s) to validate were not in the current working directory the validator was invoked. The fix was to temporarily switch to the directory to the location of the file we're validating, following what our "default" validator does.

Unit tests were also added to test file/remote url resolving

Resolves #199

### DIFF
--- a/src/cli-validator/runValidator.js
+++ b/src/cli-validator/runValidator.js
@@ -242,11 +242,27 @@ const processInput = async function(program) {
       process.chdir(originalWorkingDirectory);
     }
 
-    // run validator & spectral, print the results, and determine if validator passed
+    // run spectral and save the results
+    let spectralResults;
+    try {
+      process.chdir(path.dirname(validFile));
+      // let spectral handle the parsing of the original swagger/oa3 document
+      spectralResults = await spectral.run(originalFile);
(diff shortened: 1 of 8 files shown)


dpopp07 (approved):
Looks great! 👍 Thanks for adding that test

PR #202, opened 2020-10-02
feat: allow spectral rules to be defined in the validaterc

rmkeezer:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2170
followup of: https://github.com/IBM/openapi-validator/pull/195

These changes allow spectral rules to have their severity level changed in the `.validatrc` and allow custom rules defined in a `.spectral.yaml` file to not be overridden by the `.defaultsForSpectral.yaml`. Although the `.defaultsForSpectral.yaml` will still override the `.spectral.yaml` severity level of any existing rule `.defaultsForSpectral.yaml` defines a severity for, including the extended rules.

### DIFF
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -106,6 +106,30 @@ const defaults = {
     'schemas': {
       'json_or_param_binary_string': 'warning'
     }
+  },
+  'spectral': {
+    'rules': {
+      'no-eval-in-markdown': "warning",
+      'no-script-tags-in-markdown': "warning",
+      'openapi-tags': "warning",
+      'operation-description': "warning",
+      'operation-tags': "warning",
+      'operation-tag-defined': "warning",
+      'path-keys-no-trailing-slash': "warning",
+      'typed-enum': "warning",
+      'oas2-api-host': "warning",
+      'oas2-api-schemes': "warning",
+      'oas2-host-trailing-slash': "warning",
+      'oas2-valid-example': "warning",
+      'oas2-valid-definition-example': "error",
+      'oas2-anyOf': "warning",
+      'oas2-oneOf': "warning",
+      'oas3-api-servers': "warning",
+      'oas3-examples-value-or-externalValue': "warning",
+      'oas3-server-trailing-slash': "warning",
+      'oas3-valid-example': "error",
+      'oas3-valid-schema-example': "error"
+    }
   }
 };
 
--- a/src/cli-validator/utils/processConfiguration.js
+++ b/src/cli-validator/utils/processConfiguration.js
@@ -10,6 +10,8 @@ const defaultConfig = require('../../.defaultsForValidator');
 // global objects
 const readFile = util.promisify(fs.readFile);
 const defaultObject = defaultConfig.defaults;
+// Clear spectral rules from defaultObject, only use the explicit values in the validaterc file
+defaultObject.spectral.rules = {};
 const deprecatedRuleObject = defaultConfig.deprecated;
 const configOptions = defaultConfig.options;
 
@@ -42,6 +44,10 @@ const validateConfigObject = function(configObject, chalk) {
   const allowedSpecs = Object.keys(defaultObject);
   const userSpecs = Object.keys(configObject);
   userSpecs.forEach(spec => {
+    // Do not check "spectral" spec rules
+    if (spec === 'spectral') {
+      return;
+    }
     if (!allowedSpecs.includes(spec)) {
       validObject = false;
       configErrors.push({
--- a/src/spectral/utils/spectral-validator.js
+++ b/src/spectral/utils/spectral-validator.js
@@ -1,8 +1,9 @@
 const MessageCarrier = require('../../plugins/utils/messageCarrier');
 const config = require('../../cli-validator/utils/processConfiguration');
 const { isOpenApiv2, isOpenApiv3 } = require('@stoplight/spectral');
+const { mergeRules } = require('@stoplight/spectral/dist/rulesets');
 // default spectral ruleset file
-const defaultSpectralRuleset =
+const defaultSpectralRulesetURI =
   __dirname + '/../rulesets/.defaultsForSpectral.yaml';
 
 const parseResults = function(results, debug) {
@@ -46,7 +47,7 @@ const parseResults = function(results, debug) {
 };
 
 // setup: registers the oas2/oas3 formats, and attempts to load the ruleset file
-const setup = async function(spectral) {
+const setup = async function(spectral, configObject) {
   if (!spectral) {
     const message =
       'Error (spectral-validator): An instance of spectral has not been initialized.';
@@ -57,12 +58,22 @@ const setup = async function(spectral) {
   spectral.registerFormat('oas3', isOpenApiv3);
 
   // load the spectral ruleset, either a user's or the default ruleset
-  const spectralRuleset = await config.getSpectralRuleset(
-    defaultSpectralRuleset
+  const spectralRulesetURI = await config.getSpectralRuleset(
+    defaultSpectralRulesetURI
   );
 
+  // Combine user ruleset with the default ruleset
+  // The defined user ruleset will take precendence over the default ruleset
+  // Any rules specified in both will have the user defined rule severity override the default rule severity
+  await spectral.loadRuleset([defaultSpectralRulesetURI, spectralRulesetURI]);
+
+  // Combine default/user ruleset with the validaterc spectral rules
+  // The validaterc rules will take precendence in the case of duplicate rules
+  const userRules = Object.assign({}, spectral.rules); // Clone rules
   try {
-    return await spectral.loadRuleset(spectralRuleset);
+    return await spectral.setRules(
+      mergeRules(userRules, configObject.spectral.rules)
+    );
   } catch (err) {
     return Promise.reject(err);
   }
--- a/test/spectral/tests/spectral-validator.test.js
+++ b/test/spectral/tests/spectral-validator.test.js
@@ -1,5 +1,11 @@
 const { getCapturedText } = require('../../test-utils');
 const spectralValidator = require('../../../src/spectral/utils/spectral-validator');
+const inCodeValidator = require('../../../src/lib');
+const oas3InMemory = require('../mockFiles/oas3/enabled-rules-in-memory');
+const path = require('path');
+const config = require('../../../src/cli-validator/utils/processConfiguration');
+const defaultConfig = require('../../../src/.defaultsForValidator');
+const defaultObject = defaultConfig.defaults;
 
 describe('spectral - test spectral-validator.js', function() {
   let consoleSpy;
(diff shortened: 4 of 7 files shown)


src/cli-validator/utils/processConfiguration.js
```diff
@@ -42,102 +42,112 @@ const validateConfigObject = function(configObject, chalk) {
...
-      return; // skip rules for categories for invalid spec
-    }
-
-    // check that all categories are valid
-    const allowedCategories = Object.keys(defaultObject[spec]);
-    const userCategories = Object.keys(configObject[spec]);
-    userCategories.forEach(category => {
-      if (!allowedCategories.includes(category)) {
+    // Do not check "spectral" spec rules
+    if (spec !== 'spectral') {
```
rmkeezer, line 46: This entire file is just indention changes with a single if statement, not sure why the github diff shows it so poorly

src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -1,6 +1,6 @@
 extends: [[spectral:oas, off]]
 formats: [oas2, oas3]
-functionsDir": ../functions
+functionsDir: ../functions
```
jorge-ibm, line 3: Surprised spectral didn't complain about that

src/spectral/utils/spectral-validator.js
```diff
@@ -57,12 +62,27 @@ const setup = async function(spectral) {
...
-    defaultSpectralRuleset
+  const spectralRulesetURI = await config.getSpectralRuleset(
+    defaultSpectralRulesetURI
   );
+  const spectralRuleset = await readRuleset(spectralRulesetURI);
 
+  // Combine user ruleset with the default ruleset
+  // The default ruleset will take precendence in any rules it defines (including extended rules)
+  const defaultRuleset = await readRuleset(defaultSpectralRulesetURI);
+  mergeRules(spectralRuleset.rules, defaultRuleset.rules);
```
jorge-ibm, line 73: Might be worth trying to see if you can achieve the same merging by simply calling ` await spectral.loadRuleset([spectralRuleset, defaultSpectralRulesetURI]);`

rmkeezer: Yup this does achieve the same, I wasn't confident it did at first but it seems to be working as expect! I updated the PR with this

jorge-ibm (changes requested):
It's looking better, just a couple of more things that need to be changed. Also, if possible, some unit tests to test the validaterc changes would be helpful.

src/.defaultsForValidator.js
```diff
@@ -106,6 +106,30 @@ const defaults = {
     'schemas': {
       'json_or_param_binary_string': 'warning'
     }
+  },
+  'spectral': {
```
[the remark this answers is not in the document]
rmkeezer: Since i'm calling `getConfigObject` if there is no spectral section in the validaterc, it will take the rules from the spectral rules from `.defaultsForValidator.js` and these changes also allow the `init` command to generate the new spectral section

src/spectral/utils/spectral-validator.js
```diff
@@ -57,12 +57,19 @@ const setup = async function(spectral) {
...
   // load the spectral ruleset, either a user's or the default ruleset
-  const spectralRuleset = await config.getSpectralRuleset(
-    defaultSpectralRuleset
+  const spectralRulesetURI = await config.getSpectralRuleset(
+    defaultSpectralRulesetURI
   );
 
+  // Combine user ruleset with the default ruleset
+  // The default ruleset will take precendence in any rules it defines (including extended rules)
+  await spectral.loadRuleset([spectralRulesetURI, defaultSpectralRulesetURI]);
```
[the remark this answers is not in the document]
rmkeezer: I dont think there is a problem with loading a ruleset twice since it won't create any duplicates. Currently it will load the user config and the `.defaultsForSpectral.yaml` config or the `.defaultsForSpectral.yaml` config twice, how should this be changed? I can change it to load the `.defaultsForSpectral.yaml` first, should I make any other changes?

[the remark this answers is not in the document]
jorge-ibm: Okay. Yes, I think we should load the ``.defaultsForSpectral.yaml `` first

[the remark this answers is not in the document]
rmkeezer: Ok i updated it to load the defaults first

jorge-ibm:
@rmkeezer I'm not sure if my suggestion to use ``loadRuleset`` vs ``mergeRules`` broke something, but I pulled your fork locally and wasn't able to overwrite the severity level of our rules using the validaterc file.

rmkeezer:
@jorge-ibm Ah, apparently `spectral.mergeRules` does not work but using the `mergeRules` function from the spectral ruleset does. I've updated the PR with that change, it should now correctly override the severity again.

jorge-ibm:
Looks like everything is working as expected now, though I still think some tests would be helpful. The only thing I think should be fixed before merging is this duplicate message displaying whenever the validaterc file isn't present:
```
JorgeWoComputer:openapi-validator-rkeezer jrangel$ lint-openapi -p test/spectral/mockFiles/oas3/enabled-rules.yml 
[Warning] No .validaterc file found. The validator will run in default mode.
To configure the validator, create a .validaterc file.
[Warning] No .validaterc file found. The validator will run in default mode.
To configure the validator, create a .validaterc file.
```

I'm assuming we want to add documentation on how to enable/edit our spectral rules from the validaterc file. @dpopp07 what are your thoughts?

jorge-ibm (changes requested):
Sorry for requesting another set of changes, but I think this PR is really close.

jorge-ibm (approved):
LGTM. I'll approve but we should hold off on merging until @dpopp07 or @mkistler review it as well

src/spectral/utils/spectral-validator.js
```diff
@@ -46,7 +48,7 @@ const parseResults = function(results, debug) {
 };
 
 // setup: registers the oas2/oas3 formats, and attempts to load the ruleset file
-const setup = async function(spectral) {
+const setup = async function(spectral, defaultMode = false) {
```
dpopp07, line 51: Is there anywhere `setup` would be called without specifying this value? It looks like the one place we call it, we provide the value. So this default seems redundant. If it can be called in other places, that's a different story!

src/spectral/utils/spectral-validator.js
```diff
@@ -57,12 +59,22 @@ const setup = async function(spectral) {
...
   // load the spectral ruleset, either a user's or the default ruleset
-  const spectralRuleset = await config.getSpectralRuleset(
-    defaultSpectralRuleset
+  const spectralRulesetURI = await config.getSpectralRuleset(
+    defaultSpectralRulesetURI
   );
 
+  // Combine user ruleset with the default ruleset
+  // The default ruleset will take precendence in any rules it defines (including extended rules)
+  await spectral.loadRuleset([defaultSpectralRulesetURI, spectralRulesetURI]);
```
dpopp07, line 68: So this handles the case in which users define their own configuration with the Spectral format? But we don't let them override our default Spectral rules with that format?

I think this is what we want but as someone not very familiar with Spectral, I find it a little confusing. I appreciate the comments here, they are helpful, but will you add a couple more to provide context for why this is there?

dpopp07 (changes requested):
This looks good! I left a few minor comments but the primary reason I am "requesting changes" is because we need to have some tests for this behavior to ensure it does what we expect it to and to prevent regressions as we keep working on the Spectral stuff

src/spectral/utils/spectral-validator.js
dpopp07: Okay, I found it 🙂 This happens [on this line](https://github.com/IBM/openapi-validator/blob/master/src/lib/index.js#L18) but we have the default value there as a variable - so I think it would be better to pass it in that let it be potentially redefined. This should be addressed before approval

src/spectral/utils/spectral-validator.js
jorge-ibm: Matthew correct me if I'm wrong but with this PR this is what the user can do now:
1. They can override the severity level and disable/enable our spectral rules from the validaterc file.
2. They can _add_ any new custom rules to the validator using a custom rule set. Even if they don't include our spectral rules in their ruleset file, our spectral rules will still be loaded.
3. If a user wants to override our spectral rules using their ruleset file, they would have to:
   - remove that spectral rule from the validaterc file
   - add a rule with the same name to their ruleset file

rmkeezer: 1. Correct
2. Correct
3. Correct, but I realized that the omitting the spectral rule from the validaterc file did _not_ remove it and still overrode the severity the user would specify.

This was due to the config.get() function getting the defaultObject when there was no validaterc found, this default object specified the same rules as the default validaterc and would take priority. So I added a new flag to prevent this defaultObject from overriding it so it acts as you should expect it to like 3.

package.json
```diff
@@ -8,7 +8,8 @@
   "scripts": {
     "link": "npm install -g",
     "unlink": "npm uninstall -g",
-    "test": "jest test/",
+    "test": "jest test/ && jest test/spectral/tests/spectral-validator.test.js --runInBand",
+    "test:seq": "jest test/spectral/tests/spectral-validator.test.js --runInBand",
```
rmkeezer, line 12: I added tests to check reading the default config .validaterc, and the user config .spectral.yml, but to test them I had the test create a temporary config file.

However, this interferes with other tests when run in parrallel, so I have this test file running sequentially after the regular tests.

Is there a better solution for tests like these or is it fine to run this file sequentially?

jorge-ibm: Maybe try to spy on the fs package? https://stackoverflow.com/questions/50066138/mock-fs-function-with-jest . Not sure, @dpopp07 might have more experience doing something like that.

test/spectral/tests/spectral-validator.test.js
```diff
@@ -108,3 +112,155 @@ describe('spectral - test spectral-validator.js', function() {
     );
   });
 });
+
+const sequential = process.argv.includes('--runInBand');
```
dpopp07, line 116: My only question on the tests is - is there a way to just look at what rules spectral is ready to use after it's defined? Then we could just call the setup function and make sure it does what we want it to. I think that would be much simpler.

That said, I do like the idea of having an end-to-end test to verify the behavior we're looking for in a real life scenario. We can discuss later

dpopp07 (changes requested):
Getting better but I think there are some simplifications to be made that I did not notice before, specifically around reading the config object. Let me know what you think

rmkeezer:
I updated the PR with @dpopp07 's suggestions for using the defaultObject instead of calling config.get() inside the spectral run. This worked well! I also updated the unit test to using jest.spyOn to replace the file URL and the returned config file with their respective spectral changes. So the test does not need to run sequentially anymore!

test/spectral/tests/spectral-validator.test.js
rmkeezer: I was able to spyOn the config files so the end-to-end test isn't as troublesome anymore. For non-end-to-end tests, it's a bit difficult with spectral since creating a ruleset needs a file first and add/modifying rules is a bit messy using spectral's functions.

src/spectral/utils/spectral-validator.js
```diff
@@ -57,12 +58,22 @@ const setup = async function(spectral) {
...
+    defaultSpectralRulesetURI
   );
 
+  // Combine user ruleset with the default ruleset
+  // The defined user ruleset will take precendence over the default ruleset
+  // Any rules specified in both will have the user defined rule severity override the default rule severity
+  await spectral.loadRuleset([defaultSpectralRulesetURI, spectralRulesetURI]);
+
+  // Combine default/user ruleset with the validaterc spectral rules
+  // The validaterc rules will take precendence in the case of duplicate ruless
```
dpopp07, line 71:
```suggestion
  // The validaterc rules will take precendence in the case of duplicate rules
```

src/cli-validator/utils/processConfiguration.js
```diff
@@ -10,6 +10,8 @@ const defaultConfig = require('../../.defaultsForValidator');
 // global objects
 const readFile = util.promisify(fs.readFile);
 const defaultObject = defaultConfig.defaults;
+// Clear spectral rules from defaultObject
```
dpopp07, line 13: I think this makes sense but I was confused at first - maybe just add a bit more info in the comment like 
```suggestion
// Clear spectral rules from defaultObject, only use the explicit values in the validaterc file
// the default values for spectral come from another file
```

jorge-ibm (commented):
Looks really good! My last recommendation would be to add at least 1 test case that uses the ``commandLineValidator`` vs the ``inCodeValidator`` to invoke the validator.

dpopp07 (approved):
Looks good! I left some extremely minor comments to address before merging but I think it's good otherwise

jorge-ibm (approved):

test/spectral/tests/spectral-validator.test.js
```diff
@@ -108,3 +114,157 @@ describe('spectral - test spectral-validator.js', function() {
...
+
+    // Ensure mockConig was called and revert it to its original state
+    expect(mockConfig).toHaveBeenCalled();
+    mockConfig.mockRestore();
+  });
+
+  // There should be no errors and 2 warnings for a rule not off
+  it('test no errors and no warnings', function() {
+    expect(validationResults.errors.length).toBe(0);
+    expect(validationResults.warnings.length).toBe(2);
```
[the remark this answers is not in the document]
rmkeezer: Yup, there is a warning for `Definition was declared but never used in document`, i'm clarifying the comment for this

mkistler (approved):
Looks good! 👍

PR #214, opened 2020-12-20
refactor: upgrade chalk package to latest level

mkistler:
While working on arf/planning-sdk-squad#2400, I discovered that the old version of chalk we were using in the validator does not support the newer and more rationale color constants in the latest version, so this PR is a simple upgrade to the latest version to make these new colors available.


christiancompton (approved):
LGTM!

PR #220, opened 2020-12-26
Rework Spectral configuration and add documentation

mkistler:
This PR makes some usability changes to the configuration of the Spectral validator such that users can easily extend our ruleset in a manner consistent with Spectral's base configuration mechanism.

Rather than disabling all rules and then selectively enabling those we want to keep, the default configuration file is changed to explicitly enable or disable all the `spectral:oas` "recommended" rules. This approach makes the setting "sticky" -- any rule not explicitly changed by the user retains the configuration we gave it.

In the process, I documented the rationale for each rule that is disabled -- with some rules giving no rationale because we should consider enabling them.

I also added the `--ruleset` command line option to allow the Spectral configuration file to be explicitly named, much like our own configuration file and consistent with Spectral's CLI.

I've added some tests for custom configurations and custom Spectral rules.  I have tested custom spectral rules that use functions and they do work, but I couldn't figure out how to test these (no way to mock the "functions" directory).


test/spectral/tests/info-and-hint.test.js
```diff
@@ -1,60 +0,0 @@
-const path = require('path');
```
dpopp07, line 1: Was this file deleted intentionally?

dpopp07 (approved):
I think all of these changes look great. I left one comment to make sure you weren't deleting a file you didn't mean to delete

test/spectral/tests/info-and-hint.test.js
mkistler: I renamed that file to `spectral-config.test.js` since I wanted to expand it to cover the new configuration options, but I guess I crossed some threshold of similarity since git now thinks I deleted and added a file.  But you can see the the original "info-and-hint" test is there in the new file.

PR #221, opened 2020-12-28
Add rule to check response code is 3-digit integer or default

### LINKED ISSUE #163: Validate Response status codes to be 3 digit
HTTP status codes are supposed to be [3 digit ](https://tools.ietf.org/html/rfc7231#section-6) and and it gets through the validator

i.e
```
        "responses": {
          "0": {
            "description": "Response on failure.",
            "schema": {
              "$ref": "#/definitions/ErrorReturnObject"
            }
          },
```
autorest validates to 
"Parsing error(s): Property '0' has not been defined and the schema does not allow additional properties. Path 'paths['/plans'].post.responses.0',

mkistler:
This PR adds a custom Spectral rule to validate the status code of responses is a 3-digit integer or "default".

I've added a short test for this.

Something missing from this PR is documentation for this new rule.  This will be the first Spectral rule we add over and above the standard `spectral:oas` rules, so we need to decide where to document such rules.  My current thinking is to create a Wiki page on the repo that documents these in style of the Spectral rules.  Then I suppose we should update the README.md to point to that page.

Fixes #163.

### DIFF
--- a/src/spectral/rulesets/.defaultsForSpectral.yaml
+++ b/src/spectral/rulesets/.defaultsForSpectral.yaml
@@ -92,3 +92,21 @@ rules:
   oas3-schema: off
   # Turn off - duplicates non-configurable validation in base validator
   oas3-unused-components-schema: off
+
+  #
+  # Custom rules
+  #
+
+  # Response code must be 3-digit integer or "default"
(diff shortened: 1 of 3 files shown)


mkistler:
Actually ... let's hold off on this one for now.  I'm going to submit to Spectral and see if they take it.  If not then we'll add it ourselves.

PR #222, opened 2020-12-29
Display rule names for messages in verbose mode

mkistler:
This PR adds the ability to display the associated rule name for each issue reported by the validator.  Since this makes the already lengthy output longer, this feature must be enabled with a command line option.  I chose to use the `verbose` option, but this does come at some cost -- namely that the `-v` that previously was an alias of `--version` is now aliased to `--verbose`.  Note that `--version` is still supported and works as previously.

The bulk of the content in this PR was originally developed by @barrett-schonefeld and submitted in PR #147 way back in February.  I just had to make a few changes, add some tests, and update the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,14 +99,16 @@ Once pulled, the container can be run directly, but mount a volume containing th
 ##### [options]
 -  -s (--report_statistics) : Print a simple report at the end of the output showing the frequency, in percentage, of each error/warning.
 -  -e (--errors_only) : Only print the errors, ignore the warnings.
+-  -j (--json) : Output results as a JSON object
 -  -d (--default_mode) : This option turns off [configuration](#configuration) and runs the validator in [default mode](#default-mode).
 -  -p (--print_validator_modules) : Print the name of the validator source file the error/warning was caught it. This can be helpful for developing validations.
 -  -n (--no_colors) : The output is colored by default. If this bothers you, this flag will turn off the coloring.
--  -v (--version) : Print the current semantic version of the validator
+-  -v (--verbose) : Increase verbosity of reported results.  Use this option to display the rule for each reported result.
 -  -h (--help) : This option prints the usage menu.
 -  -c (--config) <path/to/your/config> : Path to a validator configuration file.  If provided, this is used instead of .validaterc.
 -  -r (--ruleset) <path/to/your/ruleset> : Path to Spectral ruleset file, used instead of .spectral.yaml if provided.
 -  --debug : Enable debugging output.
+-  --version : Print the current semantic version of the validator
 
 _These options only apply to running the validator on a file, not to any commands._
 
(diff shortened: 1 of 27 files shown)


src/cli-validator/utils/printResults.js
```diff
@@ -94,7 +99,13 @@ module.exports = function print(
...
 
   // print the stats here, if applicable
-  if (reportingStats && (stats.errors.total || stats.warnings.total)) {
+  if (
+    reportingStats &&
+    (stats.errors.total ||
+      stats.warnings.total ||
+      stats.infos.total ||
+      stats.hints.total)
+  ) {
```
mkistler, line 108: I snuck this change into this PR.  It actually addresses [this comment](https://github.com/IBM/openapi-validator/pull/219/files/91760490fa119aa1199dd2f91837e5a1422f3ee5#r549391425) on PR #219.

README.md
```diff
@@ -223,6 +225,7 @@ The supported categories are described below:
 #### Rules
 
 Each category contains a group of rules. The spec that each rule applies to is marked in the third column. For the actual configuration structure, see the [default values](#default-values).
+You can use the `-v` option of the CLI validator to display the rule for each reported result.
```
dpopp07, line 228: This may not be the place for it but I think it may be helpful to mention in the README somewhere that messages with rule "builtin" cannot be customized.

src/spectral/utils/spectral-validator.js
```diff
@@ -28,16 +28,16 @@ const parseResults = function(results, debug) {
         if (typeof severity === 'number' && code && message && path) {
           if (severity === 0) {
             // error
-            messages.addMessage(path, message, 'error');
+            messages.addMessage(path, message, 'error', code);
```
dpopp07, line 31: I admit I don't know what the values for `code` look like but if it's not obvious that they are Spectral rules, it may lead to confusion as to where that "key" is in the ruleset. Perhaps we should add a distinction to this string and document that Spectral rules are configured a different way, so that users can figure out what to do when they see a Spectral rule.

dpopp07 (approved):
This looks great! I left a couple minor comments and one question, but I'll go ahead and approve to let you decide if anything should be changed

src/cli-validator/index.js
```diff
@@ -47,11 +47,20 @@ program
...
+  )
   .option(
     '--debug',
     'enable debugging output'
   );
 
+function increaseVerbosity(dummyValue, previous) {
+  return previous + 1;
+}
+
```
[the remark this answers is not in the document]
mkistler: Yes, that was the idea.  I've seen this used elsewhere (e.g. `git branch -vv`) and thought it might be useful someday.  And it was cheap to add.

src/spectral/utils/spectral-validator.js
mkistler: I'm hesitant to change the code displayed since that weakens the connection to the actual Spectral rule, but I sympathize with your concern so I've added some text to the README to hopefully make this clear.

PR #224, opened 2021-01-03
Specify rule name for all configurable rules

mkistler:
This PR adds the rule name to the `addMessage` call for some cases that were missed in PR #222.


dpopp07 (approved):
Looks good! 👍

PR #226, opened 2021-01-06
docs: add further justification for inline response schema warnings

### LINKED ISSUE #125: Unexpected $ref warning when a definition is declared and used once
Validator v0.17.0 with the following v3 OpenAPI specification:
```yaml
openapi: 3.0.0
info:
  description: Pancakes are the best!
  version: 1.33.7
  title: Pancakes API
paths:
  /pancakes:
    get:
      operationId: list_pancakes
      summary: Returns all pancakes.
      description: For pancakes...
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  name:
                    description: The name of the pancake.
                    type: string
```

I see the following warning:
```bash
warnings

  Message :   Response schemas should be defined with a named ref.
  Path    :   paths./pancakes.get.responses.200.content.application/json.schema
  Line    :   17

statistics

  Total number of errors   : 0
  Total number of warnings : 1

  warnings
   1 (100%) : Response schemas should be defined with a named ref.
```

I'm trying to understand whether this is a valid warning, as based on the ["Using $ref"](https://swagger.io/docs/specification/using-ref/) documentation:

> it is common to have some features which you use across several of API resources. In that case, you can create a snippet for such elements in order to use them multiple times when you need it.

it seems that `$ref` was introduced to follow DRY principle, however in this case I can't see the validity of this being a warning as the definition is declared and used only once.

Probably, the desired behaviour would be to check for duplicated definitions before raising the named ref warning.

dpopp07:
Resolves #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ The supported rules are described below:
 ##### [responses][4]
 | Rule                      | Description                                                  | Spec |
 | ------------------------- | ------------------------------------------------------------ | ---- |
-| inline_response_schema    | Flag any response object with a schema that doesn't reference a named model. | shared |
+| inline_response_schema    | Flag any response object with a schema that doesn't reference a named model. Even if the model is only used once, naming it offers significant benefits for SDK generation. | shared |
 | no_response_codes         | Flag any response object that has no valid response codes.   | oas3 |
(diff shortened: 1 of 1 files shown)


mkistler (approved):
Looks good! 👍

PR #227, opened 2021-01-08
fix: allow null values in 'default' property

### LINKED ISSUE #155: Unable to set default value to null
Hi,

First off, I love the linting capabilities of this project, it is so much more valuable than just straight spec validation to ensure consistent API conventions and encourage strong documentation within the spec.

My project uses null values to indicate a property has never been configured and many properties default to null if not set.

When I run validation on my schema file I'm getting errors like:
>  Message :   Null values are not allowed for any property.
>  Path    :   components.schemas.ConnectId.properties.profile_picture.default
>  Line    :   234

That property is defined in my schema as
> profile_picture:
>   type: string
>   nullable: true
>   default: null
>   description: "A base64 encoded image string for the user's avatar"

I went looking for a linter option to change the behavior and found null values are just globally considered errors and there is no option to allow null as a default value.  See https://github.com/IBM/openapi-validator/blob/master/src/plugins/validation/2and3/semantic-validators/walker-ibm.js#L51

Based on this clarification of the `nullable` keyword in OAS 3.0.3 I believe my use case is valid.
https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/003_Clarify-Nullable.md#is-null-allowed-as-the-default-value-of-a-nullable-schema-see-2057

christiancompton:
Closes #155.

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/walker-ibm.js
+++ b/src/plugins/validation/2and3/semantic-validators/walker-ibm.js
@@ -52,7 +52,7 @@ module.exports.validate = function({ jsSpec, resolvedSpec }, config) {
 
     // check for and flag null values - they are not allowed by the spec and are likely mistakes
     Object.keys(obj).forEach(key => {
-      if (obj[key] === null) {
+      if (key !== 'default' && obj[key] === null) {
         messages.addMessage(
           [...path, key],
           'Null values are not allowed for any property.',
--- a/test/plugins/validation/2and3/walker-ibm.test.js
+++ b/test/plugins/validation/2and3/walker-ibm.test.js
@@ -97,7 +97,7 @@ describe('validation plugin - semantic - walker-ibm', () => {
     expect(res.warnings.length).toEqual(0);
   });
 
-  it('should return an error when a property contains a null value', () => {
+  it('should return an error when a non-default property contains a null value', () => {
     const spec = {
       paths: {
         '/pets': {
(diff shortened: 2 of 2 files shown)


src/plugins/validation/2and3/semantic-validators/walker-ibm.js
```diff
@@ -49,17 +49,6 @@ module.exports.validate = function({ jsSpec, resolvedSpec }, config) {
...
-    // check for and flag null values - they are not allowed by the spec and are likely mistakes
-    Object.keys(obj).forEach(key => {
-      if (obj[key] === null) {
-        messages.addMessage(
-          [...path, key],
-          'Null values are not allowed for any property.',
-          'error'
-        );
-      }
-    });
```
dpopp07, line 62: Rather than removing the check completely, I wonder if we should keep the check but allow `null` values on certain fields. From what I understand, null values are allowed on the `default` field or any annotations (fields that start with `x-`). 

Or, at the very least, just make it configurable so that users can turn it off if they need. We may want to make this a warning too, if our justification is they "are likely mistakes".

My suggestion might be to do all 3: allow certain fields, make it configurable, change the default to a warning. What do you think @mkistler ?

dpopp07 (changes requested):
This check was intentionally added so I am wary of removing it outright. If we discussed this and decided to remove the check, let me know (in that case, I forgot). Otherwise, let me know what you think of the suggestions I left

src/plugins/validation/2and3/semantic-validators/walker-ibm.js
mkistler: I would also like to see a more surgical fix.  We can address #155 by simply changing the condition in line 55 to
```
      if (key !== "default" && obj[key] === null) {
```

mkistler (changes requested):
Let's just make an exception for `default`.

christiancompton:
I have implemented the requested changes to ignore the warnings only for the "default" property.

dpopp07 (approved):
Looks good to me 👍

mkistler (approved):
Looks good! 👍

PR #228, opened 2021-01-08
chore: decrease severity of valid schema/content example validation

jorge-ibm:
This PR changes the severity of the spectral rules ``oas2-valid-definition-example`` and ``oas3-valid-schema-example`` from ``error`` to ``warn``. The spectral rule ``oas3-valid-oas-content-example`` was also enabled with a ``warn`` severity, since the linked issue reported our validator was not reporting warnings in request or response body examples. 

The ``src/.defaultsForValidator.js`` file was also updated to match the values in our default spectral ruleset.

fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/2303

### DIFF
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -120,15 +120,14 @@ const defaults = {
       'oas2-api-host': "warning",
       'oas2-api-schemes': "warning",
       'oas2-host-trailing-slash': "warning",
-      'oas2-valid-example': "warning",
-      'oas2-valid-definition-example': "error",
+      'oas2-valid-definition-example': "warning",
       'oas2-anyOf': "warning",
       'oas2-oneOf': "warning",
       'oas3-api-servers': "warning",
       'oas3-examples-value-or-externalValue': "warning",
       'oas3-server-trailing-slash': "warning",
-      'oas3-valid-example': "error",
-      'oas3-valid-schema-example': "error"
+      'oas3-valid-oas-content-example': "warning",
+      'oas3-valid-schema-example': "warning"
     }
   }
 };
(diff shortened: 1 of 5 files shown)


src/.defaultsForValidator.js
```diff
@@ -120,15 +120,14 @@ const defaults = {
...
       'oas2-host-trailing-slash': "warning",
-      'oas2-valid-example': "warning",
-      'oas2-valid-definition-example': "error",
+      'oas2-valid-definition-example': "warning",
       'oas2-anyOf': "warning",
       'oas2-oneOf': "warning",
       'oas3-api-servers': "warning",
       'oas3-examples-value-or-externalValue': "warning",
       'oas3-server-trailing-slash': "warning",
-      'oas3-valid-example': "error",
```
dpopp07, line 130: It looks like this rule is dropped altogether - is that intentional?

I'm guessing it was intention, just double checking because I'm not as familiar with the Spectral rules

dpopp07 (approved):
This looks good to me as is 👍 

I just left one question to double check something

src/.defaultsForValidator.js
jorge-ibm: Well I removed it from the defaultsForValidator.js because it looks like it was removed from our default spectral ruleset.. @mkistler did you accidentally remove this rule or was there a reason why we didn't want this rule in our ruleset ?

mkistler: I removed this because there isn't a Spectral rule with this name in the current version of Spectral.  I'm guessing that it was there in the past but then broken up into the various `oas-valid-<thing>-example` rules.

I think dropping it here is fine and actually I think we should consider removing support for configuring Spectral rules via `.validaterc` altogether.  We've never documented this, and now that the Spectral configuration is clearly documented, I'm not sure this adds enough value to justify the added complexity and maintenance burden.

mkistler (approved):
Looks good! 👍

PR #230, opened 2021-01-26
feat: exclude common inconsistent property names

### LINKED ISSUE #181: inconsistent_property_type is too aggressive, with no individual way to skip
Especially where generic, context-sensitive property names are used, such as `code`, `inconsistent_property_type` property type can lead to false positives across unrelated schemas. I believe we ought to rethink how this check works, or provide a way to skip the check for specific properties.

christiancompton:
Closes #181 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ The default values for each rule are described below.
 | no_property_description     | warning |
 | description_mentions_json   | warning |
 | array_of_arrays             | warning |
-| inconsistent_property_type  | warning |
+| inconsistent_property_type  | warning, [code, default, type, value]]<br>(list of property names to exclude)|
 | property_case_convention    | error, lower_snake_case |
 | property_case_collision     | error   |
 | enum_case_convention        | warning, lower_snake_case |
--- a/src/.defaultsForValidator.js
+++ b/src/.defaultsForValidator.js
@@ -65,7 +65,7 @@ const defaults = {
       'no_property_description': 'warning',
       'description_mentions_json': 'warning',
       'array_of_arrays': 'warning',
-      'inconsistent_property_type': 'warning',
+      'inconsistent_property_type': ['warning', ['code', 'default', 'type', 'value']],
       'property_case_convention': [ 'error', 'lower_snake_case'],
       'property_case_collision': 'error',
       'enum_case_convention': [ 'warning', 'lower_snake_case'],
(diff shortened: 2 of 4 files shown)


src/.defaultsForValidator.js
```diff
@@ -65,7 +65,7 @@ const defaults = {
       'no_property_description': 'warning',
       'description_mentions_json': 'warning',
       'array_of_arrays': 'warning',
-      'inconsistent_property_type': 'warning',
+      'inconsistent_property_type': ['warning', 'code', 'default', 'type', 'value'],
```
dpopp07, line 68: I would make a suggestion for a small tweak here. I think the list of keys to exclude should be their own list. Right now, it looks to me like these values are separate arguments with potentially different purposes. Since they are all meant to be grouped as a list though, I think it makes more sense to have them be one argument in list form, like:

```js
['warning', ['code', 'default', 'type', 'value']]
```

README.md
```diff
@@ -483,7 +483,7 @@ The default values for each rule are described below.
 | no_property_description     | warning |
 | description_mentions_json   | warning |
 | array_of_arrays             | warning |
-| inconsistent_property_type  | warning |
+| inconsistent_property_type  | warning, code, default, type, value] |
```
dpopp07, line 486: For rules with config options like `warning, lower_snake_case` and `case_convention` in the name, it's fairly obvious what the configuration represents.

In this new case, it may not be as obvious. I think we should document what the list means. Something like

```suggestion
| inconsistent_property_type  | warning, [code, default, type, value] | (List of property names excluded for this check)
```

I don't know if that would screw up the table formatting though. Maybe there's another place we could document it.

src/.defaultsForValidator.js
hudlow, line 68: Would it make more sense this way?
```suggestion
      'inconsistent_property_type': ['warning', ['code', 'default', 'type', 'value']],
```

src/plugins/validation/2and3/semantic-validators/schema-ibm.js
```diff
@@ -617,23 +617,26 @@ function checkProperties(
...
           );
         }
       } else {
-        propertiesToCompare[key] = {
-          type: value.type,
-          path: contextPath.concat(['properties', key]).join('.'),
-          printed: false
-        };
+        if (configOption && !configOption.includes(key)) {
+          // add property if the name is not generic
```
dpopp07, line 633: This is really picky 😅, but users may exclude non-generic properties

```suggestion
          // add property if the name is not excluded
```

dpopp07 (changes requested):
This looks good, I think the approach is great. I left a few comments for very small changes that we can discuss if need be

README.md
christiancompton: I did need to modify this slightly to get the table to render properly

dpopp07 (approved):
Looks good! 👍

PR #237, opened 2021-02-17
Add validation for use of 101 status code

hudlow:
Currently `101` is not considered a success and therefore it's not valid for an operation to support `101` without also supporting a `2xx` code. In fact, we'd prefer to validate that `101` *MUST NOT* be used along with a `2xx` response code.

### DIFF
--- a/src/plugins/validation/oas3/semantic-validators/responses.js
+++ b/src/plugins/validation/oas3/semantic-validators/responses.js
@@ -77,12 +77,19 @@ module.exports.validate = function({ resolvedSpec }, config) {
           }
         }
         // validate all success codes
-        if (!successCodes.length) {
+        if (!successCodes.length && !("101" in obj)) {
           messages.addMessage(
             path,
             'Each `responses` object SHOULD have at least one code for a successful response.',
             config.no_success_response_codes,
             'no_success_response_codes'
+          );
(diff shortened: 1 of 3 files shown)


src/plugins/validation/oas3/semantic-validators/responses.js
```diff
@@ -77,13 +77,20 @@ module.exports.validate = function({ resolvedSpec }, config) {
           }
         }
         // validate all success codes
-        if (!successCodes.length) {
+        if (!successCodes.length && !("101" in obj)) {
```
mkistler, line 80: I think I would have implemented this differently.  I would have changed `getResponseCodes` to return `101` in `successCodes`, since we want to treat it like one.  I'm curious why you chose this implementation instead.

hudlow: I actually did that first but it would have required changes to the check below that verifies that 204 is the only success code sans body.

mkistler: I'm good with this then.  We shouldn't over-think this.

mkistler (approved):
Looks good! 👍

dpopp07 (approved):
Looks good! 👍

Issue #241, opened 2021-02-19
Items property is not validated properly for arrays


jormaechea:
### TL;DR;
Array items validation gives false positives for certain inputs.

### Extended version

Our team spotted a false positive in array properties validation.

Given a property in a schema with `type: array`, the validator demands that the `items` property is set and it's value must be of type object.
So far, so good. The problem arises when something that has a `typeof items === 'object'` comes to play. And for javascript, an array is an object.

Therefore, the following is being reported as valid when it's not:

```yaml
components:
  schemas:
    MySchema:
      type: array
      items:
        - type: string
```

I have already located the bug and if your good with it I can send a PR.

jormaechea:
Inspecting the code a little bit more I found more than a dozen of places where objects are not validated properly.

I'm thinking in making a `isObject` utility and patching everywhere.

dpopp07:
@jormaechea Thanks for reporting - those bugs are tricky to spot sometimes. We would absolutely welcome a PR to resolve!

jormaechea:
@dpopp07 you can close this whenever you want! 🚀

dpopp07:
Resolved in #242

PR #242, opened 2021-02-21
fix: object type validations

jormaechea:
This resolves Issue #241.

I made the following changes:

- [x] Fixed validation of schema's array properties with an invalid `items` property (and added a regression test)
- [x] Fixed multiple occurrences of `typeof something === 'object'` and `typeof something !== 'object'`
- [x] Fixed usages of `_.isObject`, replacing them with `_.isPlainObject` instead. This new method performs the desired validations. See docs of [isObject](https://lodash.com/docs/4.17.15#isObject) and [isPlainObject](https://lodash.com/docs/4.17.15#isPlainObject).

### DIFF
--- a/src/cli-validator/runValidator.js
+++ b/src/cli-validator/runValidator.js
@@ -3,6 +3,7 @@ const util = require('util');
 const fs = require('fs');
 const path = require('path');
 const readYaml = require('js-yaml');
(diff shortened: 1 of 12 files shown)


dpopp07 (approved):
Looks great! Thank you for the PR

PR #245, opened 2021-02-23
test: create in-memory mock object directly from mock file

barrett-schonefeld:
Addresses:
- Some Spectral test resources have to be maintained in parallel, which is time-consuming and error-prone

Changes:
- Instead of maintaining a hardcoded object, load the API definition into memory to prevent incongruities between the API definition and its in-memory representation

Tests:
- Update one test to account for difference between mock file and in-memory mock object


dpopp07 (approved):
This was a great idea, thanks for doing it. Looks good 👍

PR #247, opened 2021-02-24
chore: add .secrets.baseline

christiancompton:
This PR adds a file to ignore false positives codescan results.


dpopp07 (approved):
Looks good! 👍

PR #248, opened 2021-02-24
refactor: use Spectral to check for duplicate parameters

barrett-schonefeld:
Purpose:
- Simplify the validator ruleset by using an equivalent Spectral rule
- Avoid resolving spec in-place to ensure inCodeValidator does not modify user's original spec object

Changes:
- Enable the operation-parameters Spectral rule
- Remove the in-code validation for parameters with duplicate `in:` and `name:` properties
- Pass a copy of the original spec to the reference resolver to prevent modification of the original spec

Tests:
- Remove tests meant to ensure the operation-parameters test is disabled
- Remove invalid oas3 from `disable-rules.yml` meant to invoke operation-parameters errors
- Remove irrelevant tests for in-code validation
- Add test to ensure operation-parameters check catches duplicate parameters for oas3 document

### DIFF
--- a/test/spectral/mockFiles/swagger/enabled-rules.yml
+++ b/test/spectral/mockFiles/swagger/enabled-rules.yml
@@ -53,6 +53,39 @@ paths:
           description: "Validation exception"
         default:
           description: "Error"
+  /pet/{pet_id}:
+    get:
+      tags:
+        - pet
+      summary: Find pet by ID
+      description: Returns a single pet
+      operationId: get_pet_by_id
+      produces:
+      - "application/json"
+      parameters:
+      - name: pet_id
+        in: path
+        description: ID of pet to return
+        required: true
+        schema:
+          type: integer
+          format: int64
+      - name: pet_id
+        in: path
+        description: ID of pet to return
+        required: true
+        schema:
+          type: integer
+          format: int64
+      responses:
+        200:
+          description: successful operation
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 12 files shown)


test/spectral/mockFiles/swagger/enabled-rules.yml
```diff
@@ -166,6 +199,19 @@ definitions:
...
+      properties:
+        pet_id:
+          description: pet_id
+          type: integer
+          format: int64
+          example: 10
+        name:
+          description: pet name
+          type: string
+          example: doggie
```
mkistler, line 214: This is incredibly picky but it will demonstrate that I _did_ look somewhat carefully.  I think the section I highlighted is indented one level more than it needs to be.

mkistler (approved):
Looks good! 👍

test/plugins/validation/oas3/responses.test.js
```diff
@@ -645,7 +645,6 @@ describe('validation plugin - semantic - responses - oas3', function() {
     };
 
     const res = validate({ resolvedSpec: spec }, config);
-    console.log(res.warnings);
```
dpopp07, line 648: Good catch!

dpopp07 (approved):
Looks great! 👍

test/spectral/mockFiles/swagger/enabled-rules.yml
barrett-schonefeld: Yes, nice catch!

PR #249, opened 2021-02-25
refactor: use Spectral to check for query string in path

barrett-schonefeld:
Purpose:
- Simplify validator ruleset by using equivalent Spectral rule

Changes:
- Enable the path-not-include-query Spectral rule
- Remove the in-code validation for query string in paths

Tests:
- Remove irrelevant tests that ensure path-not-include-query Spectral rule disabled
- Remove irrelevant tests for in-code validation of in-path query strings
- Note: did not add enabled-rules test for path-not-include-query rule. Adding a query string to a path in the `enabled-rules.yml` test file also violates the validator's in-code paths_case_convention rule. This caused test failures that I thought would be ugly to fix, so thus far, we do not add a positive test for the path-not-include-query rule on our side.

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/paths.js
+++ b/src/plugins/validation/2and3/semantic-validators/paths.js
@@ -15,6 +15,7 @@
 
 // Assertation 6:
 // Paths cannot have literal query strings in them.
+// Handled by the Spectral rule, path-not-include-query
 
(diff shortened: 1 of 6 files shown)


mkistler (approved):
Looks good! 👍

dpopp07 (approved):
Looks good! 👍

PR #250, opened 2021-02-25
refactor: use Spectral to check for unused definitions

barrett-schonefeld:
Purpose:
- Simplify validator rulest by using equivalent Spectral rule

Changes:
- Enable the oas2-unused-definition and oas3-unused-components-schema Spectral rules
- Remove the in-code validation for unused definitions

Tests:
- Remove irrelevant tests used to ensure mentioned Spectral rules are disabled
- Remove unused definitions from test files used to ensure the mentioned Spectral rules are disabled
- Remove irrelevant tests for in-code validation for unused definitions


dpopp07 (approved):
Looks good! 👍

mkistler (changes requested):
You should put this one on ice for a bit.  It needs the change that I submitted to Spectral in this PR: 

https://github.com/stoplightio/spectral/pull/1440

The PR is merged but not yet pulled into a release.  When it is, we'll want to use the `oas3-unused-components` rule, as this will catch all the unused components -- not just schemas -- which will be equivalent to the current behavior.

PR #252, opened 2021-02-25
fix: bundle original spec to account for multi-file API definitions

### LINKED ISSUE #229: CLI does not dereference $refs
Perhaps this is by design, but i feel as this might deserve a mention in the readme.

Was scratching my head trying to figure out why the CLI and node module were turning out different statistics, the $refs being the culprit.

Provided that the node module expects a parsed object then it is up to the client to decide if the document should be dereferenced or not. Given that the CLI takes the file directly, then the dereferencing responsibility is (arguably) moved onto the CLI and its internal parser.

OR

Simply mention in the readme that the CLI will not dereference. Could perhaps be added as a option?

My two cents,

Herman

dpopp07:
We do some validations on the "resolved spec" or the full API definition with all of its
references resolved into one large document. This takes multi-file specs into account and
resolves all references, even to other files. However, some validations are performed only
on the original spec as it provides some benefits, such as gathering all schemas without
worrying about duplicates. The problem with our current approach is that, if the spec is
split into multiple files, we only validate the entry file and none of the others.

This adds a quick step to "bundle" all of the files together into one "file" that only has
internal references. This way, the unresolved "js spec" will represent the entire definition.

Resolves #229

### DIFF
--- a/src/cli-validator/utils/buildSwaggerObject.js
+++ b/src/cli-validator/utils/buildSwaggerObject.js
@@ -22,11 +22,16 @@ const schemas = {
 module.exports = async function(input) {
   // initialize an object to be passed through all the validators
   const swagger = {};
+  const parser = new RefParser();
+
+  // in case of multi-file specifications, bundle the spec into one file with
+  // only internal refs. use this for the spec string and the js spec
+  const bundledSpec = await parser.bundle(copyObject(input));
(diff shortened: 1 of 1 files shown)


barrett-schonefeld (approved):
Looks good! 👍

PR #255, opened 2021-03-02
refactor: replace info object check with a custom Spectral rule

barrett-schonefeld:
Outstanding problems to be resolved:
- When a user provides a Spectral configuration file, our custom rules will not be invoked. Considerations
    - Need to ensure our custom rules are invoked when the user provides a Spectral configuration file.
    - Need to ensure the custom rules have access to their corresponding custom functions. Need to consider that user's Spectral configuration file may be in different directories and adjust the path to the custom functions directory.
    - Need to ensure users are able to adjust the severity of custom rules. Users will be able to configure previously "builtin" rules.
    - May want to adjust the way we process user's Spectral config file to, first, process our default Spectral configuration file then load the user's Spectral configuration file and overwrite the default values. Simply put, I think it'd be better to merge the user's Spectral configuration file into the default Spectral configuration file.
    - Not my idea but a good idea: Have a `Spectral config file updater` to give users a command to add new rules to their Spectral configuration file.

Purpose:
- Simplify the validator ruleset by replacing good candidates with an equivalent rule
- Establish a pattern in the validator for adding custom spectral rules and custom spectral functions

Changes:
- Add custom spectral rule to ensure that an info object is provided in the spec
- Add custom spectral function
- Remove in-code validation of info object

Tests:
- Move existing tests to ensure the equivalent spectral rule produces the same behavior

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2164,18 +2164,18 @@
       }
     },
     "@stoplight/better-ajv-errors": {
-      "version": "0.0.0",
-      "resolved": "https://registry.npmjs.org/@stoplight/better-ajv-errors/-/better-ajv-errors-0.0.0.tgz",
-      "integrity": "sha512-jA1i4J5HmCXMmZdZigdeEHpisFCGVg5QfK++ouxP58pxshATJ5G40U9uGec2vYnuR2MQnzWHAXYf0f8jvJhBwg==",
+      "version": "0.0.3",
+      "resolved": "https://registry.npmjs.org/@stoplight/better-ajv-errors/-/better-ajv-errors-0.0.3.tgz",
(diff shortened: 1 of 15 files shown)


mkistler:
I think the general approach to adding custom rules and custom functions in this PR looks sound, but I don't think the rule to check for an `info` object is the right application.  The `info` object is **required** by the OpenAPI 3.0 specification, so the Spectral `oas3-schema` rule should issue an appropriate error if it is missing or is not an object.  So the rule defined here would be redundant with that Spectral rule -- something we should avoid.

Let's identify another rule to implement that won't duplicate a Spectral rule to prove out this approach.

barrett-schonefeld:
Closing this PR in favor of #258 because #258 both establishes a pattern for custom Spectral rules and implements a custom Spectral rule for which there is no existing, equivalent Spectral rule.

PR #256, opened 2021-03-03
feat: include the validator version in the JSON output

barrett-schonefeld:
Changes:
- Get the validator version from `package.json` and include it in the json output
- Update `package.json` and `package-lock.json` to resolve differences in Spectral versions now

Tests:
- Update tests to match latest Spectral behavior

### DIFF
--- a/test/cli-validator/tests/expected-output.test.js
+++ b/test/cli-validator/tests/expected-output.test.js
@@ -78,14 +78,44 @@ describe('cli tool - test expected output - Swagger 2', function() {
     expect(messageCount).toEqual(ruleCount);
   });
 
+  it('should include the validator version in JSON output', async function() {
+    const program = {};
+    program.args = ['./test/cli-validator/mockFiles/clean.yml'];
+    program.default_mode = true;
+    program.json = true;
+
+    const exitcode = await commandLineValidator(program);
+    expect(exitcode).toBe(0);
+
+    const capturedText = getCapturedText(consoleSpy.mock.calls);
+    const jsonOutput = JSON.parse(capturedText);
+    const expectedValidatorVersion = require('../../../package.json').version;
+    expect(jsonOutput.version).toBeTruthy();
+    expect(jsonOutput.version).toBe(expectedValidatorVersion);
+  });
+
+  it('should include the validator version in JSON output for the inCodeValidator', async function() {
+    const content = fs
+      .readFileSync('./test/cli-validator/mockFiles/clean.yml')
+      .toString();
+    const spec = yaml.load(content);
+
+    const defaultMode = true;
+    const validationResults = await inCodeValidator(spec, defaultMode);
+
+    const expectedValidatorVersion = require('../../../package.json').version;
+    expect(validationResults.version).toBeTruthy();
+    expect(validationResults.version).toBe(expectedValidatorVersion);
+  });
+
   it('should include the associated rule with each error and warning in JSON output', async function() {
     const program = {};
     program.args = ['./test/cli-validator/mockFiles/err-and-warn.yaml'];
     program.default_mode = true;
     program.json = true;
 
-    await commandLineValidator(program);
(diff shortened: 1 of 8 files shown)


mkistler (changes requested):
This looks good but I'd like to see a test that verifies that the version is included in the JSON output.

test/cli-validator/tests/expected-output.test.js
```diff
@@ -78,6 +78,22 @@ describe('cli tool - test expected output - Swagger 2', function() {
...
   });
 
+  it('should include the validator version in JSON output', async function() {
+    const program = {};
+    program.args = ['./test/cli-validator/mockFiles/err-and-warn.yaml'];
+    program.default_mode = true;
+    program.json = true;
+
+    await commandLineValidator(program);
+    // exit code is not set for JSON output
```
dpopp07, line 88: What do mean when you say the exit code isn't set? Any execution of the command line validator should return an exit code and if this doesn't happen (though I'm pretty sure it does in `src/cli-validator/index.js`) then that's a bug that should be fixed. Either way, I think we should check the exit code here

test/cli-validator/tests/info-and-hint.test.js
```diff
@@ -32,7 +32,7 @@ describe('test info and hint rules - OAS3', function() {
     consoleSpy.mockRestore();
 
     // errors for non-unique operation ids
-    expect(jsonOutput['errors']['spectral'].length).toBe(2);
+    expect(jsonOutput['errors']['spectral'].length).toBe(1);
```
dpopp07, line 35: Why did all of this stuff change (including the option-handling and enabled-rules tests)?

dpopp07 (changes requested):
This pretty much looks good as is, only requesting changes to verify that the CLI validator is in fact returning an exit code when JSON output is turned on

test/cli-validator/tests/info-and-hint.test.js
barrett-schonefeld: This changed because the Spectral, `operation-operationIds-unique` check, changed to not error on the first instance of a duplicate operation ID.

test/cli-validator/tests/expected-output.test.js
barrett-schonefeld: I used an adjacent test as a template. This assumption comes from the adjacent test. I will check the exit code and see if we need an issue outlining the exit code problem.

test/cli-validator/tests/info-and-hint.test.js
dpopp07: Ah right, I forgot you updated the Spectral version. Carry on then 🙂

test/cli-validator/tests/expected-output.test.js
dpopp07: Gotcha - and I probably wrote that test haha. Funny how things can seem appropriate in one moment and then not down the road. We do want to make sure the CLI validator _always_ returns a code but like I mentioned, everything is wrapped in the `index.js` file so there shouldn't be any issues. That said, let's check just in case.

test/cli-validator/tests/expected-output.test.js
```diff
@@ -78,14 +78,30 @@ describe('cli tool - test expected output - Swagger 2', function() {
...
   it('should include the associated rule with each error and warning in JSON output', async function() {
     const program = {};
     program.args = ['./test/cli-validator/mockFiles/err-and-warn.yaml'];
     program.default_mode = true;
     program.json = true;
 
-    await commandLineValidator(program);
-    // exit code is not set for JSON output
+    const exitcode = await commandLineValidator(program);
+    expect(exitcode).toBe(0);
```
dpopp07, line 104: Thanks for making this change!

dpopp07 (approved):
Looks good! 👍

mkistler (approved):
Looks good! 👍

PR #257, opened 2021-03-03
fix: no error for inline array schema when items field is a ref

barrett-schonefeld:
Purpose:
- generator flattens reference to an array schema back into an inline schema

Changes:
- Add a check to see if the schema is an array with an items field that uses a ref or a primitive type (boolean, integer, number, string)
  - Note: this update is only for response schemas. I looked through the different semantic validators and only found the `inline_response_schema` check. I did not find an equivalent check for inline request bodies.

Tests:
- Add tests to ensure the validator does not flag different variations of this array schema
- Update expected warnings in tests that use specs with this array schema

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/responses.js
+++ b/src/plugins/validation/2and3/semantic-validators/responses.js
@@ -1,9 +1,19 @@
 const each = require('lodash/each');
 const { walk, isResponseObject } = require('../../../utils');
 const MessageCarrier = require('../../../utils/messageCarrier');
+const isPrimitiveType = require('../../../utils/isPrimitiveType');
 
 const INLINE_SCHEMA_MESSAGE =
   'Response schemas should be defined with a named ref.';
+
+function arrayItemsAreRefOrPrimitive(schema) {
+  return (
+    schema &&
+    schema.type === 'array' &&
+    schema.items &&
+    (schema.items.$ref || isPrimitiveType(schema.items))
+  );
+}
 
 module.exports.validate = function({ jsSpec, isOAS3 }, config) {
   const messages = new MessageCarrier();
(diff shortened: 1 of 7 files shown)


src/plugins/validation/2and3/semantic-validators/responses.js
```diff
@@ -36,9 +40,9 @@ module.exports.validate = function({ jsSpec, isOAS3 }, config) {
                       i < mediaType.schema[schemaType].length;
                       i++
                     ) {
-                      const hasInlineSchema = !mediaType.schema[schemaType][i]
-                        .$ref;
-                      if (hasInlineSchema) {
+                      const currentSchema = mediaType.schema[schemaType][i];
+                      const hasInlineSchema = !currentSchema.$ref;
+                      if (hasInlineSchema && !arrayItemsAreRef(currentSchema)) {
```
mkistler, line 45: While this does address the problem as described in the issue, I don't think it is quite what we want.

In addition to allowing an array of $ref, I think we probably should also allow an array of primitive types.  That is, I do not think we should flag a response schema that is an array of `string` or an array of `integer`.

mkistler (changes requested):
This looks good but I think there is a twist that we need to consider.

src/plugins/validation/2and3/semantic-validators/responses.js
dpopp07: Agreed, though this will likely be more verbose than it would be in the generator (no `isPrimitiveType` to rely on). Perhaps you could write a utility function called isPrimitiveType or something like that so we can re-use it if this comes up again

mkistler (approved):
Looks good! 👍

dpopp07 (approved):
Looks good! 👍

PR #258, opened 2021-03-03
feat: custom Spectral rule to ensure content objects contain schema

barrett-schonefeld:
Purpose:
- Establish a pattern in the validator for adding custom spectral rules and custom spectral functions
- Ensure content objects contain a schema

Changes:
- Add custom spectral rule to ensure that content objects contain a schema
- Add directory to store custom spectral rules and functions
- Add code to retrieve and merge custom spectral rules

Tests:
- Add tests for each use case outlined in the spectral rule

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,8 @@ This ruleset has the alias `ibm:oas`, and you can "extend" this ruleset or speci
 with a [Spectral ruleset file](https://meta.stoplight.io/docs/spectral/docs/getting-started/3-rulesets.md).
 Note that all of the rules in the `spectral:oas` ruleset are defined in `ibm:oas` but only the rules listed above are enabled by default.
 
+The `ibm:oas` ruleset also includes custom Spectral rules, [documented here](docs/spectral-rules.md). These are configurable rules in the `ibm:oas` ruleset that are not part of the `spectral:oas` ruleset.
+
 You can provide a Spectral ruleset file to the IBM OpenAPI validator in a file named `.spectral.yaml`
 in the current directory or with the `--ruleset` command line option of the validator.
 
--- a/docs/spectral-rules.md
+++ b/docs/spectral-rules.md
@@ -1,0 +1,30 @@
+# Custom Spectral Rules
+
+This document outlines the custom Spectral rules implemented in `ibm:oas` ruleset.
+
+## content-entry-contains-schema
+
+Any request or response body that has `content` should contain a schema.
+
+**Bad Example**
+
+```yaml
+responses:
+  200:
+    content:
+      application/json:
+        # schema not provided
+```
+
+**Good Example**
+
+```yaml
+responses:
+  200:
+    content:
+      application/json:
+        schema:
+          type: string
+```
+
+**Default Severity**: warn
--- a/src/spectral/rulesets/.defaultsForSpectral.yaml
+++ b/src/spectral/rulesets/.defaultsForSpectral.yaml
@@ -92,3 +92,15 @@ rules:
   oas3-schema: off
   # Turn off - duplicates non-configurable validation in base validator
   oas3-unused-components-schema: off
+
+  # custom Spectral rule to ensure content object contains schema
+  content-entry-contains-schema:
+    description: Content entries in request and response bodies must specify a schema
+    given:
+    - $.paths[*].[post,put,patch].requestBody.content[*]
+    - $.paths[*].[get,post,put,patch,delete].responses[*].content[*]
+    severity: warn
+    resolved: true
+    then:
+      field: schema
+      function: truthy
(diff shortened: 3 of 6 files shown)


mkistler (changes requested):
This does seem to satisfy the requirement but I think it can be simplified while still achieving the same result.

src/spectral/rulesets/custom-rules/content-entry-contains-schema.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
```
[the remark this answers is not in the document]
barrett-schonefeld: In hindsight, I think it may not be necessary to break the rules into different files. In different conversations about adding custom Sepctral rules, people mentioned that we may not want to clutter the `.defaultsForSpectral.yaml` file with custom rules.

I would say breaking custom rules in separate files has the following benefits:

1. Keeps the `.defaultsForSpectral.yaml` small and digestable.
2. Test files related to each custom rule are easier to create/find/update because test files map to a corresponding file with the same name.
3. Small benefit: When we add custom functions, some rules will need to specify the custom function it uses in a `functions` field. If the custom rules are in separate files, each real will clearly define the list of custom functions it uses rather than have a global list of custom functions in `.defaultsForSpectral.yaml`. This is only a small benefit because the custom rule will explicitly say which custom function it uses in the rule definition.

Do you think the reasons above are enough to introduce a more complicated structure now, or should we consider breaking rules into separate files as it becomes apparent that it's necessary?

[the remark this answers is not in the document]
mkistler: My take:
1. This just trades complexity within one file for the complexity of having multiple files -- which includes the extra logic in this PR for merging the multiple files.
2. Test files can be named for the rule -- I think this provides equal correlation between the rules and test files.
3. Valid but I'm not sure sufficient rationale for separate files.

I'll add to this that Spectral follows the "all rules in a single file" structure:

https://github.com/stoplightio/spectral/blob/develop/src/rulesets/oas/index.json

So I still think the single file approach is superior.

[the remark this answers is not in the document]
barrett-schonefeld: Yes, I think you're right, and I will make the updates to move forward with a single file approach.

src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -92,3 +92,15 @@ rules:
...
   # Turn off - duplicates non-configurable validation in base validator
   oas3-unused-components-schema: off
+
+  # custom Spectral rule to ensure content object contains schema
+  content-entry-contains-schema:
+    description: Content entries in request and response bodies must specify a schema
+    given:
+    - $.paths[*].[post,put,patch].requestBody.content[*]
+    - $.paths[*].[get,post,put,patch,delete].responses[*].content[*]
+    - $.components.responses[*].content[*]
```
dpopp07, line 102: I'm not that well versed in Spectral syntax but didn't Mike make a comment earlier about how `components` didn't need to be checked? Do you know if this line is necessary?

dpopp07 (approved):
I think this looks good but I did have one question about the Spectral implementation to check on before merging (might want to wait for other reviews as well)

src/spectral/rulesets/.defaultsForSpectral.yaml
barrett-schonefeld: Yes, thank you for catching!

mkistler (changes requested):
The code here looks good but another key part of creating our own custom rules is documenting them.  We probably need to add a new section to the README that describes each custom rule, like the `docs/reference/openapi-rules.md` in Spectral.

barrett-schonefeld:
> The code here looks good but another key part of creating our own custom rules is documenting them. We probably need to add a new section to the README that describes each custom rule, like the `docs/reference/openapi-rules.md` in Spectral.

To keep the `README.md` from getting overcrowded with rule descriptions, I documented the rule in `docs/spectral-rules.md` and linked to the rule descriptions in the `README.MD`.

README.md
```diff
@@ -572,6 +572,8 @@ You could also set the severity of `info-contact` explicitly to `error`, `warn`,
 
 #### Custom Spectral rules
 
+The Validator implements some validations as custom Spectral rules, [documented here](docs/spectral-rules.md).
```
mkistler, line 575: I think this is in the wrong place.  From the user's point of view, the rules we are adding are "standard rules" of our "ibm:oas" ruleset.  So this belongs in the previous section.

docs/spectral-rules.md
```diff
@@ -0,0 +1,28 @@
+# Custom Spectral Rules
+
+This document outlines the custom Spectral rules implemented in `src/spectral/rulesets/.defaultsForSpectral.yaml`.
```
mkistler, line 3:
```suggestion
This document outlines the custom Spectral rules implemented in the `ibm:oas` ruleset.
```

mkistler (changes requested):
Almost done, but we need to document this from the point of view of the user.

mkistler (approved):
Looks good. 👍

PR #260, opened 2021-03-08
feat: synchronize the inCodeValidator output to match the CLI -json

barrett-schonefeld:
Purpose:
- Both the inCodeValidator and CLI -json return objects, so we want to standardize and keep these formats in sync.
- Make output changes to fewer places when introducing new features. Simplify the code base.

Changes:
- Add `infos` and `hints` to the JSON output.
- Split the object creation and the printing of the JSON object into separate functions, so the `inCodeValidator` may utilize the same object-creation function.

Tests:
- Add a test utility that aggregates the inCodeValidator errors, warnings, etc. into a single array to convert output to a format that is still compatible with existing tests.
- Remove checks that expect the `infos` and `hints` fields to be undefined.


dpopp07:
@barrett-schonefeld I see you've included a "BREAKING CHANGE" notice in your commit message. We need to be careful here, because this will trigger a major release of the tool, aka 1.0.0. A formal "1.0" release is something that we want to plan out more and include more changes in, so I don't think we're ready for that. While your change here is "breaking", breaking changes are allowed in any 0.x version, so we don't need to make it formal. It can just be a feature release.

You will probably need to amend your commit and force push to make sure that line isn't included when you merge.

dpopp07 (approved):
I like this! Thanks for making these changes. Looks good to me 👍

mkistler (approved):
Looks good! 👍 

I particularly like `+91 -199`.  A cleaner, simpler design means less code to maintain.

PR #261, opened 2021-03-10
chore: update appscan config

christiancompton:
This issue excludes the README file from the ASOC codescanning system, which is flagging the Docker command instructions as security vulnerabilities.


rmkeezer (approved):
LGTM

PR #262, opened 2021-03-15
feat: remove error for partial path templating

barrett-schonefeld:
We could implement the check for partial path templating as a configurable Spectral rule. This rule, however, will need a custom Spectral function. The initial work for adding a custom Spectral function is on hold until we determine how to ensure that users can extend our ruleset without having problems locating our custom functions. Should we resolve the custom functions problem and implement this check as a Spectral rule, or should we move this work through and remove the check?

Purpose:
- Partial path templating is not disallowed in OAS3, so although partial path templating may not be ideal style, do not want it to be an error.

Changes:
- Remove code that checks for partial path templating.

Tests:
- Remove tests that test partial path template validation code.

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/paths.js
+++ b/src/plugins/validation/2and3/semantic-validators/paths.js
@@ -11,9 +11,6 @@
 // Paths must have unique (name + in combination) parameters
 
 // Assertation 5:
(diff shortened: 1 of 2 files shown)


mkistler (approved):
Looks good! 👍

dpopp07 (approved):
👍

PR #264, opened 2021-03-16
fix: correct the minimum required version of Node

barrett-schonefeld:
Changes:
- Node version requires all three major, minor, patch fields, so update minimum version from 10.0 to 10.0.0


dpopp07 (approved):
Looks good 👍

PR #265, opened 2021-03-16
feat: warn for missing response examples

barrett-schonefeld:
Purpose:
- Establish working pattern for writing custom functions such that, when users extend the `ibm:oas` ruleset, our custom functions are still findable by Spectral.
- Tooling to generate documentation relies on response examples provided either in the schema or "next to" the schema at response level. Warn about responses that do not provide examples in these locations.

Changes:
- Create a custom Spectral rule to warn about missing response examples
- Create a custom Spectral function to check if a response example provided in the locations supported by the code that generates documentation.

Tests:
- Add tests to ensure no warning given when a response example is provided.
- Add a test to ensure a warning is given when no response example provided.

Docs:
- Document the response-example-provided rule in `docs/spectral-rules.md`

**Possible changes:**
- Include more than `application/json` responses. To do this, we would change the `given` field of the rule from `$.paths[*][*].responses[*].content.application/json` to `$.paths[*][*].responses[*].content[*]`.
- May want to condense the error message. I had writer's block, so I went with a verbose description initially.

### DIFF
--- a/src/spectral/rulesets/ibm-oas/response-example-provided.js
+++ b/src/spectral/rulesets/ibm-oas/response-example-provided.js
@@ -1,0 +1,17 @@
+module.exports = function(response) {
+  if (!responseLevelExamples(response) && !schemaLevelExample(response)) {
+    return [
+      {
+        message: 'Response bodies should include an example response'
+      }
+    ];
+  }
+};
+
+function responseLevelExamples(response) {
+  return response.example || response.examples;
+}
+
+function schemaLevelExample(response) {
+  return response.schema && response.schema.example;
+}
--- a/test/spectral/tests/spectral-config.test.js
+++ b/test/spectral/tests/spectral-config.test.js
@@ -1,37 +1,23 @@
-const path = require('path');
 const commandLineValidator = require('../../../src/cli-validator/runValidator');
-const config = require('../../../src/cli-validator/utils/processConfiguration');
 const { getCapturedText } = require('../../test-utils');
 
 describe('Spectral - test custom configuration', function() {
   it('test Spectral info and hint rules', async function() {
-    // Set config to mock .spectral.yml file before running
-    const mockPath = path.join(
-      __dirname,
-      '../mockFiles/mockConfig/info-and-hint.yaml'
-    );
-    const mockConfig = jest
-      .spyOn(config, 'getSpectralRuleset')
-      .mockResolvedValue(mockPath);
-
     const consoleSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
+
     // set up mock user input
     const program = {};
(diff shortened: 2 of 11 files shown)


src/spectral/rulesets/ibm-oas/response-example-provided.js
```diff
@@ -0,0 +1,18 @@
+module.exports = function(response) {
+  if (!responseLevelExamples(response) && !schemaLevelExample(response)) {
+    return [
+      {
+        message:
+          'To generate good documenation, should provide response example(s) inside of or at the the same level as the schema'
```
dpopp07, line 6: This is a little long, but it's probably fine. I will make some minor suggestions:

```suggestion
          'To generate good documentation, provide response example(s) inside of, or at the the same level as, the schema'
```

dpopp07 (approved):
Looks good to me!

src/spectral/rulesets/ibm-oas/response-example-provided.js
mkistler: How about:
```suggestion
          'Response bodies should include an example response'
```

I don't think we need "To generate good documentation" -- that is the goal of many of our rules (like missing descriptions) but we don't explicitly state that in the message.

And I don't think the message needs to say where the example should be specified -- we say that in the documentation of the rule.

test/spectral/tests/spectral-config.test.js
```diff
@@ -1,37 +1,23 @@
-const path = require('path');
```
mkistler, line 1: It looks like you snuck some unrelated changes into this PR.  I don't have a problem with that but I think you should alert reviewers to this by either mentioning it in the PR description or simply commenting on your own PR to call this out.

mkistler (changes requested):
This looks good but there's one more detail that you will need to handle.  In OpenAPI 3.0 they introduced `examples` on parameters, request bodies, and response bodies.  `examples` is a slightly more elaborate way to provide an example -- you can include not just the example value but also a description and summary.  You can also have multiple named examples.

So in addition to checking for `example` you should also check for `examples` and only issue the message if neither is present.

test/spectral/tests/spectral-config.test.js
barrett-schonefeld: Yes, I will do a better job of pointing out and explaining these other changes. These tests were timing out unpredictably (maybe 1/10 or 1/20 times). The mocking seemed to cause the problems, so I removed the mocking to simplify the tests and fix the timeouts.

mkistler:
Thanks Barrett, I missed that.  Looks like you have it covered.  Nice work!

mkistler (approved):
Looks good! 👍

PR #266, opened 2021-03-23
Error responses

barrett-schonefeld:
Purpose:
- Error responses should be robust and give the user enough information to resolve their problems. The added validation recommends certain fields be included and recommends a schema for those fields.

Changes:
- Add a custom Spectral rule that applies a custom Spectral function to validate 4xx and 5xx responses.
- Add a custom Spectral rule that validates the error response schema is an object, has fields, such as `trace`, `code`, `errors` or `error`, etc. and validates the schema of those fields to ensure the error response follows good practices.
- Defines a deduplication function and passes it to spectral to avoid deduplicating `response-error-response-schema` errors with the same path.
- Add a custom Spectral rule to check for `content` object in request bodies and non-204 responses.

Tests:
- Add tests to ensure warnings are issued for missing fields or invalid/incomplete schemas for each field.
- Add negative tests to ensure no warnings for valid error response.
- Add tests to ensure warning when content object not provided

Docs:
- document `reponse-error-response-schema` rule
- document `content-entry-provided` rule

### DIFF
--- a/src/spectral/rulesets/ibm-oas/error-response-schema.js
+++ b/src/spectral/rulesets/ibm-oas/error-response-schema.js
@@ -66,7 +66,13 @@ function validateErrorModelSchema(errorModelSchema, pathToSchema) {
       errors.push({
         message:
           'Error Model should contain `code` field, a snake-case, string, enum error code',
-        path: [...pathToSchema, 'properties']
+        path: [...pathToSchema, 'properties', 'code']
+      });
+    }
+    if (!hasMessageField(errorModelSchema.properties)) {
+      errors.push({
+        message: 'Error Model should contain a string, `message`, field',
+        path: [...pathToSchema, 'properties', 'message']
       });
     }
   }
@@ -84,10 +90,15 @@ function validStatusCodeField(errorResponseProperties) {
 function hasCodeField(errorModelSchemaProperties) {
   return (
     errorModelSchemaProperties.code &&
-    // type either not defined or type defined as a string
-    (!errorModelSchemaProperties.code.type ||
-      errorModelSchemaProperties.code.type === 'string') &&
+    errorModelSchemaProperties.code.type === 'string' &&
     errorModelSchemaProperties.code.enum
+  );
+}
+
+function hasMessageField(errorModelSchemaProperties) {
+  return (
+    errorModelSchemaProperties.message &&
+    errorModelSchemaProperties.message.type === 'string'
   );
 }
 
--- a/src/spectral/rulesets/ibm-oas/error-response-schema.js
+++ b/src/spectral/rulesets/ibm-oas/error-response-schema.js
@@ -1,9 +1,14 @@
 module.exports = function(errorResponseSchema, _opts, paths) {
   const errors = [];
+  const rootPath = paths.target !== void 0 ? paths.target : paths.given;
   if (!schemaIsObjectWithProperties(errorResponseSchema)) {
-    return [{ message: 'Error response should be an object with properties' }];
+    return [
+      {
+        message: 'Error response should be an object with properties',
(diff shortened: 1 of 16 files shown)


src/spectral/rulesets/ibm-oas/error-response-schema.js
```diff
@@ -0,0 +1,142 @@
...
+    errorModelSchemaProperties.more_info.type === 'string'
+  );
+}
+
+function hasTraceField(errorResponseProperties) {
+  return (
+    errorResponseProperties &&
+    errorResponseProperties.trace &&
+    errorResponseProperties.trace.type === 'string' &&
+    errorResponseProperties.trace.format === 'uuid'
```
mkistler, line 136: Don't require the format to be "uuid".  This is not one of the [standard OpenAPI data types,](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#data-types) and while we allow it in the generator we should not require it here.

src/spectral/rulesets/ibm-oas/error-response-schema.js
```diff
@@ -0,0 +1,142 @@
...
+    !errorResponseProperties.status_code ||
+    errorResponseProperties.status_code.type === 'integer'
+  );
+}
+
+function hasCodeField(errorModelSchemaProperties) {
+  return (
+    errorModelSchemaProperties.code &&
+    errorModelSchemaProperties.code.type === 'string' &&
+    errorModelSchemaProperties.code.enum
```
mkistler, line 113: I don't think we need to require that `code` be an `enum`.

mkistler (approved):
Looks good! 👍 

I left a couple requests for changes but since they are simple fixes I'll approve now.

src/spectral/rulesets/ibm-oas/error-response-schema.js
```diff
@@ -0,0 +1,140 @@
...
+function hasMoreInfoField(errorModelSchemaProperties) {
+  return (
+    errorModelSchemaProperties.more_info &&
+    errorModelSchemaProperties.more_info.type === 'string'
+  );
+}
+
+function hasTraceField(errorResponseProperties) {
+  return (
+    errorResponseProperties &&
```
dpopp07, line 132: (Tiny comment) You don't check for the "definedness" of this in the other `hasWhatever` functions and I don't think it's necessary here

test/spectral/tests/custom-rules/error-response.test.js
```diff
@@ -0,0 +1,492 @@
...
+
+    res = await inCodeValidator(spec, true);
+  });
+
+  it('should not error when content object provided', function() {
+    const expectedWarnings = res.warnings.filter(
+      warn => warn.message === 'Error response should have a content field'
+    );
+    expect(expectedWarnings.length).toBe(0);
+  });
```
dpopp07, line 127: This style is funky but I don't dislike it!

dpopp07 (approved):
Looks good! 👍

Only one tiny comment for consistency

mkistler (approved):
Looks good! 👍

test/spectral/tests/custom-rules/error-response.test.js
barrett-schonefeld: Yes, the goal with this is to isolate tests a bit more to prevent new rules causing cascading test failures.

barrett-schonefeld: The downside for the negative test is that it will cause false positives if the error message is changed, for example. That may be enough reason to take a different approach with the negative tests but still use this technique for the positive tests.

dpopp07 (approved):
Looks good 👍

PR #269, opened 2021-04-06
fix: set the exit code for json output

barrett-schonefeld:
Purpose:
- Want the exit code set properly for json output.

Changes:
- Move the logic to set the exit code so that it applies to json output as well.
- If the warnings limit exceeded, add this as an error to the results so that it may be naturally included in the json output.

Tests:
- repair warnings limit tests and exit code tests

### DIFF
--- a/src/cli-validator/runValidator.js
+++ b/src/cli-validator/runValidator.js
@@ -288,6 +288,32 @@ const processInput = async function(program) {
       results.hint = false;
     }
 
+    // fail on errors or if number of warnings exceeds warnings limit
+    if (results.error) {
+      exitCode = 1;
+    } else {
+      // Calculate number of warnings and set exit code to 1 if warning limit exceeded
+      let numWarnings = 0;
+      for (const key of Object.keys(results.warnings)) {
+        numWarnings += results.warnings[key].length;
+      }
+      if (numWarnings > limitsObject.warnings) {
+        exitCode = 1;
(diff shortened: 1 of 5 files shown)


src/cli-validator/runValidator.js
```diff
@@ -288,6 +288,33 @@ const processInput = async function(program) {
...
+        if (!results.errors) {
+          results.errors = {};
+        }
+        results.errors['warnings-limit'] = [
+          {
+            path: [],
+            message: `Number of warnings (${numWarnings}) exceeds warnings limit (${
+              limitsObject.warnings
+            }).`,
+            rule: 'warnings_limit'
```
dpopp07, line 312: I may be overthinking this but I wonder if this isn't the most clear. We don't actually have a rule anywhere called "warnings_limit" so I don't want to mislead users into thinking that they should be setting that property somewhere. I suppose that if they see this message, they already know how to configure it, and I don't have a better option. Perhaps this error doesn't need a "rule" though.

dpopp07 (approved):
I left one comment, but looks good! I like the idea of including the warning limit message in the list of errors instead of a console log 👍

src/cli-validator/runValidator.js
barrett-schonefeld: I think you're right. This would be better suited as "builtin"

PR #272, opened 2021-04-07
feat: show path to component that caused error in verbose mode

barrett-schonefeld:
Purpose:
- Errors may be scattered throughout a resolved schema, so we want to make it easy for users to identify the source component from which the error originates.

Changes:
- Walk the unresolved spec to see if the error occurs in a ref. Identify the source location and include the path to the component if the validator is in verbose mode.
- Update the `printRuleNames` variable to be `verbose` to be more general.

Tests:
- Add tests to ensure the path to the components added correctly.

### DIFF
--- a/src/cli-validator/utils/addPathsToComponents.js
+++ b/src/cli-validator/utils/addPathsToComponents.js
@@ -1,0 +1,100 @@
+const { each } = require('lodash');
+const getPathAsArray = require('./getPathAsArray');
+
+// takes validator results dictionary in the format:
+//   {
+//     errors: {
+//       validator-name: [
+//         {
+//           path: [path, to, error],
+//           message: 'error message',
+//           rule: rule_name
+//         },
+//       ],
+//     },
+//     warnings: ...
+//   }
+//
+// adds path to the originating component if it exists
+// modifies the given results object
+module.exports = function(originalResults, unresolvedSpec) {
+  each(originalResults, function(validatorsDict) {
+    each(validatorsDict, function(errors) {
+      pointToComponents(errors, unresolvedSpec);
+    });
+  });
+};
+
+// takes an array of validator errors:
+//   [
+//     {
+//       path: [path, to, error],
+//       message: 'error message',
+//       rule: rule_name
+//     },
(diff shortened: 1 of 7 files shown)


src/cli-validator/utils/addPathsToComponents.js
```diff
@@ -0,0 +1,100 @@
...
+//           message: 'error message',
+//           rule: rule_name
+//         },
+//       ],
+//     },
+//     warnings: ...
+//   }
+//
+// adds path to the originating component if it exists
+// modifies the given results object
```
mkistler, line 19: Nice comment!

src/cli-validator/utils/getPathAsArray.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function(path) {
+  return Array.isArray(path) ? path : path.split('.');
```
mkistler, line 2: Nice!

mkistler (approved):
Looks great! 👍 

I tried this out on a real API doc and here's a snippet of the results:
```
  Message :   Error response should have a uuid `trace` field
  Rule    :   response-error-response-schema
  Path    :   paths./v1/status/health_check.get.responses.500.content.application/json.schema.properties
  Line    :   4325
  Component Path    :   components.schemas.serviceError.properties
  Component Line    :   4711

  Message :   Error Model should contain `more_info` field that contains a URL with more info about the error
  Rule    :   response-error-response-schema
  Path    :   paths./v1/status/health_check.get.responses.500.content.application/json.schema.properties
  Line    :   4325
  Component Path    :   components.schemas.serviceError.properties
  Component Line    :   4711

  Message :   Error Model should contain `code` field, a snake-case, string error code
  Rule    :   response-error-response-schema
  Path    :   paths./v1/status/health_check.get.responses.500.content.application/json.schema.properties.code
  Line    :   4325
  Component Path    :   components.schemas.serviceError.properties.code
  Component Line    :   4712
```

Very nice.

dpopp07 (approved):
This looks great!! 🎉

PR #273, opened 2021-04-07
fix: soften the response-example rule to look only at success responses

barrett-schonefeld:
Purpose:
- Do not necessarily expect response examples for non-2xx and default responses. Expected for 2xx responses.

Changes:
- Soften the check to look only at success responses.

Tests:
- Update tests to ensure errors only issued for success responses without examples.


dpopp07 (approved):
Looks good! 👍

mkistler (approved):
Looks good! 👍

Issue #270, opened 2021-04-07
Unexpected token u in JSON at position 0


rolivares:
this error is shown on console when I upgrade to v0.40 and execute the linter, is not currently appearing on 0.39

maybe related about 'response-error-response-schema' that appearing on revision too, using the same definition

`SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at Object.json-schema-validation (eval at exports.evaluateExport (C:\GIT\MTT\bono\back\node_modules\@stoplight\spectral\dist\rulesets\evaluators.js:89:80), <anonymous>:12:27)    
    at Object.exports.lintNode (C:\GIT\MTT\bono\back\node_modules\@stoplight\spectral\dist\runner\lintNode.js:30:33)
    at callback (C:\GIT\MTT\bono\back\node_modules\@stoplight\spectral\dist\runner\runner.js:38:32)
    at JSONPath._handleCallback (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:626:7)
    at JSONPath._trace (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:658:12)
    at JSONPath._trace (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:733:19)
    at JSONPath.evaluate (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:566:23)
    at new JSONPath (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:498:22)
    at Object.JSONPath (C:\GIT\MTT\bono\back\node_modules\jsonpath-plus\dist\index-umd.js:452:16)`

alecor191:
I'm getting the same error using the following command:

```
lint-openapi -v -j -p --debug 'D:\temp\my-openapi-spec.json'
```

Output:

```
[Warning] Skipping non-existent file: D:\temp\my-openapi-spec.json
SyntaxError: Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at Object.json-schema-validation (eval at exports.evaluateExport (C:\Users\aleco\AppData\Roaming\npm\node_modules\ibm-openapi-validator\node_modules\@stoplight\spectral\dist\rulesets\evaluators.js:89:80), <anonymous>:12:27)
[...]

PR #276, opened 2021-04-12
fix: error parsing JSONSchema in json-schema-validation function

barrett-schonefeld:
Some users, as outlined in #270, are getting an error parsing the `x-sdk-operations` JSON Schema in the `json-schema-validation` function.

Instead of parsing the JSON in the `json-schema-validation`, it may resolve the problem to parse the JSON file up front, and load it into `STATIC_ASSETS` as an object, which in hindsight is the more sensible approach.

This simple change passes in our testing environment. However, the original implementation also works in our testing environment. Without replicating the users' environments who are experiencing this issue, I'm not sure this resolves the problem. I am, however, certain that parsing the JSON Schema is the root of the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: node_js
+os:
+- windows
+- linux
 node_js:
-- 10
 - 12
+- 14
 cache:
   npm: false
 script:


mkistler:
It looks like both the users that encountered the problem in #270 are running on Windows.  I think we need to test your fix on Windows to see if it works there.

barrett-schonefeld:
Closing this PR until we move forward with more robust support for Windows compatibility.

rubenramosDev:
Any updates on this?

Issue #275, opened 2021-04-11
Unable to customize spectral config: Error: Cannot find module 'ajv'


alecor191:
Disclaimer: I'm new to openapi-validator so it's very likely that I'm missing something obvious.

**What I'm trying to do**
Customize default rules and disable rule `oas3-api-servers`. Essentially I'm trying to follow instructions in `README.md`:

> To completely disable a rule, use the severity of off. For example, to disable the operation-tags rule, specify the following in your Spectral ruleset file:
> ```
> extends: ibm:oas
> rules:
>   operation-tags: off
> ```

**What I have done**
I added a file `.spectral.yaml` with the following content:

```yaml
extends: ibm:oas
rules:
  oas3-api-servers: off
```

and then I'm running the Docker container of openapi-validator:

```shell
docker run --volume "<my-local-path>":/data jamescooke/openapi-validator -v -s -c /data/lint-openapi-settings.json /data/my-openapi-spec-file.json
```

Running the above, results in the following output:

```
Error: Cannot find module 'ajv'
Require stack:
- /usr/local/lib/node_modules/ibm-openapi-validator/node_modules/@stoplight/spectral/dist/rulesets/evaluators.js
- /usr/local/lib/node_modules/ibm-openapi-validator/node_modules/@stoplight/spectral/dist/spectral.js
- /usr/local/lib/node_modules/ibm-openapi-validator/node_modules/@stoplight/spectral/dist/index.js
- /usr/local/lib/node_modules/ibm-openapi-validator/src/spectral/utils/spectral-validator.js
- /usr/local/lib/node_modules/ibm-openapi-validator/src/cli-validator/utils/validator.js
- /usr/local/lib/node_modules/ibm-openapi-validator/src/cli-validator/runValidator.js
- /usr/local/lib/node_modules/ibm-openapi-validator/src/cli-validator/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:893:15)
    at Function.resolve (internal/modules/cjs/helpers.js:94:19)
    at require (/usr/local/lib/node_modules/ibm-openapi-validator/node_modules/@stoplight/spectral/dist/rulesets/evaluators.js:41:25)
    at Object.eval (eval at exports.evaluateExport (/usr/local/lib/node_modules/ibm-openapi-validator/node_modules/@stoplight/spectral/dist/rulesets/evaluators.js:89:80), <anonymous>:3:13)
[...]

PR #277, opened 2021-04-12
fix: fix imports in json-schema-validation function

barrett-schonefeld:
Purpose:
- When users extend the `ibm:oas` ruleset with their own Spectral config file, the original `ajv` and `STATIC_ASSETS` imports break.

Changes:
- If the standard attempt to import is broken, use relative import based on the structure of the Spectral package.

Needs:
- Maybe a better approach. This may be volatile to changes in the way Spectral organizes its distributions.
- If the approach is acceptable, need a test that tracks the case when the user extends the `ibm:oas` ruleset where the current working directory is outside the scope of our project.

Alternative approaches:
- Remove the `ibm-sdk-operations` rule for now and reintroduce it later. [Branch that implements this approach](https://github.com/IBM/openapi-validator/tree/fix-extends-problem).

Related Issue: #275


barrett-schonefeld:
Closing in favor of #279 for now.

PR #284, opened 2021-04-19
fix: when `limit` and `offset` properties included in `allOf`

barrett-schonefeld:
Purpose:
- When validating pagination, validator produces a false positive when the limit and offset properties are included in an allOf combined schema, and this fix resolves the bug.

Changes:
- Merges `allOf` schemas and top-level schema. Allows validation to be done as if one schema provided.

Tests:
- Ensure no errors occur when limit and offset are provided in allOf schema.

### DIFF
--- a/src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
+++ b/src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
@@ -17,6 +17,7 @@
 // - The response body must contain an array property with the same plural resource name appearing in the collection’s URL.
 
 const MessageCarrier = require('../../../utils/messageCarrier');
+const mergeAllOfSchemaProperties = require('../../../utils/mergeAllOfSchemaProperties');
 
 module.exports.validate = function({ resolvedSpec }, config) {
   const messages = new MessageCarrier();
@@ -55,9 +56,11 @@ module.exports.validate = function({ resolvedSpec }, config) {
       continue;
     }
 
+    const jsonResponseSchema = mergeAllOfSchemaProperties(jsonResponse.schema);
+
     // If no array at top level of response, skip this path
     if (
-      !Object.values(jsonResponse.schema.properties).some(
+      !Object.values(jsonResponseSchema.properties).some(
         prop => prop.type === 'array'
       )
     ) {
@@ -154,7 +157,7 @@ module.exports.validate = function({ resolvedSpec }, config) {
       'properties'
     ];
 
-    const limitProp = jsonResponse.schema.properties.limit;
+    const limitProp = jsonResponseSchema.properties.limit;
     if (!limitProp) {
       messages.addMessage(
         propertiesPath,
@@ -164,8 +167,8 @@ module.exports.validate = function({ resolvedSpec }, config) {
       );
     } else if (
       limitProp.type !== 'integer' ||
-      !jsonResponse.schema.required ||
-      jsonResponse.schema.required.indexOf('limit') === -1
+      !jsonResponseSchema.required ||
+      jsonResponseSchema.required.indexOf('limit') === -1
     ) {
       messages.addMessage(
         [...propertiesPath, 'limit'],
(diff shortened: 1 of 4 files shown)


src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -216,3 +219,16 @@ module.exports.validate = function({ resolvedSpec }, config) {
 
   return messages;
 };
+
+function mergeAllOfSchemaProperties(schema) {
+  const mergedSchema = {};
+  const allOfArr = schema['allOf'];
+  if (allOfArr && Array.isArray(allOfArr)) {
+    // delete now, so it not merged back into the final result
```
dpopp07, line 227:
```suggestion
    // delete now, so it is not merged back into the final result
```

Just a small typo

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -216,3 +219,16 @@ module.exports.validate = function({ resolvedSpec }, config) {
...
+    delete schema['allOf'];
+    allOfArr.forEach(function(allOfSchema) {
+      merge(mergedSchema, allOfSchema);
```
dpopp07, line 230: How does this handle property conflicts? For example, each `allOfSchema` will likely have a field called `properties` - will they be overwritten or are subfields merged as well? It may be a good idea to add one more schema to the `allOf` list in your test just to verify this

dpopp07 (commented):
Just leaving a couple comments before approving but looks good for the most part

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
barrett-schonefeld: The `lodash` merge function does do a deep merge.

I will add a test to ensure it is actually deep merge and to ensure we preserve the deep merge behavior in future updates.

barrett-schonefeld: Although `lodash` does do a deep merge, I may need to consider some other possible consequences here.

barrett-schonefeld: I did two things to follow up on this:

1. I made the merging function a utility function and wrote unit tests separately, which is a much better approach than adding it to the module and only having acceptance tests.
2. I added another schema to the allOf array to ensure the properties are not being overwritten. The `pagination-ibm` module uses the `required` and `properties` fields from the schema, and although it's unclear how to merge things like `description` and `summary`, the tests ensure `required` and `properties` are merged properly.

src/plugins/validation/oas3/semantic-validators/pagination-ibm.js
```diff
@@ -55,9 +56,11 @@ module.exports.validate = function({ resolvedSpec }, config) {
       continue;
     }
 
+    const jsonResponsSchema = mergeAllOfSchemaProperties(jsonResponse.schema);
```
mkistler, line 59: Not sure if it was intentional to drop the "e", but to me it looks like a typo.
```suggestion
    const jsonResponseSchema = mergeAllOfSchemaProperties(jsonResponse.schema);
```

If you accept this change you'll need to fix the other instances of `jsonResponsSchema` accordingly.

test/plugins/validation/oas3/pagination-ibm.test.js
```diff
@@ -147,6 +147,95 @@ describe('validation plugin - semantic - pagaination - oas3', function() {
...
+                              type: 'integer',
+                              format: 'int32'
+                            },
+                            prop2: {
+                              description: 'prop2',
+                              type: 'integer',
+                              format: 'int32'
+                            }
+                          }
+                        }
```
mkistler, line 222: It doesn't seem like this second oneOf element is needed to verify the behavior this test is checking.  I'd recommend removing it.

mkistler (approved):
Looks good. 👍 

I left a few comments with suggested changes, but I'll leave it up to you if want to make these.

dpopp07 (approved):
Looks good! 👍

PR #285, opened 2021-04-20
feat: warn for invalid content types

barrett-schonefeld:
Purpose:
- Identify cases where the content type should be more specific or should likely just be application/json to ensure API and API definition are following good design principles.

Changes:
- Add Spectral rule to warn on any occurrence of `*/*` content-type
- Add Spectral rule to warn when error response content type does not support `application/json` because the error response is highly likely just be `application/json`

Tests:
- Add test specs for each that include both good practices and violations and ensure the proper number of errors occur

Docs:
- Document each rule

### DIFF
--- a/test/spectral/tests/custom-rules/ibm-content-type-is-specific.test.js
+++ b/test/spectral/tests/custom-rules/ibm-content-type-is-specific.test.js
@@ -1,0 +1,81 @@
+const inCodeValidator = require('../../../../src/lib');
+
+describe('spectral - test content type is not */*', function() {
+  it('should error only when the content type is */*', async () => {
+    const spec = {
+      openapi: '3.0.0',
+      paths: {
+        path1: {
+          post: {
+            requestBody: {
+              content: {
+                // error 1
+                '*/*': {
+                  $ref: '#/components/schemas/GenericSchema'
+                }
+              }
+            },
+            parameters: [
+              {
+                name: 'param1',
+                in: 'query',
+                content: {
+                  // error 2
+                  '*/*': {
+                    $ref: '#/components/schemas/GenericSchema'
+                  }
+                }
+              }
+            ],
+            responses: {
+              '200': {
+                content: {
+                  // error 3
+                  '*/*': {
+                    $ref: '#/components/schemas/GenericSchema'
+                  }
+                }
+              },
+              '201': {
+                content: {
+                  // no error
+                  'application/json': {
+                    $ref: '#/components/schemas/GenericSchema'
+                  }
+                }
+              },
+              '202': {
+                content: {
+                  // no error
+                  'image/*': {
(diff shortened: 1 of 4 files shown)


dpopp07:
I think these are good checks but I feel slightly uneasy about this:
> This rule should be ignored when the API actually returns an error response that is not `application/json`.

I hate to put the user in a position where their API definition is correct and they can't get a clean validation because one a warning that they just have to ignore because it doesn't apply to them. I know that for these checks, this scenario is extremely unlikely so maybe it's fine. It just doesn't feel very clean to me. I can't think of a better approach though, so maybe this is fine. I know the user can disable warnings and maybe the 1% of users this will be a problem for are diligent enough not to need this check elsewhere 🙂 Just some thoughts

barrett-schonefeld:
> I think these are good checks but I feel slightly uneasy about this:
> 
> > This rule should be ignored when the API actually returns an error response that is not `application/json`.
> 
> I hate to put the user in a position where their API definition is correct and they can't get a clean validation because one a warning that they just have to ignore because it doesn't apply to them. I know that for these checks, this scenario is extremely unlikely so maybe it's fine. It just doesn't feel very clean to me. I can't think of a better approach though, so maybe this is fine. I know the user can disable warnings and maybe the 1% of users this will be a problem for are diligent enough not to need this check elsewhere 🙂
> 

I see what you're saying, and I considered this as well.

A couple of things to consider:
1. The statement is true because the rule should be ignored if you actually support all content types, so I thought it'd be good to mention the caveat. Should we leave the comment in, or does it come across as confusing?
2. Because the check assumes that these content types are almost always misused in these scenarios, we could consider making the check an `info` or `hint`. @mkistler may have a better idea for what the severity should be based on how often these patterns are misused in API reviews.

test/spectral/tests/custom-rules/ibm-content-type-is-specific.test.js
```diff
@@ -0,0 +1,73 @@
...
+                  'application/json': {
+                    $ref: '#/components/schemas/GenericSchema'
+                  }
+                }
+              }
+            }
+          }
+        }
+      },
+      components: {
```
mkistler, line 51: Please a "no error" case for a content type like `image/*` or `text/*` -- just to make clear that these are acceptable.

mkistler (approved):
Looks good. 👍 

I requested one additional test, but that should be straightforward so I'll approve now to avoid re-review.

dpopp07 (approved):
If Mike is good with it, it's good with me 👍 

> The statement is true because the rule should be ignored if you actually support all content types, so I thought it'd be good to mention the caveat. Should we leave the comment in, or does it come across as confusing?

No, the comment is good and not confusing (sorry I hadn't responded to that). I was uneasy about the actual implications of the comment, not the comment itself. It's good as is!

PR #288, opened 2021-04-29
feat: error for non-object request bodies

barrett-schonefeld:


test/spectral/tests/custom-rules/request-body-object.test.js
```diff
@@ -0,0 +1,55 @@
+const inCodeValidator = require('../../../../src/lib');
+
+describe('spectral - test validation that schema provided in content object', function() {
+  it('should error only when parameter does not provide a schema or content', async () => {
```
dpopp07, line 4: Test description looks like it may be a copy/paste error 🙂

dpopp07 (approved):
Aside from the test description, this looks good! I think we may have some existing checks around this (specifically, looking for top-level arrays) and we don't want to duplicate things - have those been removed already?

barrett-schonefeld:
> I think we may have some existing checks around this (specifically, looking for top-level arrays) and we don't want to duplicate things - have those been removed already?

We have the `no_array_responses` rule, so this will do a similar check for request bodies.

mkistler (changes requested):
This looks good but I think we need to add `request-body-object` to our documentation in `docs/spectral-rules.md`.

barrett-schonefeld:
> we need to add `request-body-object` to our documentation in `docs/spectral-rules.md`.

Good catch, and I added documentation for the rule.

dpopp07:
> We have the no_array_responses rule, so this will do a similar check for request bodies.

Aha, I was conflating requests and responses in my head! Sounds good 👍

test/spectral/tests/custom-rules/request-body-object.test.js
```diff
@@ -0,0 +1,55 @@
...
+                      type: 'string'
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    };
```
mkistler, line 45: Please add one more test -- where the request body schema has `properties` but _not_ `type: object`.

I think (I _hope_) that this will not trigger a message, but we should make sure.

mkistler (changes requested):
This looks good but I realized that there is one more test we should include.

test/spectral/tests/custom-rules/request-body-object.test.js
barrett-schonefeld: The core Spectral `enumeration` function does not catch the case where the `type` field is omitted.

I will create a custom function that catches this.

barrett-schonefeld: Fixed

mkistler: I think you misunderstood my comment.  I don't think we should require `type: object` -- it is generally assumed if there are properties. And perhaps we could check for `properties` if there is no `type` field, but I'd rather just keep this simple.  If `type` is present, make sure it is `object`, and and if not issue a message.  if `type` is not present, then don't issue any message.

mkistler (changes requested):
I think it is too strict to require `type: object`.

test/spectral/tests/custom-rules/request-body-object.test.js
barrett-schonefeld: I did misunderstand the comment. I returned the rule to its original state and added a test to ensure that no error is issued when `type` is omitted.

mkistler (approved):
Looks good! 👍

PR #290, opened 2021-04-29
feat: validate `x-sdk-operations` extension

barrett-schonefeld:
**Description:** 
- This PR uses a provided Spectral function to do JSON Schema validation. This is a positive because it is a well-defined way to invoke the AJV validator.
- The one drawback is that we need to dynamically convert the relative path to an absolute path before loading the default config into STATIC_ASSETS because we need to ensure Spectral can always find the schema regardless of the current working directory.
- The Spectral `schema` function has the same problem we observed (in Spectral `5.8.1` and below) in the `oas3-schema` rule where only one error is returned at a time. We will want to follow up on this implementation by raising this issue to Spectral and fixing the `schema` function.

Purpose:
- Validate that the x-sdk-operations extension is used correctly.

Changes:
- Load the absolute path to the JSON Schema into static assets, so when the `ibm:oas` ruleset is extended, still able to find the JSON Schema
- Replace relative references to `./schema` in `.defaultsforspectral.yaml` with their absolute path using regex replace

Tests:
- Ensure `ibm-sdk-operations` rule catches errors

Docs:
- Document the `ibm-sdk-operations` rule

### DIFF
--- a/src/spectral/utils/spectral-validator.js
+++ b/src/spectral/utils/spectral-validator.js
@@ -70,11 +70,7 @@ const setup = async function(spectral, rulesetFileOverride, configObject) {
 
   // Add IBM default ruleset to static assets to allow extends to reference it
   const staticAssets = require('@stoplight/spectral/rulesets/assets/assets.json');
-  setupStaticAssets(
-    staticAssets,
-    defaultSpectralRulesetURI,
-    spectralRulesetURI
-  );
+  setupStaticAssets(staticAssets, defaultSpectralRulesetURI);
   Spectral.registerStaticAssets(staticAssets);
 
   // Register formats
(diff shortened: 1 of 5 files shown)


src/spectral/utils/spectral-validator.js
```diff
@@ -118,6 +115,14 @@ function setupStaticAssets(staticAssets, defaultSpectralRulesetURI) {
   });
 }
 
+function addDataToSpectralConfig(content) {
+  const config = yaml.safeLoad(content);
+  config['rules']['ibm-sdk-operations']['then']['functionOptions']['schema'] = {
+    $ref: path.join(__dirname, '../rulesets/schemas/x-sdk-operations.json')
+  };
+  return yaml.safeDump(config);
```
mkistler, line 123: Could you do this substitution without converting from/to yaml?  It seems like you could just regex replace `'../rulesets`` with `_dirname + '../rulesets'` and get the same result.  What's more, if we have any new instances where we have to reference resources in this directory, we won't need additional code here.

mkistler (changes requested):
I think the doc changes are a bit messed up.

And I had an idea about using regex replace to update the paths.

src/spectral/utils/spectral-validator.js
barrett-schonefeld: That's a great idea. I will implement it that way.

mkistler (approved):
Looks good! 👍

dpopp07 (approved):
Looks good to me

PR #291, opened 2021-04-30
chore: ignore node_modules in codescan

christiancompton:
This PR excludes the `node_modules` directory from our routine code scans after the most recent runs detected problems in a dependency.


barrett-schonefeld (approved):
Looks good!

PR #292, opened 2021-05-03
feat: warn when parameters can be defined as global param

barrett-schonefeld:
Purpose:
- tooling uses the `x-sdk-global-param` field. Ensure parameters are marked properly when the parameter is eligible to use this extension.

Changes:
- Add a custom function that looks for global params and warns if they are not marked properly.
- Add a custom rule that utilizes this function

Test:
- Add a test that checks both that the validator issues a warning only in the correct cases.

Docs:
- Document the `ibm-global-params` rule

### DIFF
--- a/docs/spectral-rules.md
+++ b/docs/spectral-rules.md
@@ -107,11 +107,59 @@ responses:
 
 **Default Severity**: warn
 
+## ibm-global-params
+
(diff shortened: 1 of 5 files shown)


barrett-schonefeld:
This is not ready because, when user provides custom Spectral config, the following happens when trying to require `lodash` functions:

Screen Shot 2021-05-03 at 11 47 17 AM

PR #299, opened 2021-06-07
Update Readme.md to include "request-body-object" spectral rule.

olafur-palsson:
By default the validator doesn't accept array request bodies but that fact is not documented. This will likely help devs navigate to the spectral rule config faster. :)


barrett-schonefeld (approved):
Good catch! Looks good!

PR #306, opened 2021-06-15
fix: allow null values inside enum property.

rtician:
As described in the [docs](https://swagger.io/docs/specification/data-models/enums/) `null` values are allowed inside enums. This PR fixed the lint error when you have nullable enums.


barrett-schonefeld (approved):
The work looks good!

I updated the PR title to follow Angular commit style (i.e. `fix: ...`).

rtician:
Thanks. Any chance to release a new version with this fix?

PR #313, opened 2021-07-24
Make it easier to disable `operation_id_case_convention`

### LINKED ISSUE #312: `operation_id_case_convention` rule can't be easily disabled
With every other case convention rule, it can be disabled like this:

```js
{
	param_name_case_convention: ["off"]
}
```

If the first item in the array is `"off"`, then the linter will skip the rule.

`operation_id_case_convention` is the only rule that does not follow this. Despite providing `"off"` as the first item, it still tries to check the case which isn't provided:

https://github.com/IBM/openapi-validator/blob/2bf13e19fd3893190caa0725671197d92a086464/src/plugins/validation/2and3/semantic-validators/operations-shared.js#L112-L114

The code should only run this check if `checkStatus !== "off"` like every other rule.

For now, a temporary workaround is to specify a case anyways which will disable the check:

```js
{
	operation_id_case_convention: ["off", "lower_snake_case"]
}
```

Richienb:
Fixes #312

### DIFF
--- a/src/plugins/validation/2and3/semantic-validators/operations-shared.js
+++ b/src/plugins/validation/2and3/semantic-validators/operations-shared.js
@@ -110,15 +110,17 @@ module.exports.validate = function({ jsSpec, resolvedSpec, isOAS3 }, config) {
       } else {
         // check operationId for case convention
         const checkStatus = config.operation_id_case_convention[0];
-        const caseConvention = config.operation_id_case_convention[1];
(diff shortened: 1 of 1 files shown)


Richienb:
// @dpopp07

barrett-schonefeld (approved):
Looks good!

dpopp07 (approved):
Looks great, thanks for the PR!

Issue #311, opened 2021-07-22
Validator not recognizing missing property definition in object schemas


he-ritter:
The following invalid schema is not being recognized by the validator:
```
InfoRecord:
  type: object
  required:
    - userId
    - name
  properties:
    userId:
      type: string
      description: The user id.
```

barrett-schonefeld:
Yes, this looks like a good check to add to the Validator. Would be happy to accept a PR for this, and also would be happy to look into this in the couple of weeks.

PR #314, opened 2021-08-04
feat: check for missing required properties

barrett-schonefeld:
Look for required properties that are not defined with the schema.

The`undefined_required_properties` rule only looked in the `components` section, which is why it is deprecated.

The new rule enforces that required properties are defined in a given requestBody, response, and parameter schema and in any subschemas.

Related issue: #311


src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -158,14 +159,37 @@ rules:
...
+      function: check-major-version
+  # ensure required properties are in the schema
+  missing-required-property:
+    description: 'A required property is not in the schema'
+    message: "{{error}}"
+    formats: ["oas2", "oas3"]
+    given:
+    - $.paths[*][*][parameters][*].schema
+    - $.paths[*][*][parameters,responses][*].content[*].schema
+    - $.paths[*][*][requestBody].content[*].schema
```
dpopp07, line 179: Is this operating on the resolved API definition? I see you're not checking inside of the `components` section, is that because we would expect to see all schemas where they are actually used?

src/.defaultsForValidator.js
```diff
@@ -136,7 +135,8 @@ const defaults = {
 */
 const deprecated = {
   'no_produces_for_get': 'no_produces',
-  'parameters.snake_case_only': 'param_name_case_convention'
+  'parameters.snake_case_only': 'param_name_case_convention',
+  'undefined_required_properties': 'missing-required-property',
```
dpopp07, line 139: I forget how this value gets used (I think in a warning message) but it may be good to call out the fact that it's a spectral rule

dpopp07 (commented):
Two quick questions/comments to look at before approving - code changes look good for the most part though!

src/spectral/rulesets/.defaultsForSpectral.yaml
barrett-schonefeld: Yes, that's correct!

src/.defaultsForValidator.js
barrett-schonefeld: Good point! The key is the old rule, and the value is the replacement rule if it exists.

I could do something simple like use `missing-required-property (Spectral rule)`, which will result in, "Use `missing-required-property (Spectral rule)` instead."

That might be confusing, though, so I could update the deprecation code to produce a different message when the new rule is a Spectral rule.

dpopp07: I think that's fine, just something to indicate that the rule isn't going to be found in the validaterc file like the old one

barrett-schonefeld: added the note in parentheses

dpopp07 (approved):
Looks good! 👍

PR #315, opened 2021-08-04
feat: string schema should define enum or pattern, minLength, maxLength

barrett-schonefeld:
Purpose: encourage users to define schemas with explicit boundaries for valid string input.

Enforces use of `pattern`, `minLength`, and `maxLength` on `parameter` and `requestBody` string schemas.

Exceptions:
- `enum` defined

### DIFF
--- a/src/spectral/rulesets/ibm-oas/string-boundary.js
+++ b/src/spectral/rulesets/ibm-oas/string-boundary.js
@@ -44,7 +44,7 @@ function traverseSchema(schema, path) {
 function stringBoundaryErrors(stringSchema, path) {
   const errors = [];
   if (!stringSchema.enum) {
-    if (!stringSchema.pattern) {
+    if (!stringSchema.pattern && !['binary', 'date', 'date-time'].includes(stringSchema.format)) {
       errors.push({
         message: 'Should define a pattern for a valid string',
         path
--- a/src/spectral/rulesets/ibm-oas/string-boundary.js
+++ b/src/spectral/rulesets/ibm-oas/string-boundary.js
@@ -27,7 +27,6 @@ function traverseSchema(schema, path) {
     const composedSchemas = schema[whichComposedSchemaType];
     if (Array.isArray(composedSchemas)) {
       composedSchemas.forEach(function(composedSchema, index) {
-        // TODO: is this the right way to get the index?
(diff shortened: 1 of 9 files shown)


src/spectral/rulesets/ibm-oas/string-boundary.js
```diff
@@ -0,0 +1,70 @@
...
+  } else if (schema.allOf || schema.anyOf || schema.oneOf) {
+    const whichComposedSchemaType = schema.allOf
+      ? 'allOf'
+      : schema.anyOf
+      ? 'anyOf'
+      : 'oneOf';
+    const composedSchemas = schema[whichComposedSchemaType];
+    if (Array.isArray(composedSchemas)) {
+      composedSchemas.forEach(function(composedSchema, index) {
+        // TODO: is this the right way to get the index?
```
dpopp07, line 30: Sure, that definitely works. There are other ways but they are only different, not necessarily better

dpopp07 (approved):
Changes look great! 👍 

This seems like it could be a heavy-handed change though, I don't know if that many API definitions define constraints for _all_ of their strings and this could result in a lot of new messages. It's a warning, so I'm fine with it being there, I just think I missed the original reasoning for the PR. Did this stem from a problem at some point?

barrett-schonefeld:
> This seems like it could be a heavy-handed change though, I don't know if that many API definitions define constraints for _all_ of their strings and this could result in a lot of new messages. It's a warning, so I'm fine with it being there, I just think I missed the original reasoning for the PR. Did this stem from a problem at some point?

Yes, it will have a big impact. (Not negating the scope of the impact, but this update looks for string boundaries in request body properties and parameters, and responses are not checked.)

The motivation is the [IBM's new stance on the Robustness Principle](https://cloud.ibm.com/docs/api-handbook?topic=api-handbook-robustness). 

The old stance is that APIs should be lenient in what they accept and ignore invalid input.

The new stance is that APIs should not be lenient in what they accept and instead should clearly define the boundaries for valid input (and reject input outside of those clear boundaries).

dpopp07:
Great reasoning! Thanks for detailing that. Good to ship it whenever the PR is ready 👍

hudlow (approved):

PR #320, opened 2021-09-13
fix: 101 and 202 responses do not need content entries

### LINKED ISSUE #298: "non-204 responses should define a content object" seems too narrow
Our API has two cases which seem like legitimate exceptions to the "non-204 responses should define a content object" rule:

- Methods that transition to a WebSocket via a `101 Switching Protocols` response.  Per https://datatracker.ietf.org/doc/html/rfc7231#section-6.2.2 there's no applicable content type in this case.
- Methods that use `202 Accepted`.  While https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.3 advises a status monitor, we do not have one at present (note that we generally use `202` for `DELETE` methods, and thus a subsequent `GET` reveals whether the operation has completed).

In general, this rule is helpful, and did catch some legitimate mistakes in the API specification.  As such, I'd prefer not to simply disable it.

barrett-schonefeld:
Fixes #298.

Do not warn for missing `content` field for `101` and `202` responses because these protocols do not require response bodies.

### DIFF
--- a/src/spectral/rulesets/.defaultsForSpectral.yaml
+++ b/src/spectral/rulesets/.defaultsForSpectral.yaml
@@ -105,7 +105,7 @@ rules:
   content-entry-provided:
     description: Request bodies and non-204 responses should define a content object
     given:
-    - $.paths[*][*].responses[?(@property != '204')]
+    - $.paths[*][*].responses[?(@property != '204' && @property != '202' && @property != '101')]
     - $.paths[*][*].requestBody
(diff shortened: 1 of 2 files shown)


dpopp07 (approved):
Looks good to me 👍

dpopp07:
@meem FYI

meem:
@dpopp07 @barrett-schonefeld Excellent, thanks!

Issue #301, opened 2021-06-08
Incorrect incorrect_ref_pattern rule?


Ilyes512:
So I have a pretty large openapi yaml file that I split up into multiple files for organisational purposes. Currently I am getting allot of warnings that I previously with an older version (version 0.38 to 0.46) did not get. As far as I understand it the warning or not correct. For instance I get:

```
  Message :   schema $refs must follow this format: *#/components/schemas*
  Rule    :   incorrect_ref_pattern
  Path    :   paths./foobar/acme.get.responses.200.content.application/json.schema.allOf.0.$ref
  Line    :   56
```
The ref it looks like this:

```
- $ref: "schemas/foobar.yml#/components/schemas/FoobarResourcesDocument"
```

Is this a false positive or am I wrong? Or maybe its not even supported? (OpenApi doc says it is: https://swagger.io/docs/specification/using-ref/).

dpopp07:
Hey @Ilyes512, thanks for the issue. This does look like a false positive to me. I'm guessing that check is getting confused by the multi-file setup. Though, this isn't a new check, nor have we seen this before. I'll try to look into it when I get a chance.

barrett-schonefeld:
Yes, I think the problem here is that your document is likely being treated as a Swagger 2 document rather than the appropriate OAS3 document.

The way to resolve this would be to be more sophisticated in the validator about how we detect the `OpenAPI` version or to explicitly include the `OpenAPI` version in the document.

Ilyes512:
The root file and every file that I reference always starts with:

```yaml
---
openapi: 3.0.3
# ... the rest
```

Latest version is `3.1.0`. Not sure if updating the version to the latest will help?

barrett-schonefeld:
Looking into this now.

I created a test case with the offending `$ref`, and I am not able to reproduce the error.

If you could provide the context in which the `$ref` is used, I may be able to replicate the behavior more closely in the test case to see if I get the error.

barrett-schonefeld:
I opened PR #321, which provides a test case meant to mimic the error outlined in the initial issue.

Krapaince:
Hey @barrett-schonefeld, I ran into the same issue with  a multiple files openapi documentation.
```
  Message :   schema $refs must follow this format: *#/components/schemas*
  Rule    :   incorrect_ref_pattern
  Path    :   paths./user/login.post.requestBody.content.application/json.schema.$ref
  Line    :   18

[...]

PR #321, opened 2021-09-13
test: add test for external $ref

barrett-schonefeld:
Implements #301.

This test ensures that we do not return an error for $refs to external documents.

Because this test passes and I'm not able to reproduce the error outlined in #301, I did not provide an behavioral changes to the validator.


dpopp07:
This looks good. Since we aren't changing anything behaviorally and this is really just a confirmation of current behavior, let's hold off on merging this for now. Let's wait and see if the OP of the issue comes back with any more detail soon

PR #323, opened 2021-09-16
add configFileOverride to code validator

TheAzack9:
Currently is it impossible to specify where the .validatorrc can be found with the codevalidator. This minor change will allow the user to specify where the .validaterc is.

Currently without this change is it difficult for us to use IBM validator as the .validaterc might not be in a parent repository. The use case in our case is that we want to upload both a open api spec and a .validaterc to a separate service that validates the open api spec.

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -7,7 +7,7 @@ const spectralValidator = require('../spectral/utils/spectral-validator');
 const dedupFunction = require('../cli-validator/utils/noDeduplication');
 const { Spectral } = require('@stoplight/spectral');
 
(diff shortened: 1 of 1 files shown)


barrett-schonefeld:
This is a good idea!

I'm going to fully review this in the next two days. I want to make sure to fully consider if we should provide docs/tests with this change.

barrett-schonefeld:
A bit complicated to describe how to add a test, so I opened #331, which adds on to this initial PR!

barrett-schonefeld:
#331 merged

PR #324, opened 2021-09-16
feat: prohibit sentence style summaries

Andras-Csanyi:
This PR brings that when `summary` field has a trailing period,
an error will be displayed.

The following example will result a validation error.
```json
{
  "openapi": "3.0.0",
  "info": {
    "version": "1.0.0",
    "title": "Summary does not have trailing period"
  },
  "servers": {},
  "paths": {
    "createTrailingPeriod": {
      "post": {
        "operationId": "createTrailingPeriod",
        "description": "creates trailing period",
        "summary": "Adds trailing period."
      }
    }
  }
}
```

### DIFF
--- a/src/spectral/rulesets/.defaultsForSpectral.yaml
+++ b/src/spectral/rulesets/.defaultsForSpectral.yaml
@@ -232,3 +232,16 @@ rules:
     - $.paths[*][*].requestBody.content[*].schema
     then:
       function: string-boundary
+  # ensure summary field does not have trailing comma
+  prohibit-summary-sentence-style:
+    description: 'Summary should not have a trailing period'
+    message: "{{description}}"
+    severity: error
+    formats: ["oas3"]
+    resolved: false
+    given:
+    - $.paths[*][*].summary
+    then:
+      function: pattern
+      functionOptions:
+        notMatch: "\\.$"
--- a/src/spectral/rulesets/.defaultsForSpectral.yaml
+++ b/src/spectral/rulesets/.defaultsForSpectral.yaml
@@ -247,7 +247,7 @@ rules:
   prohibit-summary-sentence-style:
     description: 'Summary should not have a trailing period'
     message: "{{description}}"
-    severity: error
+    severity: warn
     formats: ["oas3"]
     resolved: false
     given:
--- a/test/spectral/tests/custom-rules/prohibit-sentence-style-summaries.test.js
+++ b/test/spectral/tests/custom-rules/prohibit-sentence-style-summaries.test.js
@@ -21,7 +21,7 @@ describe('spectral - summaries do not have trailing period', () => {
     };
 
     const result = await inCodeValidator(spec, true);
-    const relevantError = result.errors.find(
+    const relevantError = result.warnings.find(
       v => v.rule === 'prohibit-summary-sentence-style'
     );
     expect(relevantError).toBeUndefined();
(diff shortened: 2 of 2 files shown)


barrett-schonefeld:
I'd like @dpopp07 to weigh in on this because this seems a bit strict.

And I don't know that something that ends with `.` is a good enough measure of a sentence because it's easy to imagine using summaries like "Short non-sentence summary."

dpopp07:
It seems that the sentence structure is less the problem and the trailing period is really what we're after. Is that correct @hudlow? I believe you opened the original issue for this (2397 on our private board)

hudlow:
I actually think this is pretty reasonable given how we use the `summary` in our tooling, and would almost go even further and suggest we use this same rule to set some reasonable maximum length (I would suggest 64 characters as a possibility). But I'd like @cjcarpen to weigh in from a documentation perspective on whether these are supportable policies.

Andras-Csanyi:
> I actually think this is pretty reasonable given how we use the `summary` in our tooling, and would almost go even further and suggest we use this same rule to set some reasonable maximum length (I would suggest 64 characters as a possibility). But I'd like @cjcarpen to weigh in from a documentation perspective on whether these are supportable policies.

Thanks @hudlow ! Shall we wait for while the above viewpoint is discussed, or can we merge this PR in? What is your take?

src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -244,3 +243,16 @@ rules:
       - $.servers[*][variables][*][default]
     then:
       function: truthy
+  # ensure summary field does not have trailing comma
+  prohibit-summary-sentence-style:
+    description: 'Summary should not have a trailing period'
+    message: "{{description}}"
+    severity: error
```
hudlow, line 250: Let's make this a warning by default.

hudlow:
After more discussion, I think we should have a separate rule for length. Let's change this to a warning and get it merged. 👍

test/spectral/tests/custom-rules/prohibit-sentence-style-summaries.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+    const validationResult = await inCodeValidator(spec, true);
+    expect(validationResult.errors).not.toBeUndefined();
+    const error = validationResult.warnings.find(
+      v => v.rule === 'prohibit-summary-sentence-style'
+    );
+    expect(error).not.toBeUndefined();
+    const errorPath = ['paths', 'createTrailingPeriod', 'post', 'summary']
+      .sort()
+      .join('');
```
dpopp07, line 57: Is there a reason you sorted and joined here rather than checking for the explicit order the path was returned in like you do for the last test? I think this is fine but maybe not as robust as strictly checking the actual order

dpopp07 (approved):
Looks good! Left one comment to think about but it's up to you

barrett-schonefeld (approved):
I see Dustin's comment on the tests.

Otherwise, looks great!

hudlow (approved):

test/spectral/tests/custom-rules/prohibit-sentence-style-summaries.test.js
Andras-Csanyi: This was an initial solution for checking equality of two arrays. But @barrett-schonefeld showed me what help `jest` can offer here. I refactored this test too.

PR #328, opened 2021-09-21
feat: HEAD request response should not have body

Andras-Csanyi:
This PR bring s the capability to check whether a `HEAD` request
with status code `200` returns a body or not.

Introducing this rule makes conflict with the rule of `status code 200
should contain response body`.


src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -244,3 +244,14 @@ rules:
       - $.servers[*][variables][*][default]
     then:
       function: truthy
+  # ensure HEAD request does not have response body
+  head-request-should-not-have-response-body:
```
barrett-schonefeld, line 248: This rule is tough to name because we want it to be short yet descriptive. Those two qualities seem mutually exclusive here!

I think we should align with our other response-specific rules, `response-error-response-schema` and `response-example-provided`. I have a suggestion below. Feel free to suggest another name if you think of it!

```suggestion
  response-head-response-with-content:
```

barrett-schonefeld (approved):
We should try to simplify the name of the rule a bit, and I left a comment about this.

Once we have the name update, this is good to merge!

dpopp07:
@Andras-Csanyi what do you mean "Introducing this rule makes conflict with the rule of status code 200 should contain response body." ?

Is there something we need to address there? We don't want the user to get conflicting messages

barrett-schonefeld:
> Is there something we need to address there? We don't want the user to get conflicting messages

I missed this. Yes, we should update the existing:

    content-entry-provided:
      description: Request bodies and non-204 responses should define a content object
      given:
      - $.paths[*][*].responses[?(@property != '204' && @property != '202' && @property != '101')]
      - $.paths[*][*].requestBody
      severity: warn
      formats: ["oas3"]
      resolved: true
      then:
        field: content
        function: truthy

Specifically the path to the responses field:

      - $.paths[*][?(@property != 'HEAD' && @property != 'head' && @property != 'Head')].responses[?(@property != '204' && @property != '202' && @property != '101')]

Unfortunately, the properties are case sensitive, and I don't know of a way to get covert all `@property` chars to upper/lower for comparison, but the new path should look similar to above.

Revoking approval until we make the above update.

barrett-schonefeld (changes requested):
I provided a suggestion on how to fix the conflicting rules.

Good catch on the conflict!

dpopp07:
@barrett-schonefeld Would it be simpler to define a function to check all of those conditions?

barrett-schonefeld:
> Would it be simpler to define a function to check all of those conditions?

Although the update to the path is complicated, I think it is preferable to defining a custom function that handles the validation from the path level.

It would complicate things a bit because we would need a custom function that takes a path object, **only walks down non-HEAD operations**, walks to responses, **only walks down non-204, non-202, non-201 responses**, and checks if content is truthy.

dpopp07:
Okay, good to know! Thanks for the explanation

Andras-Csanyi:
@dpopp07 and @barrett-schonefeld thanks for the feedback! I apologise not being clear enough how the rules are conflicting.
Let me take a try with the custom function.

Andras-Csanyi:
Before we go down the custom function rabbit hole let me propose something. How about validating that the http verbs are must be lowercase? Spectral has inbuilt function for this (`casing`), so it is easy to validate it.
Once we ensured that the verbs are lowercase (`flat`), we can simplify (or keep it simple) the condition @barrett-schonefeld sketched previously.

https://meta.stoplight.io/docs/spectral/ZG9jOjExNg-core-functions#casing

barrett-schonefeld:
> Before we go down the custom function rabbit hole let me propose something. How about validating that the http verbs are must be lowercase?
> 
> https://meta.stoplight.io/docs/spectral/ZG9jOjExNg-core-functions#casing

Good idea, and I just realized that we already assume lowercase for other rules, so I think we can proceed with this rule by following the same pattern. For example:

    content-entry-contains-schema:
      description: Content entries must specify a schema
      given:
      - $.paths[*].[post,put,patch].requestBody.content[*]
      - $.paths[*].[get,post,put,patch,delete].[parameters,responses][*].content[*]
      severity: warn
      resolved: true
      then:
        field: schema
        function: truthy

Hence, for the `content-entry-provided` rule, I think we can use:

    - $.paths[*].[get,post,put,patch,delete].responses[?(@property != '204' && @property != '202' && @property != '101')]

And we can move forward with this PR.

Andras-Csanyi:
For summarising what should happen:
1. make changes in the `content-entry-contains-schema`, so that it doesn't care with case sensitivity when it goes deep other than `HEAD` verbs
2. introduce a new rule which validates that http verbs should be lowercase
3. finish/fix `response-head-response-with-content` (this name is suggested by @barrett-schonefeld )

1 and 2 possibly can be a separate PR if this one seems to too big already.

What do you think @barrett-schonefeld ?

dpopp07:
Closing, not a priority to add this rule, though valid, at the moment

PR #331, opened 2021-09-23
feat: add configFileOverride to in-code validator

barrett-schonefeld:
Expands on #323.

Provides parameter for custom `configFileOverride` file.

Adds a test to ensure the `configFileOverride` is passed to `processConfiguration.get`.

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -7,7 +7,7 @@ const spectralValidator = require('../spectral/utils/spectral-validator');
 const dedupFunction = require('../cli-validator/utils/noDeduplication');
 const { Spectral } = require('@stoplight/spectral');
 
-module.exports = async function(input, defaultMode = false) {
+module.exports = async function(input, defaultMode = false, configFileOverride = null) {
   // process the config file for the validations &
   // create an instance of spectral & load the spectral ruleset, either a user's
   // or the default ruleset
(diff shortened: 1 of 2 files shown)


dpopp07:
This is a good idea, thanks for the test. What do you think about adding an environment variable to do the same thing, so that the CLI validator is also not forced to look upwards for the config file? Since we're making this change already, it might be a good time to do that

barrett-schonefeld:
> What do you think about adding an environment variable to do the same thing...

Good idea!

> ...so that the CLI validator is also not forced to look upwards for the config file?

Are you suggesting that the environment variable should hold a path to the config file? Or is there another way you think we may be able to circumvent looking up for the config file?


Also, should we call this environment variable `CONFIG_FILE_OVERRIDE` or make it more specific with `VALIDATOR_CONFIG_FILE_OVERRIDE`?

Given the behavioral details from above and the name, I should be ready to move forward!

dpopp07:
> Are you suggesting that the environment variable should hold a path to the config file?

Yes! That is what I'm suggesting

As for the name, I don't think we need the word "override" in there. Something like `VALIDATOR_CONFIG_FILE` would be sufficient.

I also just realized that [the command line options](https://github.com/IBM/openapi-validator#options) include a method of specifying a path to the config file, so there is already a way of configuring that from the command line. Adding a new environment variable is really optional in light of that but still may be helpful. Up to you

barrett-schonefeld:
> I also just realized that [the command line options](https://github.com/IBM/openapi-validator#options) include a method of specifying a path to the config file, so there is already a way of configuring that from the command line. Adding a new environment variable is really optional in light of that but still may be helpful.

In that case, we may want to move this in-code feature along and see consider requests for an environment variable if we receive some requests in that direction.

dpopp07 (approved):
Looks good! 👍

dpopp07:
> In that case, we may want to move this in-code feature along and see consider requests for an environment variable if we receive some requests in that direction.

That works for me! Approved

PR #332, opened 2021-09-23
feat: examples' name should not contain space

Andras-Csanyi:
This PR brings the capability to validate whether examples' names
contain a space or not. Specification allows to put space in the examples'
name, but the generator fails.

If this case occurs the validator shows a `warning`.

The example below shows an example where the validator will show `warning`.

Note: commit messages don't really cover the TDD (ish) way. Sorry for that.

```yaml
openapi: 3.0.3
info:
  title: Examples name should not contain space
  description: Examples name should not contain space
  version: 1.0.0
servers:
  - url: 'https://example.com'
paths:
  /v1/users:
    get:
      operationId: get_users
      summary: returns user list
      responses:
        200:
          description: returns list of users
          content:
            'application/json':
              schema:
                properties:
                  id:
                    type: string
              examples:
                success example: #<-- this one
                  summary: 'successful example'
                  value: 'success value'
                failed example:  #<-- this one
                  summary: 'failed request'
                  value: 'failed request value'
```


src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -244,3 +244,17 @@ rules:
       - $.servers[*][variables][*][default]
     then:
       function: truthy
+  # ensure that examples name does not contain space
+  examples-name-should-not-contain-space:
```
Andras-Csanyi, line 248: I'm not sure this name is good enough. I'm open for suggestions.

dpopp07: We used to name rules after the problem it catches - something like `examples-name-contains-space`. We might use a different convention now. @barrett-schonefeld do you have an opinion?

src/spectral/rulesets/.defaultsForSpectral.yaml
```diff
@@ -244,3 +244,17 @@ rules:
...
+    message: "{{description}}"
+    severity: warn
+    resolved: false
+    formats: ["oas3"]
+    given:
+      - $.paths[*][*].responses[*][*][*].examples[*]~
+    then:
+      function: pattern
+      functionOptions:
+        notMatch: '[a-zA-Z0-9]\ +[a-zA-Z0-9]'
```
dpopp07, line 259: I don't think this would catch a space at the beginning or end, like " example_name". Would just `'\ '` be sufficient?

dpopp07 (approved):
I think the regex might need to change but I'll approve pending that change, looks good otherwise!

src/spectral/rulesets/.defaultsForSpectral.yaml
barrett-schonefeld: I think `examples-name-contains-space` is a good choice here.

src/spectral/rulesets/.defaultsForSpectral.yaml
barrett-schonefeld: Maybe this: `^(.*\s+.*)+$`

This regex is meant to positively match against any string with a space, and `notMatch` should negate this such that it fails when the string includes a space.

Source: https://stackoverflow.com/questions/9341379/check-string-contains-whitespace-along-with-some-other-char-sequence-using-regex

PR #333, opened 2021-09-27
fix: handle multi-file spec circular reference edge case

### LINKED ISSUE #326: TypeError: Cannot read properties of undefined (reading 'get')
Another issue with OpenApi as [here](https://github.com/IBM/openapi-validator/issues/319).

Need to make the error more clear to be able to find the place that course the error. 

image

 I'll send you @dpopp07  the specification and docker run command to email to easily reproduce the error.

dpopp07:
The validator chokes on a specific scenario - when a multi-file spec
has schemas that reference each other in a ciruclar manner across files.
The bundler we use to put together mutli-file specs doesn't store the
reference in a helpful way, so there's not really a way to report the
location of the circular reference the way we do for other scenarios.
All we can really do is avoid crashing completely and send the user the
fully resolved path to the problematic schema that they will need to
trace manually.

A test is added to capture the behavior. This test case would exhibit the
crash without the fix.

Resolves #326

### DIFF
--- a/test/cli-validator/tests/circular-refs-validator.test.js
+++ b/test/cli-validator/tests/circular-refs-validator.test.js
@@ -1,6 +1,9 @@
 const commandLineValidator = require('../../../src/cli-validator/runValidator');
 const circularRefsValidator = require('../../../src/cli-validator/utils/circular-references-ibm');
-const { getCapturedText } = require('../../test-utils');
+const {
+  getCapturedText,
+  getMessageAndPathFromCapturedText
+} = require('../../test-utils');
 
 describe('cli tool - test circular reference module', function() {
   it('should correctly validate a file with circular references', async function() {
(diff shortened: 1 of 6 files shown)


test/cli-validator/tests/circular-refs-validator.test.js
```diff
@@ -87,4 +87,28 @@ describe('cli tool - test circular reference module', function() {
...
+    const capturedText = getCapturedText(consoleSpy.mock.calls);
+    consoleSpy.mockRestore();
+
+    expect(exitCode).toEqual(0);
+
+    const allOutput = capturedText.join('');
+
+    expect(
+      allOutput.includes(
+        'Swagger object should not contain circular references.'
```
Legion112, line 110: It would be nice to see in the error where circular references have been found.

dpopp07: @Legion112 It is in the error, it's just not being checked in the test. Agreed though, that would be a good thing to test here. I'll add that

barrett-schonefeld (approved):
Looks good! I see that you plan to add a test, but approving so you can merge once you add the test.

PR #334, opened 2021-09-29
fix: false positive when minLength, maxLength are 0

barrett-schonefeld:


dpopp07 (approved):
Looks great! Nice add with the min/max length comparison too.

One question - do you think we also need to worry about `null` values? Sometimes `undefined` can be a little prescriptive

barrett-schonefeld:
> One question - do you think we also need to worry about `null` values? Sometimes `undefined` can be a little prescriptive

The API definition would have to explicitly contain a null value, such as `enum: null`, `minLength: null`, `maxLength: null`, etc., to cause a problem, and those seem like invalid values.

dpopp07:
> The API definition would have to explicitly contain a null value (enum: null, minLength: null, maxLength: null, pattern: null) to cause a problem, and that would be invalid for any of these properties

Right, if any of the properties are explicitly set to null, we won't catch them here (they will pass these checks). When you say that would be invalid, do you mean that they are validated somewhere else and so we don't need to worry about that here?

barrett-schonefeld:
> Right, if any of the properties are explicitly set to null, we won't catch them here (they will pass these checks). When you say that would be invalid, do you mean that they are validated somewhere else and so we don't need to worry about that here?

I was thinking it would be covered separately, but now that you pose the question again and I've thought about it more, it would be good to have a helper function here that checks for `null` and `undefined`.

dpopp07 (approved):
Ah I like the helper function! Looks good 👍

PR #335, opened 2021-10-08
fix: pac-resolver force dependency to 5.0.0

rmkeezer:
This change uses the `npm-force-resolutions` package as a dev-dependency to up-version the pac-resolver dependency to 5.0.0 from 4.2.0. This is a workaround until Spectral is upgraded to 6.0.0.

I also added the vulnerability ignore for the ansi-regex dev-dependency so the Ad-hoc scan is clean.

### DIFF
--- a/.cra/.cveignore
+++ b/.cra/.cveignore
@@ -1,0 +1,6 @@
+[
+    {
+        "cve": "CVE-2021-3807",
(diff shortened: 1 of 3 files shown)


barrett-schonefeld (approved):
Looks good!

PR #341, opened 2021-11-01
Path exclusion

### LINKED ISSUE #294: Any way to suppress specific errors/warnings without completely disabling a rule?
I have an existing spec which has some violations of rules that I would actually like to keep enabled to prevent further violations, but I'd like to ignore the existing ones. Is this currently possible?

Grahahax:
Fixes #294 

Draft to gather feedback on approach.

TODO: Add tests, address feedback etc

### DIFF
--- a/src/cli-validator/utils/processConfiguration.js
+++ b/src/cli-validator/utils/processConfiguration.js
@@ -35,21 +35,41 @@ const printConfigErrors = function(problems, chalk, fileName) {
   }
 };
 
+const validateExclusionsConfig = (configObject,configErrors) => {
+  //TODO more validation 
+  if(configObject.exclude)
+  {
+    const paths = Object.keys(configObject.exclude)
+    paths.forEach((path) => {
+      if(!Array.isArray(configObject.exclude[path]))
+      {
+        configErrors.push({
+          message: `Excluded path:${path} must specify an array of excluded rules`,
+          correction: `Change "${configObject.exclude[path]}" to be an array of named rules`
+        });
+      }
+    })
+  }
+}
+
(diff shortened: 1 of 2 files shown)


dpopp07:
@Grahahax I apologize for the long delay on reviewing this. We recently had to significantly refactor this tool and wanted to wait until we did that to have a look. That said, I think this seems like a solid approach. The one thing I want to consider before moving forward is [Spectral `overrides`](https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTg5-custom-rulesetsundefined#overrides). This is existing functionality within Spectral that I think meets the criteria you are looking for. We don't currently support it right now but I am working on getting that fixed. If we're able to do that, we would have path exclusions in place for all Spectral-formatted rules (we are working towards moving all validations to this format). If we're not able to get that working, I think we can move forward with this approach. I would just prefer to use existing technology than to add new code to the tool.

Grahahax:
@dpopp07 I would certainly prefer existing technology and if you are moving all validations to spectral format then that sounds like the way to go.

I'll leave this draft open until you decide if you can get all the validations working, at which point we can close or progress this change.

dpopp07:
@Grahahax Sounds good. I have a [PR](https://github.com/IBM/openapi-validator/pull/358) to enable Spectral `overrides` open now. Migrating all of our existing rules will be on ongoing effort but once that PR is merged in, you will be able to exclude specific paths for any Spectral-defined rules

dpopp07:
@Grahahax Thanks again for taking the time to put this together! Now that Spectral overrides are enabled, I'm going to close this draft

PR #349, opened 2021-12-16
feat: upgrade to spectral 6 and convert to monorepo

dpopp07:
This commit refactors the project into a monorepo and separates the project
into two NPM packages. One is the existing ibm-openapi-validator package and
one is the new package for our custom Spectral ruleset. As part of this change,
we upgraded to Spectral 6 which contained a number of breaking changes.

There are some incompatible changes for this tool (not many) as a result. These
are described in the migration guide. Since a number of these files were being
moved around already, I took the opportunity to rename a bunch of files to match
our kebab-case convention of file naming.


.travis.yml
```diff
@@ -10,7 +10,7 @@ script:
 deploy:
 - provider: script
   skip_cleanup: true
-  script: npx semantic-release
+  # script: npx -p @qiwi/semrel-toolkit multi-semrel
```
dpopp07, line 13: Will uncomment this before merging, just didn't want to make any unintentional moves before everything is in place

packages/ruleset/src/functions/required-property.js
```diff
@@ -34,7 +33,7 @@ function checkRequiredProperties(schema, path) {
         let message;
         if (schema.allOf) {
           message = `Required property, ${requiredPropName}, must be defined in at least one of the allOf schemas`;
-        } else if (schema.anyOf || schema.allOf) {
+        } else if (schema.anyOf || schema.oneOf) {
```
dpopp07, line 36: Caught a bug with the new tests 🙂

packages/ruleset/src/functions/string-boundary.js
```diff
@@ -1,6 +1,5 @@
-module.exports = function(schema, _opts, paths) {
-  const rootPath = paths.target !== void 0 ? paths.target : paths.given;
```
dpopp07, line 2: This line is no longer relevant with the new spectral - it is removed from a number of files

packages/ruleset/test/utils/root-document.js
```diff
@@ -0,0 +1,137 @@
+module.exports = {
+  openapi: '3.0.0',
```
dpopp07, line 2: This is a clean API definition that I use for the tests. Easier to edit and make one thing go wrong than it is to write new definitions a bunch

packages/validator/src/.defaultsForValidator.js
```diff
@@ -98,29 +98,6 @@ const defaults = {
...
-      'oas2-host-trailing-slash': "warning",
-      'oas2-valid-definition-example': "warning",
-      'oas2-anyOf': "warning",
-      'oas2-oneOf': "warning",
-      'oas3-api-servers': "warning",
-      'oas3-examples-value-or-externalValue': "warning",
-      'oas3-server-trailing-slash': "warning",
-      'oas3-valid-oas-content-example': "warning",
-      'oas3-valid-schema-example': "warning"
-    }
```
dpopp07, line 123: I decided we shouldn't support configuring Spectral rules from the `.validaterc` file anymore. It's just a confusing way to do it and it's tough to merge these rules in. If that is too heavy handed, let me know

packages/validator/src/cli-validator/run-validator.js
```diff
@@ -261,6 +254,13 @@ const processInput = async function(program) {
       if (debug) {
         console.log(err.stack);
       }
+      // this check can be removed once we support spectral overrides
+      if (err.message.startsWith('Document must have some source assigned.')) {
+        console.log(
+          'This error likely occurred because Spectral `exceptions` are deprecated and `overrides` are not yet supported.\n' +
+            'Remove these fields from your Spectral config file to proceed.'
+        );
```
dpopp07, line 262: Due to some bizarre behavior in Spectral, we can't support overrides yet. That will have to be a follow on change. This gives users a better error message than Spectral does

packages/validator/src/lib/index.js
```diff
@@ -17,20 +18,17 @@ module.exports = async function(
   // or the default ruleset
   let configObject;
   let spectralResults;
-  const spectral = new Spectral({
-    computeFingerprint: dedupFunction
-  });
```
dpopp07, line 22: We were duplicating too much setup code, I moved all of this into the `setup` function that is called later

packages/validator/test/cli-validator/alternate-spectral-configs/extends-default.yaml
```diff
@@ -1,4 +1,4 @@
-extends: ibm:oas
+extends: '@dpopp07/test-ibm-oas-ruleset'
```
dpopp07, line 1: I'll change all instances of this before merging

packages/validator/test/cli-validator/mock-files/oas3/testoneof.yaml
```diff
@@ -1,4 +1,6 @@
 openapi: 3.0.0
+servers:
+  - url: https://oneoftester.com/api
```
dpopp07, line 3: Spectral throws the craziest exception if an API def is missing a servers field lol. I plan to open an issue about it at some point

padamstx (approved):
Looks great to me!
Other than the version # issue I mentioned on slack, I can't find anything else to complain about!!! :)

hudlow (approved):
Looks good as far as I was able to understand it.

PR #350, opened 2021-12-16
build: run release script on node v14

dpopp07:
Bumping this up from v12 as semantic release now recommends
running with v14.


padamstx (approved):
LGTM

PR #353, opened 2021-12-22
fix: replace custom 'unused component' validation with spectral rule

dpopp07:
We had Spectral's "unused component" rules for OAS 2 and 3 turned off because
of our custom, unconfigurable rule validating a similar thing but covering
more scenarios. Now, the Spectral rules covers all scenarios that we do so
we can remove our rule enable the Spectral rules.


padamstx (approved):
LGTM

PR #357, opened 2022-01-07
fix(string-boundary): include parameters defined at path level

### LINKED ISSUE #352: `string-boundary` check doesn't appear to include `path` parameters
Our API includes many parameters such as:

```
  parameters:
    snapshotId:
      name: snapshot_id
      in: path
      description: |-
        The snapshot identifier
      required: true
      schema:
        type: string
```

(Note the use of `type: string` without any `minLength`, `maxLength`, or `pattern` qualifiers)

However, these do not appear to included in the validator's `string-boundary` check, while `query` parameters are (correctly)  checked.

dpopp07:
Currently, the custom string-boundary rule only validates parameters
defined at the operation level and request body properties. However,
parameters can also be defined at the path level. This commit adds
that scenario to the rule.

Resolves #352 

FYI @meem

### DIFF
--- a/packages/ruleset/src/rules/string-boundary.js
+++ b/packages/ruleset/src/rules/string-boundary.js
@@ -8,6 +8,8 @@ module.exports = {
   formats: [oas3],
   resolved: true,
   given: [
+    '$.paths[*][parameters][*].schema',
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

PR #358, opened 2022-01-11
feat: enable spectral overrides

dpopp07:
This commit updates the format in which we provide the document data to spectral,
which in turns allows end users to take advantage of spectral "overrides". This
feature allows users to ignore specific parts of their document if they need,
rather than turning off entire rules.

In doing this, we give spectral greater control over the validations it returns.
This allows us to take advantage of another spectral feature - automatically
resolving component paths. Before, if a schema contained a problem and two references
pointed to it, some rules would return two validations pointing to the reference
instances. Now, the validation returns one validation, pointing to the base schema.
This will result in some output changes for some users but it should be a positive
change overall.

This commit also adds a customized verson of spectral's deduplication logic to prevent
the resolved component paths from being duplicated.

### DIFF
--- a/packages/validator/src/cli-validator/run-validator.js
+++ b/packages/validator/src/cli-validator/run-validator.js
@@ -9,6 +9,9 @@ const path = require('path');
 const readYaml = require('js-yaml');
 const util = require('util');
 
+const { Document } = require('@stoplight/spectral-core');
+const Parsers = require('@stoplight/spectral-parsers');
+
 const addPathsToComponents = require('./utils/add-paths-to-components');
 const buildSwaggerObject = require('./utils/build-swagger-object');
 const config = require('./utils/process-configuration');
@@ -246,9 +249,14 @@ const processInput = async function(program) {
         chalk
       );
 
-      process.chdir(path.dirname(validFile));
-      // let spectral handle the parsing of the original swagger/oa3 document
-      spectralResults = await spectral.run(originalFile);
+      const fileExtension = ext.getFileExtension(validFile);
+      let parser = Parsers.Json;
+      if (['yaml', 'yml'].includes(fileExtension)) {
+        parser = Parsers.Yaml;
+      }
+
+      const doc = new Document(originalFile, parser, validFile);
+      spectralResults = await spectral.run(doc);
     } catch (err) {
       printError(chalk, 'There was a problem with spectral.', getError(err));
       if (debug) {
@@ -271,9 +279,6 @@ const processInput = async function(program) {
       }
       exitCode = 1;
       continue;
-    } finally {
-      // return the working directory to its original location
-      process.chdir(originalWorkingDirectory);
     }
 
     // run validator, print the results, and determine if validator passed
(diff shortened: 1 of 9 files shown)


packages/validator/src/cli-validator/run-validator.js
```diff
@@ -246,9 +249,14 @@ const processInput = async function(program) {
...
-      process.chdir(path.dirname(validFile));
-      // let spectral handle the parsing of the original swagger/oa3 document
-      spectralResults = await spectral.run(originalFile);
+      const fileExtension = ext.getFileExtension(validFile);
+      let parser = Parsers.Json;
+      if (['yaml', 'yml'].includes(fileExtension)) {
+        parser = Parsers.Yaml;
+      }
+
+      const doc = new Document(originalFile, parser, validFile);
```
dpopp07, line 258: This is the meat of the change - using a Spectral `Document` is required for `overrides`

packages/validator/src/cli-validator/utils/custom-deduplication.js
```diff
@@ -0,0 +1,24 @@
...
+  if (rule.source !== void 0) {
+    id += rule.source;
+  }
+
+  if (rule.message !== void 0) {
+    id += rule.message;
+  }
+
+  return hash(id);
+};
```
dpopp07, line 24: I'm going to open an issue (and possibly a PR) against Spectral to see if they'll update their default implementation. It causes problems for their own rules. If they agree to it, we can delete this code. For now, we need it

packages/validator/test/cli-validator/tests/expected-output.test.js
```diff
@@ -377,12 +377,12 @@ describe('test expected output - OpenAPI 3', function() {
 
   it('should include the path to the component (if it exists) when in verbose mode', async function() {
     const program = {};
-    program.args = ['./test/cli-validator/mock-files/oas3/testoneof.yaml'];
+    program.args = ['./test/cli-validator/mock-files/oas3/component-path-example.yaml'];
```
dpopp07, line 380: Updating the code resolved all the validations present in `testoneof`, so component paths are not relevant anymore. I wrote a new API definition that includes unresolved validations from our own rules to test the "component paths" feature.

packages/validator/test/cli-validator/tests/info-and-hint.test.js
```diff
@@ -35,7 +35,7 @@ describe('test info and hint rules - OAS3', function() {
     expect(jsonOutput['errors'].length).toBe(4);
 
     // Verify warnings
-    expect(jsonOutput['warnings'].length).toBe(37);
+    expect(jsonOutput['warnings'].length).toBe(16);
```
dpopp07, line 38: All of these were duplicated validations spread across different references of the same schema

padamstx (approved):
LGTM

hudlow (approved):

packages/validator/src/cli-validator/run-validator.js
```diff
@@ -246,9 +249,14 @@ const processInput = async function(program) {
         chalk
       );
 
-      process.chdir(path.dirname(validFile));
-      // let spectral handle the parsing of the original swagger/oa3 document
-      spectralResults = await spectral.run(originalFile);
+      const fileExtension = ext.getFileExtension(validFile);
+      let parser = Parsers.Json;
+      if (['yaml', 'yml'].includes(fileExtension)) {
```
hudlow, line 254: I guess you weren't appalled by the way I did this. 😂

dpopp07: Haha nope! Maybe not the exact same as I would've done but functionally the same. And this code was already there 🙂

PR #365, opened 2022-01-29
Add new response-array-required rule

hudlow:


hudlow:
Abandoned for now — we need more discussion on the guidance behind this rule. Utilities released as a part of #424.

PR #367, opened 2022-02-01
remove broken discriminator rule in favor of new spectral rule

### LINKED ISSUE #360: [Bug] False positive: "The discriminator property name used must be defined in this schema"
Hello, I'm evaluating this tool for use with our project and have spotted an issue with one of the rules that we can't actually opt out of that produces the following error against my spec:

> The discriminator property name used must be defined in this schema

The error appears to be produced by the `builtin` rule, which I can't seem to opt-out of, but I _believe_ that its a false positive due to the way that we're using `allOf` references in our polymorphic object definition. I have put together the following schema to reproduce this issue:

<details> 

<summary> <b>Schema</b> </summary>

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "Test",
    "version": "0.0.1"
  },
  "paths": {

  },
  "components": {
    "schemas": {
      "Polymorphic": {
        "anyOf": [
          {
            "$ref": "#/components/schemas/Foo"
          },
          {
            "$ref": "#/components/schemas/Bar"
          }
        ],
        "discriminator": {
          "propertyName": "type",
          "mapping": {
            "foo": "#/components/schemas/Foo",
            "bar": "#/components/schemas/Bar"
          }
        }
      },
      "Foo": {
        "type": "object",
        "required": [
          "type"
        ],
        "properties": {
          "type": {
            "type": "string",
            "enum": ["foo"]
          },
          "foo": {
            "type": "string"
          }
        }
      },
      "Bar": {
        "type": "object",
        "required": [
          "type"
        ],
        "properties": {
          "type": {
            "type": "string",
            "enum": ["bar"]
          },
          "bar": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

</details>

**Note 1:** This technique follows guidance described in Swaggers docs: https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism

**Note 2:** `mapping` isn't necessarily related, but I use it in my real use-case so wanted to include it. You can also achieve the same using `anyOf` and `oneOf`. I can't remember the reason why we use `anyOf` but I imagine that we'd need to support both.

When validating the above schema, I get the following output:

```
$ lint-openapi schema.json

errors

  Message :   The discriminator property name used must be defined in this schema
  Path    :   components.schemas.Polymorphic.discriminator.propertyName
  Line    :   22
```

I think that I was able to trace it back to the following code:

https://github.com/IBM/openapi-validator/blob/668f2e7bd94914ebc8c4eb4a63a7ff776f7c9706/packages/validator/src/plugins/validation/oas3/semantic-validators/discriminator.js#L40-L49

The check here is relatively simple and is expecting that `properties` contains `type`, but it doesn't account for `allOf`/`oneOf`/`anyOf`. 

If I can help to patch this, please let me know!  I'm not familiar with the project so I'm not sure of the approach that we should be taking to resolve an absolute list of properties. Alternatively I guess that having a way to just disable this rule could also be an option.

Thanks!

dpopp07:
This rule was buggy and not previously configurable. The same rule with
identical (albeit fixed) behavior is now available in our Spectral ruleset
under the name 'discriminator' and is configurable.

I also did a lot of refactoring to reduce duplication and make it easier to add new
rules to our spectral ruleset.

Resolves #360


packages/ruleset/src/collections/index.js
```diff
@@ -0,0 +1,19 @@
...
+//
+// note that this does not include "components.schemas" to avoid duplication.
+// this collection should be used in a rule that has "resolved" set to "true".
+// we separately validate that all schemas in "components" need to be used.
+const schemas = [
+  '$.paths[*][parameters][*].schema',
+  '$.paths[*][parameters][*].content[*].schema',
+  '$.paths[*][*][parameters][*].schema',
+  '$.paths[*][*][parameters,responses][*].content[*].schema',
+  '$.paths[*][*][requestBody].content[*].schema'
```
hudlow, line 14: Don't these all need recursion to account for schemas referenced/embedded in object schema properties?

README.md
```diff
@@ -496,32 +496,32 @@ The default values for each rule are described below.
...
+oas2-unused-definition: true
 oas3-api-servers: true
 oas3-examples-value-or-externalValue: true
 oas3-server-trailing-slash: true
+oas3-valid-media-example: 'warn'
+oas3-valid-schema-example: 'warn'
 oas3-schema: true
-oas3-valid-example: true
-oas3-valid-schema-example: true
+oas3-unused-component: true
```
hudlow, line 524: Not sure I see what the new order is here. Any possibility we could alphabetize? Any possibility we could enforce correctness (and alphabetical order) in a build check?

packages/ruleset/src/utils/validate-subschemas.js
```diff
@@ -0,0 +1,36 @@
+const validateSubschemas = (schema, path, validate) => {
```
hudlow, line 1: Can't JSONPath do all this already? Or maybe not precisely enough?

README.md
dpopp07: These hadn't been updated since the Spectral major release upgrade. I copied and pasted them from our ruleset file so they'd be 100% accurate. I didn't think much about the ordering. I think enforcing some kind of correctness with the build is a good idea but probably outside the scope of this PR

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: I'm not skilled enough in JSONPath to know how to do this. The recursive strategy was already being used by two of our rules so I generalized it. Do you know how to JSONPath subschemas? Happy to alter the approach if there's a better way!

packages/ruleset/src/collections/index.js
dpopp07: Addressed in the functions - let's use the thread off your other comment to discuss

hudlow: Well, I was thinking along the lines of this (for request body schemas, as an example):

```
'$.paths[*][*][requestBody].content[*].schema',
'$.paths[*][*][requestBody].content[*].schema..properties[*]',
'$.paths[*][*][requestBody].content[*].schema..items'
```

But now that I look closely, I think that could result in false positives for object schema properties named `properties` or `items` or for the use of those key names within extensions. So... maybe there is no way around the manual recursion. :-(

packages/ruleset/src/utils/validate-subschemas.js
hudlow: Well, I was thinking along the lines of this (for request body schemas, as an example):

```
'$.paths[*][*][requestBody].content[*].schema',
'$.paths[*][*][requestBody].content[*].schema..properties[*]',
'$.paths[*][*][requestBody].content[*].schema..items'
```

But now that I look closely, I think that could result in false positives for object schema properties named `properties` or `items` or for the use of those key names within extensions. So... maybe there is no way around the manual recursion. 😢

On the other hand, it looks like built-in Spectral rules are willing to use schema-finding approaches that appear to have the possibility of false positives: https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/index.ts#L674

hudlow, line 1: How confident are we in the term "subschemas" here? Should we reconsider it or at least define it in a header comment? Perhaps:

```
Subschemas include property schemas (for an object schema), item schemas (for an array schema), and applicator schemas (such as those in an `allOf`, `anyOf`, or `oneOf` property), plus all subschemas _of those schemas_.
```

packages/ruleset/src/utils/validate-subschemas.js
```diff
@@ -0,0 +1,36 @@
...
+          [...path, 'properties', property[0]],
+          validate
+        )
+      );
+    }
+  } else if (schema.items) {
+    errors.push(
+      ...validateSubschemas(schema.items, [...path, 'items'], validate)
+    );
+  } else if (schema.allOf || schema.anyOf || schema.oneOf) {
```
hudlow, line 21: I know we don't do much with it in our tooling, but is there any reason for us to ignore `not` here? I believe it is completely legal/valid for OAS 3.0.

padamstx (approved):
Looks great!

packages/ruleset/src/utils/validate-subschemas.js
```diff
@@ -0,0 +1,36 @@
...
+  // invoke validation function
+  errors.push(...validate(schema, path));
+
+  // recursively process subschemas
+  if (schema.properties) {
+    for (const property of Object.entries(schema.properties)) {
+      errors.push(
+        ...validateSubschemas(
+          property[1],
+          [...path, 'properties', property[0]],
```
hudlow, line 12: Since we're building the path ourself here (from what I assume is a "resolved" document, is it safe to assume Spectral will back-solve for the real path from this resolved path?

packages/ruleset/src/utils/validate-subschemas.js
```diff
@@ -0,0 +1,36 @@
+const validateSubschemas = (schema, path, validate) => {
+  const errors = [];
+  // invoke validation function
+  errors.push(...validate(schema, path));
+
+  // recursively process subschemas
```
hudlow, line 6: Are Spectral's resolved documents somehow protected from loops or do we need to do some loop protection of our own?

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: It was the first thing that came to me so I'm not particularly attached to it but I feel it describes the intent (all schemas "below" the current schema). I'm open to suggestions, though. And I like the idea of adding that comment, I can make that change

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: Actually, I have no idea if Spectral protects its resolved documents from loops (I hope that they would).

However, I did add loop protection in the early days of this tool and we have test cases with intentionally placed circular references. Our existing loop protection doesn't apply to Spectral but I believe we run the Spectral rules on these test cases anyways. I think that's evidence that Spectral is doing some kind of protection, or those test cases should infinitely loop with these existing recursive validations. I can double check this.

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: Yes, we return the actual path from the resolved document to Spectral and they will back-solve for the real path when returning results (assuming that the results are coming from a file and not an inline JSON object. I don't know why Spectral requires the filepath to back-solve but they do).

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: Candidly, I was not aware that existed. I suppose I don't have a good reason to ignore that, I can look into it

packages/ruleset/src/utils/validate-subschemas.js
hudlow: I think the definition I proposed above is consistent with how JSON Schema itself uses the word but I still think it might be worth defining our own use.

padamstx: "validateSchemaSubTree()" :)

packages/ruleset/src/utils/validate-subschemas.js
hudlow: It is called out in OAS 3.0 (and it's the only other supported "applicator"): https://spec.openapis.org/oas/v3.0.3#schema-object

But reviewing that section of the spec reminded me that we also appear to be missing support for `additionalProperties` schemas. 😁

hudlow: Also... I think almost all these validations should be independent. For example, a schema could have `properties`, and `oneOf` and `not`, and `additionalProperties`.

Of course, there are good reasons for us to prohibit certain combinations (like `anyOf` and `oneOf`) but I kind of thing all these validations (even `items` and `properties`) should be checked independently in this general-purpose function.

padamstx: good catch... I'm embarassed that I didn't catch that myself :)

README.md
hudlow: @dpopp07 should I open an issue?

dpopp07: Sure, that would be great

hudlow: Done: https://github.com/IBM/openapi-validator/issues/368

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: Okay I confirmed that there is protection, circular references are not an issue here 👍

packages/ruleset/src/utils/validate-subschemas.js
dpopp07: This has now been addressed - `not` and `additionalProperties` are now included and all forms of subschemas are independently validated

packages/ruleset/src/utils/validate-subschemas.js
```diff
@@ -0,0 +1,72 @@
...
+    processComposedSchemas(schema.oneOf, 'oneOf');
+  }
+
+  if (schema.anyOf) {
+    processComposedSchemas(schema.anyOf, 'anyOf');
+  }
+
+  if (schema.not) {
+    processComposedSchemas(schema.not, 'not');
+  }
```
hudlow, line 67:
```suggestion
  for (const applicatorType of ['allOf', 'oneOf', 'anyOf', 'not']) {
    if (Array.isArray(schema[applicatorType])) {
      schema[applicatorType].forEach((s, i) => {
        errors.push(
          ...validateSubschemas(s, [...path, whichType, i], validate)
        );
      });
    }
  }
```

dpopp07: Ah, nice! No good reason. In my head, I wanted each schema checked to be its own "if" statement but it also bothered me that I was doing the same thing four times in a row. I like this much better! There's a small typo in your suggestion so I'm just going to apply the change manually and re-push

packages/ruleset/test/validate-subschemas.test.js
```diff
@@ -0,0 +1,102 @@
...
+      function: ruleFunction
+    }
+  };
+
+  it('should find all subschemas', async () => {
+    await testRule('rule-name', ruleForTesting, allSchemasDocument);
+
+    expect(visitedSchemas[0]).toBe(
+      'paths./schema.get.responses.200.content.application/json.schema'
+    );
```
hudlow, line 31: Should we do this style instead?
```
    expect(visitedSchemas).toContain(
      'paths./schema.get.responses.200.content.application/json.schema'
    );
```

dpopp07: I intentionally chose not to use that style in order to make the test say "we found these schemas _and only_ these schemas". In essence, to verify that there were no false positives in addition to no missing positives. I suppose my test still doesn't fully capture that though, because there is no check on the length of the array. I'll add that now

hudlow: Okay, sounds good!

hudlow (approved):

PR #375, opened 2022-02-14
Convert property-case-convention rule to Spectral format

dpopp07:
Removes old, JS `property-case-convention` rule in favor of a new Spectral rule. The new rule has the same name and behavior but the configuration has changed to match Spectral's case convention option (e.g. 'lower_snake_case' is now 'snake'). I did some customization of the the Spectral results to make them a bit more helpful as well.

See [docs on the casing function](https://meta.stoplight.io/docs/spectral/ZG9jOjExNg-core-functions#casing) for more info.

Some tests had to be updated and new tests are added. The new tests cover all I really thought was necessary due to the testing of the helper functions, etc. elsewhere.

I also took the opportunity to update the README with the recent new rule conversions, I had forgotten about that previously.


packages/validator/test/cli-validator/mock-files/oas3/err-and-warn.yaml
```diff
@@ -163,7 +163,7 @@ components:
       description: string
       required:
         - name
-        - photoUrls
+        - photo_urls
```
dpopp07, line 166: This was causing a couple of errors to pop up due to Spectral reporting each path initially, so I just removed it. We didn't need it for the purpose of this test API definition

packages/validator/test/cli-validator/tests/option-handling.test.js
```diff
@@ -119,25 +119,6 @@ describe('cli tool - test option handling', function() {
...
-       1  (20%) : Property names must follow case convention
-
-      warnings
-       1  (13%) : Potentially unused definition has been detected.
-       2  (25%) : Operations must have a non-empty `operationId`.
-       2  (25%) : A tag is defined but never used
-       1  (13%) : Schema must have a non-empty description.
-       1  (13%) : Schema properties must have a description with content in it.
-       1  (13%) : A security scheme is defined but never used
-    */
```
dpopp07, line 139: This had gotten out of date and is obnoxious to keep up to date, so I just removed it

padamstx: Agreed, much easier to just print out the text via console.log temporarily.

padamstx (approved):
Looks great!

PR #376, opened 2022-02-14
fix: pin spectral dependencies to avoid configuration bugs

dpopp07:
This updates Spectral dependencies to their latest versions and pins them so that
they don't change underneath us and impact our users. We will have to manually update
all Spectral dependencies from now on, which I think is worth it for the time being.

The latest Spectral core release also includes my fixes to the de-duplication logic,
so I removed our custom function from this code. We can rely on the default Spectral
behavior now.


padamstx (approved):
LGTM

PR #377, opened 2022-02-15
feat: convert 'property-case-collision' rule to spectral style

pyrooka:
This PR converts the old JS style `property-case-collision` rule to Spectral style. This means a new rule has been created with the same logic and the old one has been removed. Tests got updated too.


packages/ruleset/src/functions/property-case-collision.js
```diff
@@ -0,0 +1,32 @@
+const { validateSubschemas } = require('../utils');
+
+module.exports = function(schema, _opts, { path }) {
+  return validateSubschemas(schema, path, propertyCaseCollision);
+};
+
+const errorMsg = 'Property names should not be identical';
```
dpopp07, line 7: I think a slightly more descriptive message would be helpful, since the property names are _technically_ not identical

packages/ruleset/src/rules/property-case-collision.js
```diff
@@ -0,0 +1,15 @@
+const { oas2, oas3 } = require('@stoplight/spectral-formats');
+const { propertyCaseCollision } = require('../functions');
+const { schemas } = require('../collections');
+
+module.exports = {
+  description: 'Property names should not be identical',
```
dpopp07, line 6: Since the actual message printed for the user is coming from the error, this can also be more descriptive about the rule itself. For context, the purpose of the rule is to prevent tooling from generating colliding property names after converting to a consistent case convention

dpopp07 (changes requested):
The logic looks great! I think we need to change up the messaging before merging though for the best user experience

pyrooka:
@dpopp07 What do you think now? Feel free to correct/suggest, maybe I still didn't word it right...

packages/ruleset/src/functions/property-case-collision.js
```diff
@@ -0,0 +1,32 @@
+const { validateSubschemas } = require('../utils');
+
+module.exports = function(schema, _opts, { path }) {
+  return validateSubschemas(schema, path, propertyCaseCollision);
+};
+
+const errorMsg = 'Property names should not be identical, even if different naming conventions are used';
```
padamstx, line 7: My first reaction to this error message is that 'identical' might not be the right word, precisely speaking.   I think "duplicate" might be a better word in this case since duplicate implies that the two properties are the same in some aspect but not necessarily "identical".
```suggestion
const errorMsg = 'Avoid duplicate property names within a schema, even if different case conventions are used';
```

packages/ruleset/src/functions/property-case-collision.js
```diff
@@ -0,0 +1,32 @@
...
+
+    for (const [propName, prop] of Object.entries(schema.properties)) {
+      // Skip check for deprecated properties.
+      if (prop.deprecated === true) continue;
+
+      const caselessPropName = propName.replace(/[_-]/g, '').toLowerCase();
+      if (prevProps.includes(caselessPropName)) {
+        errors.push({
+          message: errorMsg,
+          path
```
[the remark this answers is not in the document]
dpopp07: Yes, good catch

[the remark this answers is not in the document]
pyrooka: That's right, I totally missed that!

padamstx (approved):
LGTM!

PR #380, opened 2022-02-19
feat(enum-case-convention): add new enum-case-convention rule, remove old rule

padamstx:
This commit replaces the old enum_case_convention rule with a new
spectral-style enum-case-convention rule that will enforce
the restriction that string-based enum values use snake case.


padamstx:
@pyrooka, @dpopp07 and I discussed the case convention topic today and decided to hard-code the enum-case-convention and property-case-convention rules so they enforce snake-case.  This was mainly due to how difficult it is for users to change the configuration of one of our rules, so we went with just implementing the rules to enforce snake-case.
This PR has been updated to reflect those changes.

pyrooka (approved):
Still looks good! :)

padamstx:
Re-based this PR on top of the "staging" branch.

padamstx:
Discussed the case convention stuff with @dpopp07 again yesterday and we decided to reverse our earlier decision...  so the enum-case-convention and property-case-convention rules have been updated slightly to get the "casing config" from the rule definition.   Both rules are now configured to use snake case by default, but a user could override that by re-defining the rule in their custom ruleset and changing the config to a different casing style.

dpopp07 (approved):
Looks good! 👍

PR #381, opened 2022-02-20
fix: display details for 'Error running Nimma' errors

padamstx:
This commit makes a small change to our exception handling
so that if spectral returns a "Error running Nimma" exception,
we'll print out the details of the exception even if debug is
not currently enabled.


padamstx:
Any easy way to test this out is to pull the branch, then run:
```
node packages/validator/src/cli-validator/index.js -sv <api definition>
```
One easy way to trigger a relevant error is to set the server url in the API definition to "https://" :)

dpopp07 (approved):
Looks good! 👍

PR #382, opened 2022-02-22
Refactor and fix bugs in `checkSubschemasForProperty()` utility function

hudlow:


padamstx:
@hudlow could you please add a PR description to briefly explain what the original problem was?  That'll help me review the changes :)   Thanks

packages/ruleset/test/check-subschemas-for-prop.test.js
```diff
@@ -0,0 +1,183 @@
...
+      checkSubschemasForProperty(compliantSimpleSchema, 'my_property')
+    ).toBe(true);
+  });
+
+  it('should return `false` for a schema with empty `oneOf`', async () => {
+    const schemaWithEmptyOneOf = { oneOf: [] };
+    expect(
+      checkSubschemasForProperty(schemaWithEmptyOneOf, 'my_property')
+    ).toBe(false);
+  });
```
hudlow, line 28: This test failed.

packages/ruleset/test/check-subschemas-for-prop.test.js
```diff
@@ -0,0 +1,183 @@
...
+
+  it('should return `false` for a schema with empty `anyOf`', async () => {
+    const schemaWithEmptyOneOf = { anyOf: [] };
+    expect(
+      checkSubschemasForProperty(schemaWithEmptyOneOf, 'my_property')
+    ).toBe(false);
+  });
```
hudlow, line 35: This test failed.

packages/ruleset/test/check-subschemas-for-prop.test.js
```diff
@@ -0,0 +1,183 @@
...
+        { properties: { my_property: {} } },
+        { properties: { my_property: {} } }
+      ],
+      oneOf: [{ properties: { my_property: {} } }, {}],
+      allOf: [{}, {}]
+    };
+    expect(
+      checkSubschemasForProperty(schemaWithOnlyAnyOfCompliance, 'my_property')
+    ).toBe(true);
+  });
```
hudlow, line 123: This test failed.

packages/ruleset/test/check-subschemas-for-prop.test.js
```diff
@@ -0,0 +1,183 @@
...
+  it('should return `true` for `allOf` compliance even without `oneOf` or `anyOf` compliance', async () => {
+    const schemaWithOnlyAllOfCompliance = {
+      allOf: [{}, { properties: { my_property: {} } }, {}],
+      oneOf: [{}, { properties: { my_property: {} } }],
+      anyOf: [{ properties: { my_property: {} } }, {}]
+    };
+    expect(
+      checkSubschemasForProperty(schemaWithOnlyAllOfCompliance, 'my_property')
+    ).toBe(true);
+  });
```
hudlow, line 134: This test failed.

packages/ruleset/src/utils/check-subschemas-for-prop.js
```diff
@@ -1,36 +1,15 @@
...
-      }
-    }
-  }
-
-  return propertyIsDefined;
+/**
+ * This function will return `true` if all possible variations of a (possibly) composite schema
+ * define a property with the specified name.
+ */
+const checkSubschemasForProperty = (schema, name) => {
```
hudlow, line 8: @dpopp07 @padamstx We might want to consider if this function is misnamed. Here, "subschemas" seems to mean something very different than what it means in `validateSubschemas()` which includes schemas embedded in `properties`, `items`, etc.

packages/ruleset/src/utils/check-composite-schema-for-constraint.js
```diff
@@ -0,0 +1,65 @@
...
+ * This function will return `true` if all possible variations of a (possibly) composite schema are
+ * required to have a constraint, as checked by a lambda function which checks for the constraint on
+ * a simple (non-composite) schema. It does this by recursively checking if:
+ *
+ * - a schema has the constraint itself, or
+ * - one of the schemas it composes with `allOf` has the constraint, or
+ * - all of the schemas it composes with `oneOf` have the constraint, or
+ * - all of the schemas it composes with `anyOf` have the constraint
+ */
+const checkCompositeSchemaForConstraint = (schema, hasConstraint) => {
```
hudlow, line 13: @dpopp07 @padamstx I'm not sure if "composite" and "compose" are good ways to differentiate the nested schemas this relates to vs all "subschemas." Looking back at OO concepts, it looks like "composed" schemas would usually mean the equivalent of schemas referred to in `properties`. Maybe we should refer to schemas in `allOf`, etc as "combined" schemas and refer to the combining schema as a "complex" schema? That's what the Swagger documentation does: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/

Alternatively, we could refer to "complex" schemas containing "applied" (?) schemas based on the JSON Schema terminology: https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.5

padamstx (approved):
LGTM

packages/ruleset/src/utils/check-composite-schema-for-constraint.js
padamstx: If possible let's continue using the terms that we might already be familiar with instead of new ones.
I think we have typically used the term "composed schema" (or "composed model") to mean a schema that is defined with an allOf/anyOf/oneOf, so I'm ok with the current name `checkCompositeSchemaForConstraint`.  I didn't get tripped up on the name when reviewing the PR :)

packages/ruleset/src/utils/check-subschemas-for-prop.js
dpopp07: That's true, they are checking different things, so I'm not opposed to a rename. `validateSubschemas()` does check all potential "schema" objects underneath a schema but this does not. I would support a more specific name that captures what this does better if you'd like to change it

packages/ruleset/src/utils/check-subschemas-for-prop.js
```diff
@@ -1,36 +1,15 @@
...
-      if (checkSubschemasForProperty(s, name)) {
-        propertyIsDefined = true;
-        break;
-      }
-    }
-  }
-
-  return propertyIsDefined;
+/**
+ * This function will return `true` if all possible variations of a (possibly) composite schema
```
dpopp07, line 5: Something about this wording seems not quite right to me, since the schema may just have properties and not be a "composite" schema. Maybe something like
```suggestion
 * This function will return `true` if a schema, or all possible variations of a (possibly) composite schema,
```

Do you know what I mean? Feel free to disagree. I don't know if my wording really accomplishes much, it just made me scratch my head for a second that this function would be specifically focused first and foremost on composite schemas (even if the composite-ness was only a "possibility" 🙂 ).

Or maybe even just
```suggestion
 * This function will return `true` if all possible variations of a (possibly composite) schema
```

dpopp07 (approved):
Looks good! Left some thoughts on the naming but overall I think it's fine as is. The PR is definitely an improvement 👍

dpopp07 (approved):
Looked over the most recent changes, it all looks good!

PR #386, opened 2022-02-24
Minor fixes

dpopp07:
First commit: remove .npmignore file in favor of .gitignore

If there is no .npmignore file, NPM will use the patterns within the
.gitignore instead. Since we don't want anything in the .gitignore file
to be published and there is nothing outside the .gitignore that we don't
want published, it seems best to rely on the .gitignore file for now.

Second commit: add debug option to programmatic validator api

This allows those using the in-code validator to print out debugging statements
that would otherwise be swallowed.

Also, remove warning about running the validator in default mode. We no longer
want to be pushing that method of configuration and it is confusing when printed
alongside Spectral config file warnings.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,0 @@
-node_modules
-.DS_Store
-npm-debug.log*
-.eslintcache
-.validaterc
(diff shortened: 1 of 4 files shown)


padamstx (approved):
LGTM

PR #387, opened 2022-02-25
feat(parameter-description):  add new 'parameter-description' rule and remove old rule

padamstx:
This commit converts the old no_parameter_description rule
to a new Spectral-style 'parameter-description' rule.

### DIFF
--- a/packages/ruleset/src/functions/parameter-description.js
+++ b/packages/ruleset/src/functions/parameter-description.js
@@ -1,0 +1,24 @@
+module.exports = function(param, _opts, { path }) {
+  return parameterDescription(param, path);
+};
+
+function parameterDescription(param, path) {
+  if (!shouldSkip(param) && !paramHasDescription(param)) {
+    return [
+      {
+        message: 'Parameter should have a non-empty description',
+        path
+      }
+    ];
+  }
+
+  return [];
+}
+
+function paramHasDescription(param) {
+  return param.description && param.description.toString().trim().length;
+}
+
+function shouldSkip(param) {
+  return param['x-sdk-exclude'] === true;
+}
--- a/packages/ruleset/test/array-of-arrays.test.js
+++ b/packages/ruleset/test/array-of-arrays.test.js
@@ -211,12 +211,11 @@ describe('Spectral rule: array-of-arrays', () => {
     it('Array of array of ints used within operation-level referenced parameter', async () => {
       const testDocument = makeCopy(rootDocument);
 
-      testDocument.components.parameters = {
-        ArrayParam: {
-          name: 'array_param',
-          in: 'header',
-          schema: arrayOfArrayOfInt
-        }
+      testDocument.components.parameters['ArrayParam'] = {
+        name: 'array_param',
+        description: 'the array parameter',
+        in: 'header',
+        schema: arrayOfArrayOfInt
       };
       testDocument.paths['/v1/drinks'].post.parameters = [
         {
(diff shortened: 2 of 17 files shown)


packages/ruleset/test/array-of-arrays.test.js
```diff
@@ -211,12 +211,11 @@ describe('Spectral rule: array-of-arrays', () => {
     it('Array of array of ints used within operation-level referenced parameter', async () => {
       const testDocument = makeCopy(rootDocument);
 
-      testDocument.components.parameters = {
-        ArrayParam: {
-          name: 'array_param',
-          in: 'header',
-          schema: arrayOfArrayOfInt
-        }
+      testDocument.components.parameters['ArrayParam'] = {
```
padamstx, line 214: Needed to be a bit more surgical since the root document now has other named parameters defined.

packages/ruleset/test/array-of-arrays.test.js
```diff
@@ -243,15 +242,13 @@ describe('Spectral rule: array-of-arrays', () => {
...
-      testDocument.components.parameters = {
-        ArrayParam: {
-          name: 'array_param',
-          in: 'header',
-          schema: {
-            type: 'array',
-            items: {
-              $ref: '#/components/schemas/ArrayOfString'
-            }
+      testDocument.components.parameters['ArrayParam'] = {
```
padamstx, line 245: ditto

packages/ruleset/test/description-mentions-json.test.js
```diff
@@ -50,7 +50,7 @@ describe('Spectral rule: description-mentions-json', () => {
         'This is a JSON object.';
 
       const results = await testRule(ruleId, rule, testDocument);
-      expect(results).toHaveLength(2);
+      expect(results).toHaveLength(3);
```
padamstx, line 53: now there's an additional occurrence of this run due to the new operation in the root document

packages/ruleset/test/parameter-description.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+      delete testDocument.components.parameters['VerboseParam'].description;
+
+      const results = await testRule(ruleId, rule, testDocument);
+      expect(results).toHaveLength(1);
+      expect(results[0].code).toBe(ruleId);
+      expect(results[0].message).toBe(expectedMsg);
+      expect(results[0].severity).toBe(expectedSeverity);
+      expect(results[0].path.join('.')).toBe(
+        'paths./v1/drinks/{drink_id}.get.parameters.0'
```
padamstx, line 42: this seems to "look better" than comparing actual arrays

packages/ruleset/test/utils/root-document.js
```diff
@@ -60,6 +60,56 @@ module.exports = {
         }
       }
     },
+    '/v1/drinks/{drink_id}': {
```
padamstx, line 63: new operation with a path param

packages/ruleset/src/functions/parameter-description.js
```diff
@@ -0,0 +1,26 @@
+module.exports = function(param, _opts, { path }) {
+  return parameterDescription(param, path);
+};
+
+const errorMsg = 'Parameter should have a non-empty description';
+
+function parameterDescription(param, path) {
+  const results = [];
```
dpopp07, line 8: I was reading through the Spectral custom function guide the other day and noticed they recommend against defining too many variables for performance reasons, since these functions may be called lots and lots of times.

For many of our functions, this is fairly unavoidable and it's fine. For this, since we aren't gathering multiple errors per function call, we could probably optimize a bit by returning an inline-defined array if the condition is met and returning an empty array at the end of the function.

This is just a suggestion though, not necessary. I'll leave it up to you

dpopp07 (approved):
Looks great! One suggestion but good to merge as is

packages/ruleset/src/functions/parameter-description.js
padamstx: done

PR #389, opened 2022-02-28
feat(parameter-case-convention): add new 'parameter-case-convention' rule and remove old rule

padamstx:
This commit converts the old 'param_name_case_convention' rule
 into a new Spectral-style rule 'parameter-case-convention'.

### DIFF
--- a/packages/ruleset/src/functions/parameter-case-convention.js
+++ b/packages/ruleset/src/functions/parameter-case-convention.js
@@ -1,0 +1,86 @@
+const { casing } = require('@stoplight/spectral-functions');
+const { isSdkExcluded, isDeprecated } = require('../utils');
+
+// Error message prefix for each parameter type.
+const errorMsgPrefix = {
+  query: 'Query parameter names ',
+  path: 'Path parameter names ',
+  header: 'Header parameter names '
+};
+
+const errorMsgNoName = 'Parameters must have a name';
+const errorMsgNoIn = "Parameters must have a valid 'in' value";
+
+module.exports = function(param, options, { path }) {
+  return parameterCaseConvention(param, path, options);
+};
+
+/**
+ * This function will check 'param' to make sure that its name
+ * follows the appropriate case convention, depending on its parameter type.
+ * @param {*} param the parameter object to check
+ * @param {*} path the location of 'param' within the OpenAPI document
+ * @param {*} casingConfig this is the value of the 'functionOptions' field
+ * within this rule's definition (see src/rules/parameter-case-convention.js).
+ * This should be an object with an entry (key) for each parameter type (i.e. 'in' value).
+ * The value of each entry should be an object which is the config to be
+ * passed to Spectral's casing() function to enforce case conventions for that parameter type.
+ * @returns an array containing zero or more error objects
+ */
+function parameterCaseConvention(param, path, casingConfig) {
+  // Don't bother enforcing the rule on excluded or deprecated parameters.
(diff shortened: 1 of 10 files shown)


packages/ruleset/src/utils/is-deprecated.js
```diff
@@ -0,0 +1,11 @@
+const isObject = require('./is-object');
+
+/**
+ * Returns `true` if the input is an object with x-sdk-exclude
+ * extension set to true.
```
pyrooka, line 5: It seems you forgot to reword this comment.

pyrooka (approved):
Looks good!

packages/ruleset/src/utils/is-deprecated.js
padamstx: done

padamstx:
The parameter-case-convention rule was updated such that it's default configuration is now defined within the rule definition.  This means that a user can can now override the default configuration by defining their own rule with a different config.

packages/ruleset/src/functions/parameter-case-convention.js
```diff
@@ -0,0 +1,84 @@
...
+ * @param {*} path the location of 'param' within the OpenAPI document
+ * @param {*} casingConfig this is the value of the 'functionOptions' field
+ * within this rule's definition (see src/rules/parameter-case-convention.js).
+ * This should be an object with an entry (key) for each parameter type (i.e. 'in' value).
+ * The value of each entry should be an object which is the config to be
+ * passed to Spectral's casing() function to enforce case conventions for that parameter type.
+ * @returns an array containing zero or more error objects
+ */
+function parameterCaseConvention(param, path, casingConfig) {
+  const errors = [];
```
dpopp07, line 31: Like you've done for other rules, might be a good idea to defer this array creation until you've check for exclusion, etc. Probably doesn't make a big difference for this one since most parameters won't be excluded but thought I would mention it

packages/ruleset/src/functions/parameter-case-convention.js
```diff
@@ -0,0 +1,84 @@
+const { casing } = require('@stoplight/spectral-functions');
+const { isSdkExcluded, isDeprecated } = require('../utils');
+
+// Error message prefix for each parameter type.
+const errorMsgPrefix = {
+  query: 'Query parameter names ',
+  path: 'Path parameter names ',
+  header: 'Header parameter names '
+};
```
dpopp07, line 9: The [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#parameter-object) also supports a parameter type of "cookie" - I'm guessing we don't have any prescriptions about this type but I wanted to make sure 🍪

dpopp07 (approved):
Looks good! 👍

packages/ruleset/src/functions/parameter-case-convention.js
padamstx: done

packages/ruleset/src/functions/parameter-case-convention.js
padamstx: Correct, we were actually only checking case conventions for query and path params in the old rule, so I added checking for header params in the new rule.
In the SDK generator, we don't do anything with cookie-type params so I figured I'd steer clear of those in the validator.
I actually haven't seen an API yet for an IBM Cloud service that uses cookie params anyway :)

PR #391, opened 2022-03-03
feat(pagination-style): add new 'pagination-style' rule and remove old rule

padamstx:
This commit converts the old 'pagination_style' rule
into a new Spectral-style rule 'pagination-style'.


padamstx:
This PR has been rebased on top of the staging branch and should be a little easier to review now :)

pyrooka (approved):
This is massive. LGTM!

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,363 @@
...
+    debug('Response schema has no array property');
+    return [];
+  }
+
+  // Grab the operation's parameters list for each access below.
+  const params = operation.parameters;
+
+  // Check to see if the operation defines a "start" query param.
+  // This could have any of the names below.
+  const startParamNames = ['start', 'token', 'cursor', 'page_token'];
```
hudlow, line 94: I've seen "page" before too.

But I am not sure why we're being so liberal about what we'll tolerate. I agree all these should be checked as a part of the heuristic for finding a paginated collection, but I think all but `start` should be _at least_ a warning.

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,363 @@
...
+  // The actual path string (e.g. '/v1/resources') will be the last element in 'path'.
+  const pathStr = path[path.length - 1];
+
+  // Retrieve this path item's 'get' operation.
+  const operation = pathItem.get;
+
+  // We'll bail out now if any of the following are true:
+  // 1. If the path string ends with a path param reference (e.g. '{resource_id}'
+  // 2. If the path item doesn't have a 'get' operation.
+  // 3. The 'get' operation is excluded via the 'x-sdk-exclude' extension.
```
hudlow, line 47: I'm not sure I see justification for skipping based on SDK exclusion.

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,363 @@
...
+      message:
+        'A paginated list operation must include an array property whose name matches the final segment of the path',
+      path: responseSchemaPath
+    });
+  }
+
+  // Check #8: If the response body contains a "total_count" property, it must be type integer and required.
+  // References:
+  // - https://cloud.ibm.com/docs/api-handbook?topic=api-handbook-pagination#additional-response-fields-with-token
+  // - https://cloud.ibm.com/docs/api-handbook?topic=api-handbook-pagination#additional-response-fields-with-offset
```
hudlow, line 291: Maybe a future issue, but should we look for nonconforming variations such as `total` or `count`?

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  // Checks 9-12 below will verify that the response schema contains the
+  // "first", "last", "previous", and "next" properties which are supposed to
+  // provide links (in the form of a request url string) to the first/last/previous/next result pages.
+  // Each of these properties should be an object that contains an "href" string property.
+  // Reference: https://cloud.ibm.com/docs/api-handbook?topic=api-handbook-pagination#pagination-links
+
+  // Check #9: The response body should contain a "first" property which links to the first page of results.
+  results.push(...checkPageLink(responseSchemaPath, responseSchema, 'first'));
```
hudlow, line 314: Should we check that `first` is required?

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,363 @@
...
+    debug('Results: ' + JSON.stringify(results, null, 2));
+  } else {
+    debug('PASSED');
+  }
+
+  return results;
+}
+
+// Verify that the '<name>' property exists within 'responseSchema'
+// and represents a valid "page link" value (an object with an "href" field).
```
hudlow, line 337: Should we also check for the page token to be reflected? (Maybe a future issue.)

padamstx: My understanding from the API handbook is that the page token query param should be returned by the server as part of the URL string within the "href" field.   If that's the case, then I don't know what we could possibly do in the validator since we are not validating the actual responses coming back from the server (we're validating the schema within the API definition).

packages/ruleset/src/functions/pagination-style.js
padamstx: Ok, I can add "page" to the list of possible names to look for.
And I can add an additional check (that is performed once we determine the operation is a paginated list operation) to warn if the param is not named "start".

packages/ruleset/src/functions/pagination-style.js
padamstx: My thinking was that since this new rule is a bit stricter than the old rule, users are likely to see new warnings crop up since very few APIs actually follow the letter of the law wrt pagination.  With that in mind, it might be a little overbearing if we start returning a bunch of new warnings for an operation that is excluded from SDK generation.

packages/ruleset/src/functions/pagination-style.js
padamstx: I was trying to implement the guidance in the handbook as closely as I could and it doesn't mention anything about "total" or "count" :)

packages/ruleset/src/functions/pagination-style.js
padamstx: In the section that talks about page links (https://cloud.ibm.com/docs/api-handbook?topic=api-handbook-pagination#pagination-links), I read that text from the perspective of what the service should/must/may do in terms of returning the values in the response body.   And that led me to conclude that from a JSON schema standpoint, none of those 4 properties would actually be a required schema property.

packages/ruleset/src/functions/pagination-style.js
padamstx: done

packages/ruleset/src/functions/pagination-style.js
```diff
@@ -0,0 +1,383 @@
...
+  if (
+    !Object.values(responseSchema.properties).some(
+      prop => prop.type === 'array'
+    )
+  ) {
+    debug('Response schema has no array property');
+    return [];
+  }
+
+  // Grab the operation's parameters list for each access below.
```
dpopp07, line 89: Is this meant to say "easy" ? I guess it kind of works either way haha, just checking what your intention was

dpopp07 (approved):
Looks good! 👍

packages/ruleset/src/functions/pagination-style.js
padamstx: yes, it's fixed

PR #392, opened 2022-03-03
feat(property-description): add new 'property-description' rule

padamstx:
A recent change was made that converted the old
'no_property_description' and 'no_schema_description' rules
into a single new 'schema-description' Spectral-style rule.
This commit introduces a new 'property-description' rule
and removes the property-related checks from the
'schema-description' rule.
So now, 'schema-description' will specifically ensure that
schemas have a non-empty description, while the new
'property-description' rule will ensure that
schema properties have a non-empty description.
This change is being made so that each rule can be
enabled/disabled separately according to user requirements.

### DIFF
--- a/packages/ruleset/src/functions/schema-description.js
+++ b/packages/ruleset/src/functions/schema-description.js
@@ -8,56 +8,33 @@ module.exports = function(schema, _opts, { path }) {
   return validateSubschemas(schema, path, schemaDescription);
 };
 
-const errorMsgSchema = 'Schema should have a non-empty description';
-const errorMsgProp = 'Schema property should have a non-empty description';
-
 function schemaDescription(schema, path) {
-  const results = [];
-
-  // This rule is configured with "resolved=true", so we shouldn't be called
-  // with a "ref" schema.
-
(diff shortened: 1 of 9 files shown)


packages/ruleset/test/schema-description.test.js
```diff
@@ -4,15 +4,13 @@ const { makeCopy, rootDocument, testRule, severityCodes } = require('./utils');
 const rule = schemaDescription;
```
padamstx, line 4: The main changes in this file are to remove the property-related tests (moved to the new testcase) and to simplify the way we verify that the correct path is reported in various testcases.

packages/ruleset/src/functions/schema-description.js
```diff
@@ -8,51 +8,28 @@ module.exports = function(schema, _opts, { path }) {
   return validateSubschemas(schema, path, schemaDescription);
```
padamstx, line 8: Mainly just removed the property-related checks in this file.

dpopp07 (approved):
Looks good! 👍

PR #408, opened 2022-04-05
feat: add new `property-inconsistent-name-and-type` rule and remove old rule

dpopp07:
## PR summary

This removes the old `inconsistent_property_type` rule in favor of a new, Spectral
rule called `property-inconsistent-name-and-type`. This performs the same function -
it validates that all properties in an API definition of the same name use a consistent
type. This also updates the error message from what it was before so that it is more
clear.

Implementing this rule in the Spectral format required a little bit of funny business. We need to perform the check in the context of the entire API definition but Spectral usually provides one schema to the function at a time, based on the `given` field. One solution to this issue is to ask Spectral for the whole spec (`$`) but then we would have to write the logic to find all of the schemas ourselves, which I really wanted to avoid. The other solution was to define an object at the rule level and pass it the function as an option. Since it is defined outside of the function and is passed by reference, the object will collect all of the properties as the function processes the different OpenAPI artifacts. I don't believe this should cause any weirdness during normal execution but it is slightly unconventional. It does have some minor implications in the tests, which I indicated in the comments and it would have implications if the user wanted to extend this rule with their own functionality (so I didn't document it as specifically extensible). Let me know if y'all have any additional thoughts or considerations.

### General checklist

- [x] Removed old rule implementation (packages/validator/src/plugins/validation/*.js)
- [x] Removed or adjusted testcases (packages/validator/test)
- [x] Updated default configuration to deprecate old rule (packages/validator/src/.defaultsForValidator.js)
- [x] Removed docs of old rule (docs/ibm-legacy-rules.md)


packages/ruleset/test/property-inconsistent-name-and-type.test.js
```diff
@@ -0,0 +1,394 @@
...
+      expect(validation.message).toBe(
+        'Properties with the same name have inconsistent types: running_time.'
+      );
+      expect(validation.path.join('.')).toBe(
+        'paths./v1/books.post.requestBody.content.application/json.schema.properties.running_time'
+      );
+      expect(validation.severity).toBe(expectedSeverity);
+    });
+
+    // due to the nature of the rule initializing the "visitedProperties" object one time, this rule MUST
```
padamstx, line 309:
```suggestion
    // due to the nature of the rule initializing the "visitedProperties" object one time, this test MUST
```
typo?

packages/ruleset/test/property-inconsistent-name-and-type.test.js
```diff
@@ -0,0 +1,394 @@
...
+    // use a different property to flag than the previous test, or the first instance won't be reported
```
padamstx, line 310: Could we avoid this problem if we explicitly set the rule's "visitedProperties" field to {} at the beginning of this test?

padamstx (commented):
I have a few questions/comments to address before approving.

packages/ruleset/src/functions/property-inconsistent-name-and-type.js
```diff
@@ -0,0 +1,54 @@
...
+function propertyInconsistentNameAndType(schema, path) {
+  if (schema.properties) {
+    const errors = [];
+
+    for (const [propName, prop] of Object.entries(schema.properties)) {
+      // Skip check for deprecated properties.
+      if (prop.deprecated === true) continue;
+
+      if (visitedProperties[propName]) {
+        if (visitedProperties[propName].type !== prop.type) {
```
hudlow, line 21: `prop.type` is going to miss type for composed schemas, including a simple `allOf` reference, right? But if this is how our current rule behaves I suggest we leave it that way and open a separate issue to cover the more sophisticated behavior. (Especially since I have some helpful utilities for this in flight.)

packages/ruleset/src/functions/property-inconsistent-name-and-type.js
```diff
@@ -0,0 +1,54 @@
...
+  return validateSubschemas(schema, path, propertyInconsistentNameAndType);
+};
+
+function propertyInconsistentNameAndType(schema, path) {
+  if (schema.properties) {
+    const errors = [];
+
+    for (const [propName, prop] of Object.entries(schema.properties)) {
+      // Skip check for deprecated properties.
+      if (prop.deprecated === true) continue;
```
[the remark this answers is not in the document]
padamstx: Upon further review, we do not want to add the isSdkExcluded() check here.

packages/ruleset/src/functions/property-inconsistent-name-and-type.js
```diff
@@ -0,0 +1,54 @@
+const { validateSubschemas } = require('../utils');
+
+let visitedProperties;
+let excludedProperties;
+
+module.exports = function(schema, options, { path }) {
+  visitedProperties = options.visitedProperties;
```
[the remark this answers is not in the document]
dpopp07: Hmm I don't think so but it's worth checking. I think the important thing is that the object is _defined_ in the rule file. This way, `visitedProperties` in this context is effectively just a pointer to the one defined in the rule file.

This gives me another idea though, something that I didn't think would work before but actually might, now that I'm thinking about it again. I'll follow up here with what I find

packages/ruleset/test/property-inconsistent-name-and-type.test.js
dpopp07: I don't actually think we can do that but this made me realize we might be able to reset the function options within the test, I'll try that.

packages/ruleset/src/functions/property-inconsistent-name-and-type.js
dpopp07: That's a good point. This is how the existing rule works but now you're making me realize that perhaps we shouldn't go about this rule this way. Instead of looking at the `properties` list, we could examine each schema as a whole. The `validateSubSchemas` function should give us all of the schemas in the API definition - we could just look to see what the type of each is and compare it to the other schemas we've seen.

Does this make sense? Would that address the issue you mentioned?

packages/ruleset/src/functions/property-inconsistent-name-and-type.js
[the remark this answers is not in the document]
dpopp07: I was wrong before - I was able to move initialization of the object to the function file, which I believe results in a much cleaner rule.

packages/ruleset/test/property-inconsistent-name-and-type.test.js
dpopp07: I learned that we can reset the loaded modules using `jest` and so I've now configured it to do that after each test. This is still slightly unusual-looking but it isolates the workaround to a small snippet of code in the test file only. I think this is the best solution.

padamstx (approved):
LGTM

PR #434, opened 2022-04-21
fix: fix rule deprecation support

dpopp07:
By removing categories from the .defaultsForValidator.js, the deprecated rule logic
is prevented from finding the old rules to map to new ones. This commit adds the
categories back in so that users have an easier time converting their configurations
to the new format.

I also removed the `init` and `migrate` commands because they are largely irrelevant
at this point and we don't want anyone using them.


padamstx (approved):
LGTM

PR #440, opened 2022-04-29
refactor: simplify ruleset index files

hudlow:
With apologies for the conflicts, but is there anything wrong with this simplification?


dpopp07 (approved):
I was under the impression using require statements like that wasn't a great practice but I can find nothing to support that idea, nor can I personally think of any reason why that would be the case. Probably something I conflated with something else a long time ago. It definitely makes these files easier to update so if it's okay with the linter, it's definitely okay with me!

padamstx (approved):
LGTM, just had a question about a possible optimization

PR #448, opened 2022-05-27
feat(schema-type): add new 'schema-type' rule

### LINKED ISSUE #444: Provide rule to check that all schemas have a `type`
With the understanding that some APIs make use of schemas that have no type (i.e., this might default to being optional), our API expects all schemas to have a type, but this is only policed manually at present.  It'd be very helpful to have this enforced by the validator.

dpopp07:
## PR summary

This rule validates that a schema has a non-empty `type` field.

Resolves #444 

### General checklist


padamstx:
It's actually valid to not specify the type field for a schema (that's an "any type").   Plus, if a schema has no type field, but specifies "properties", then we have always interpretted that as type=object in the generator (actually down in the swagger-parser I think).
So, I'm not sure we actually want to introduce this rule unless it's disabled by default and individual users can choose to enable it.

meem:
@padamstx Indeed, I commented in #444 that I'd be fine with requiring this rule to be explicitly enabled.  At least in the VPC API, we do not make use of "any" types, and do not want to rely on consumer-specific heuristics such as the presence of `properties` inferring an `object` type.

padamstx (approved):
LGTM, but we'll need to decide if we're going to make the rule disabled by default or not.  If we do that, we should make sure it is documented as such in the reference docs.

padamstx (approved):
LGTM

PR #460, opened 2022-07-07
fix: remove update-notifier from package

dpopp07:
The `update-notifier` package in its latest version is ESM-only and
is not compatible with the validator in its current state. Additionally,
previous versions of the package contain vulnerabilities.

If we migrate to ESM in the future, we may add the package back in but
for now, we are removing it completely (we had some other trouble with
it as well).


padamstx (approved):
LGTM

PR #471, opened 2022-07-27
refactor: move $ref check out of individual functions

dpopp07:
When invoking a function on a rule with the resolved spec, it is prudent to
check for schemas defined by a $ref. In the resolved spec, schemas defined with
a $ref are the result of circular references being detected and it isn't
necessary to validate the schemas.

To avoid duplicating this check for each function and requiring manual
maintenance in all functions (many were missing the check), I moved it to the
validate functions that they all rely on to recursively check schemas.

The only potential downside here is that it would prevent us from writing a
function that recursively walks through the unresolved spec and that also
relies on the presence of $refs in the schemas. This feels extremely unlikely
to me so it seems a worthwhile optimization.

Happy to be corrected if I'm making a bad assumption here!

### DIFF
--- a/packages/ruleset/src/functions/array-boundary.js
+++ b/packages/ruleset/src/functions/array-boundary.js
@@ -17,12 +17,6 @@ function debug(msg) {
 }
 
 function arrayBoundaryErrors(schema, path) {
-  // If "schema" is a $ref, that means it didn't get resolved
-  // properly (perhaps due to a circular ref), so just ignore it.
-  if (schema.$ref) {
-    return [];
-  }
-
   if (getSchemaType(schema) !== SchemaType.ARRAY) {
     return [];
   }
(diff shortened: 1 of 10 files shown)


hudlow (commented):
Looks good, but should we add a note to the headers of the two (three?) functions making this assumption like "Warning: this is only safe to use with resolved schemas."?

dpopp07:
> Looks good, but should we add a note to the headers of the two (three?) functions making this assumption like "Warning: this is only safe to use with resolved schemas."?

Good call - I left a note on that where I added the check but it makes much more sense in the context of the header and should also be included with `validateSubschemas`.

dpopp07:
@hudlow Updates made to the headers

padamstx (approved):
LGTM

PR #472, opened 2022-07-27
fix: stop considering x-sdk-excluded annotation in validations

dpopp07:
Back when the validator was primarily intended to prepare OpenAPI definitions
for SDK generation, we decided to ignore any artifacts of the definition marked
by the `x-sdk-exclude` annotation. Now that the validator is intended to assist
with API governance more broadly, we want to enforce high quality API definitions,
even when not being used for SDK generation. This commit removes all checks for
SDK exclusion in the validations so that all aspects of the API definition are
considered for all rules.


padamstx (approved):
LGTM

PR #473, opened 2022-08-04
fix(string-boundary): improve handling of composed schemas

padamstx:
## PR summary
This commit modifies the "string-boundary" rule so that it
more accurately performs the checks for the various schema
attributes within composed schemas.
For example, if a string schema contains an allOf, then the
"pattern" field could be specified in any of the allOf
list element schemas instead of the main schema itself.
In contrast, for a string schema containing a oneOf or anyOf,
the "pattern" field would need to be specified in either
the main schema itself OR within EACH of the oneOf/anyOf
list element schemas.

Previously, the rule was not accurate enough with its
checks within composed schemas.

Signed-off-by: Phil Adams <phil_adams@us.ibm.com>

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/string-boundary.js
+++ b/packages/ruleset/src/functions/string-boundary.js
@@ -1,7 +1,12 @@
-const { validateSubschemas } = require('../utils');
+const {
+  checkCompositeSchemaForConstraint,
+  getCompositeSchemaAttribute,
+  isStringSchema,
+  validateNestedSchemas
+} = require('../utils');
 
 module.exports = function(schema, _opts, { path }) {
-  return validateSubschemas(schema, path, stringBoundaryErrors, true, false);
+  return validateNestedSchemas(schema, path, stringBoundaryErrors, true, false);
 };
 
 // Rudimentary debug logging that is useful in debugging this rule.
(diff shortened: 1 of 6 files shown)


packages/ruleset/src/functions/string-boundary.js
```diff
@@ -20,68 +25,74 @@ const bypassFormats = {
...
 };
 
+const compositionFields = ['allOf', 'anyOf', 'oneOf'];
+
 function stringBoundaryErrors(schema, path) {
+  // We don't want to flag a violation for a (potentially) partial schema
+  // defined in an allOf/anyOf/oneOf list element.
+  if (path.length >= 2 && compositionFields.includes(path[path.length - 2])) {
+    return [];
+  }
```
dpopp07, line 35: I *think* you can handle this by just using `validateNestedSchemas` instead of `validateSubschemas`

packages/ruleset/src/utils/get-composite-schema-attribute.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+function getCompositeSchemaAttribute(schema, attrName) {
+  let value = undefined;
+  const foundConstraint = checkCompositeSchemaForConstraint(schema, s => {
+    if (attrName in s && s[attrName] !== undefined && s[attrName] !== null) {
+      value = s[attrName];
+      return true;
+    }
+  });
+  return foundConstraint ? value : undefined;
```
dpopp07, line 22: Seems this is probably equivalent 🙂 

```suggestion
  return value;
```

But the code is definitely good and clear as is

dpopp07 (approved):
Looks good! I think we should use the `validateNestedSchemas` utility function for this rule, as its purpose is to validate schemas without validating individual composed schema children, but good to go otherwise

packages/ruleset/src/utils/get-composite-schema-attribute.js
padamstx: I initially thought so too, but that is actually not correct because it's possible that we were able to save off the value of the attribute from an individual anyOf/oneOf list element if we happened to find it there, but yet we would still get back a false from checkComposedSchemaForConstraint() if for example the attribute was not present in all elements of an anyOf or oneOf list.   So in that scenario, we'd want to return undefined for the attribute value.

packages/ruleset/src/functions/string-boundary.js
padamstx: I'll change this...

packages/ruleset/src/utils/get-composite-schema-attribute.js
dpopp07: Ohh right, I didn't even think about that 🤦‍♂️

Thanks for the explanation!

packages/ruleset/src/functions/string-boundary.js
padamstx: I verified that using validateNestedSchemas works and I can remove that if-block.

PR #475, opened 2022-08-09
fix: allow relative urls in security schemes

### LINKED ISSUE #466: security-scheme-attributes does not account for relative URLs
Using the example from the [swagger docs](https://swagger.io/docs/specification/api-host-and-base-path/) for Relative URLs
```yaml
openapi: 3.0.0
servers:
  - url: https://api.example.com
  - url: https://sandbox-api.example.com
# Relative URLs to OAuth2 authorization and token URLs
components:
  securitySchemes:
    oauth2:
      type: oauth2
      flows:
        authorizationCode:
          authorizationUrl: /oauth/dialog
          tokenUrl: /oauth/token
```
the validator produces the error messages 
```
  Message :   security scheme 'authorizationUrl' property must be a valid URL
  Rule    :   security-scheme-attributes
  Path    :   components.securitySchemes.oauth2.flows.authorizationCode.authorizationUrl
  Line    :   12

  Message :   security scheme 'tokenUrl' property must be a valid URL
  Rule    :   security-scheme-attributes
  Path    :   components.securitySchemes.oauth2.flows.authorizationCode.tokenUrl
  Line    :   13
```

But I believe this should be valid because, according to the [OpenAPI 3 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#relativeReferences):

> Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2). Relative references are resolved using the URLs defined in the [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#serverObject) as a Base URI.

and in this example the `authorizationUrl` and `tokenUrl` prepended with the `servers.url` paths are valid URLs.

dpopp07:
The OpenAPI specification [allows for relative references](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#relativeReferences) in URL properties
throughout an API document. We were not considering this in the
`security-scheme-attributes` rule. This commit adjusts the logic to allow
for relative URLs.

Resolves #466

### DIFF
--- a/packages/ruleset/src/functions/security-scheme-attributes.js
+++ b/packages/ruleset/src/functions/security-scheme-attributes.js
@@ -1,7 +1,11 @@
 const stringValidator = require('validator');
 
-module.exports = function(securityScheme, _opts, { path }) {
-  return checkSecuritySchemeAttributes(securityScheme, path);
+module.exports = function(securityScheme, _opts, context) {
+  return checkSecuritySchemeAttributes(
+    securityScheme,
+    context.path,
+    context.document
+  );
 };
 
 const API_KEY = 'apiKey';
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

PR #476, opened 2022-08-09
fix(content-entry-provided): make exceptions for head, options, and trace operations

### LINKED ISSUE #235: no_response_body exclusion for OPTIONS requests
General convention for OPTIONS requests is to return an empty 200 response with Content-Length=0.

Doing some investigation, some builds of FireFox even flag preflight requests that return anything other than 200 as an error (not sure on the behavior for the current release).

The no_response_body rule should either ignore empty response bodies for OPTIONS requests or allow configuration to exclude them.

dpopp07:
We don't expect content entries for the operation types HEAD, OPTIONS, and TRACE and
we should loosen the `content-entry-provided` rule to make exceptions in those cases.

Resolves #235


padamstx (approved):
LGTM

hudlow (approved):

PR #493, opened 2022-09-13
feat(inline-request-schema): add new spectral-style rule

padamstx:
## PR summary
This commit introduces the new spectral-style
'inline-request-schema' validation rule which will check for requestBody objects that are defined with an inline object schema. A $ref to a named schema is preferred in those situations.

Signed-off-by: Phil Adams <phil_adams@us.ibm.com>

### General checklist


pyrooka (approved):
LGTM!

packages/ruleset/src/functions/inline-schema-rules.js
```diff
@@ -0,0 +1,116 @@
...
+  if (!schema.$ref && isJsonMimeType(mimeType)) {
+    if (isObjectSchema(schema)) {
+      // If "schema" is an inline object schema that will be refactored by the SDK generator, then return a warning.
+      //
+      // TODO - We need to decide if we want to consider a requestBody schema's "explosiveness" when deciding
+      // whether to return a warning or not.
+      // Technically, for an inline requestBody schema that will have its body vars exploded we probably don't need
+      // to log a warning since we don't actually need the requestBody schema to be a $ref in that case.
+      // Currently, this function does not consider this when logging a warning, so any inline object schema
+      // found in a requestBody will cause a warning.
```
dpopp07, line 76: My opinion on this is that we _should_ consider explosiveness and avoid logging warnings where there is not an SDK generation benefit.

I would understand the other perspective too though, which I suppose would say that references are always a good idea for organization and keeping things DRY, etc.

dpopp07 (approved):
Looks good! Though, in response to your TODO, I wouldn't mind seeing "explosiveness" taken into account

hudlow (commented):
I'm afraid I have to disagree with Dustin—I think we should continue to assert that any unnamed object schema should, as a best practice, be named and moved to `components`.

@padamstx I feel like I'm missing a part of how this works, though maybe it's just the exhaustion of the day...can you point me to where the ref sibling pattern is detected? I only see the tests.

padamstx:
> @padamstx I feel like I'm missing a part of how this works, though maybe it's just the exhaustion of the day...can you point me to where the ref sibling pattern is detected? I only see the tests.

The rule doesn't actually need to do anything special with a ref-sibling pattern. Mainly because a ref-sibling composition couldn't (by definition) involve an inline object schema.

hudlow:
@padamstx Hmm, if I'm reading it correctly, this code is relying on the fact that `isObjectSchema()` returns `false` for ref-sibling pattern schemas that have not been resolved? If so, I'd regard that behavior is a bug—they're still object schemas in that case, even if the heuristic fails to detect that fact. (I wonder if we need to throw exceptions in utility functions that encounter a `$ref` when they don't expect one and won't reason correctly about it.)

To put a finer point on it, shouldn't these tests pass? https://github.com/IBM/openapi-validator/pull/495/files

To be honest, I'm really nervous about continuing down this path without defining the contract for the ref-sibling pattern outside of implementations.

dpopp07:
> I think we should continue to assert that any unnamed object schema should, as a best practice, be named and moved to `components`

I definitely resonate with that logic, so that's all good by me. I think in that case though, the docs file should be updated to include the point that defining schemas in components is a general best practice, rather than only stating it is for the benefit of SDK generation in this case.

padamstx:
> To put a finer point on it, shouldn't these tests pass? https://github.com/IBM/openapi-validator/pull/495/files

I've updated the PR with different logic for detecting an inline object schema.
I added your additional testcases and they now all pass.

PR #499, opened 2022-10-03
fix: fix broken pkg script and publish binaries to github

dpopp07:
The pkg script was broken because of some clever usage of "rename" that wasn't compatible with all devices. This commit removes the need for it by moving the logic into a separate bash script with way more clarity.

This also updates semantic release to include the built platform binaries in the "assets" section of the release on GitHub.


padamstx (approved):
LGTM!

PR #501, opened 2022-10-05
feat: add ruleset version string to `--version` output

### LINKED ISSUE #498: print RuleSet version
Hi,
would it be possible to print the version of the ruleset package included in the validator?
For instance, when I run 'lint-openapi --version', it would be useful to have something like this:
ValidatorVersion:"0.95.0", RuleSetVersion:"0.44.0"

The format is not important for me, just the information.

The version of the RuleSet package is useful to look at RuleSet's changelog to identify which rule has been modified in that particular version.

Thanks

dpopp07:
This commit adds the ruleset verion number to the `--version` output so it is easier for the user to link their current version to the updates in the changelog, etc.

This also makes a modification to allow for publishing of binaries to the GitHub release.

Example output:
```
$ node packages/validator/src/cli-validator/index.js --version
validator: 0.95.1; ruleset: 0.44.0
```

Resolves #498

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
 script:
 - npm run test-travis
 - npm run lint
-- npm run pkg
+- npm run pkg && mv bin packages/validator
 deploy:
 - provider: script
   skip_cleanup: true
--- a/packages/validator/src/cli-validator/index.js
+++ b/packages/validator/src/cli-validator/index.js
@@ -6,8 +6,8 @@
 require('./utils/check-version')('10.0.0');
 
 const program = require('./utils/modified-commander');
+const getVersionString = require('./utils/get-version-string');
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
 script:
 - npm run test-travis
 - npm run lint
-- npm run pkg
+- npm run pkg && mv bin packages/validator
```
dpopp07, line 12: I think that all that went wrong last time is that the `.releaserc` file is moved to the individual packages for release. So, the `bin` folder wasn't present for the validator package, which we want to receive the binaries. Hopefully this works 🤞

ibm-devx-sdk (approved):
LGTM

padamstx (approved):
LGTM

PR #503, opened 2022-10-11
test: Spectal assumption for resolved reference equivalence

hudlow:
## PR summary

Add a test for the assumption that when Spectral resolves a schema, two identical `$ref` references will result in the same (`===` equivalent) in-memory object.

### General checklist

### DIFF
--- a/packages/ruleset/test/meta/spectral-depedencies.test.js
+++ b/packages/ruleset/test/meta/spectral-depedencies.test.js
@@ -1,0 +1,49 @@
+const { testRule } = require('../utils');
+
+// Test cases for dependencies on subtle and possibly fragile Spectral behavior
+describe('Spectral behavior dependencies', () => {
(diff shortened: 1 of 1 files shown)


packages/ruleset/test/schema-resolution.test.js
```diff
@@ -0,0 +1,48 @@
+const { testRule } = require('./utils');
+
+describe('Spectral behavior assumptions', () => {
```
dpopp07, line 3: If we want to contain any Spectral behavior assumption tests to this suite, which might be a good idea, we could rename this file and use it for any future tests too

dpopp07 (approved):
Looks good! 👍

padamstx (approved):
LGTM
I agree with @dpopp07 that this new testcase could be a container for all the tests involving Spectral behavior assumptions.  We could either rename it now or when (if?) we need to add more tests.

dpopp07 (approved):
Looks good! 👍

PR #506, opened 2022-10-12
refactor: work-in-progress proposed refactor of composed schema restrictions

hudlow:
Currently one failing test and some missing tests for the new utility, but posting this PR for feedback on the approach.


packages/ruleset/src/functions/composed-schema-restrictions.js
```diff
@@ -1,123 +1,44 @@
...
-            };
-            errors.push(error);
-          }
-        } else {
-          // Property doesn't exist in main schema, so just add it.
-          targetProperties[name] = prop;
-        }
+  // Only object schemas have properties
+  if (isObjectSchema(schema)) {
+    // Collects all composed property schemas indexed by the name of the property they define
```
dpopp07, line 29: Would be easier to read (for me, at least) if this commented elaborated slightly more on the data structure returned by `getPropertySchemasByName`. Something to let the reader know the map values are lists of schemas all defining properties of common names.

dpopp07 (commented):
Okay after re-reading, my only feedback is to add a little detail to one of the comments for overall readability.

The other feedback I had turned out to be irrelevant on second look.

I still think it would be helpful to 1) validate that all schemas in a composed schema are objects and 2) add some additional test cases

padamstx:
> ...validate that all schemas in a composed schema are objects

Actually, if we encounter a primitive schema that also happens to contain a oneOf/anyOf (perhaps to define different sets of contstraints), we don't really care what is contained in the oneOf/anyOf list because we'll end up just refactoring the oneOf/anyOf list away during pre-processing.
If we're dealing with an object schema that contains a oneOf/anyOf, then its oneOf/anyOf list must contain only object schemas.

dpopp07:
> Actually, if we encounter a primitive schema that also happens to contain a oneOf/anyOf (perhaps to define different sets of contstraints), we don't really care what is contained in the oneOf/anyOf list because we'll end up just refactoring the oneOf/anyOf list away during pre-processing.
If we're dealing with an object schema that contains a oneOf/anyOf, then its oneOf/anyOf list must contain only object schemas.

Yes, I summarized around that nuance - thanks for clarifying!

PR #508, opened 2022-10-13
perf: remove accidental recursion in given statements

hudlow:
## PR summary

I'm wary of proclaiming that this definitely fixes our recent performance issues since @dpopp07 saw some really unpredictable behavior that seemed unrelated to this specific rule. But, I did profile the validator with `node --prof` and, in the resulting log, a fragment of this JSON path expression was being executed hundreds of times for an extremely simple API definition and millions of times for a complex definition.

It bothers me that I can't figure out exactly what a `.[*]` construct is **supposed** to do in JSON path, but in practice, it seems to result in some sort of recursion (maybe it's a synonym for `..`?). I'm guessing that the three instances meant it was `O(n^3)` or something.

In any case, the update didn't cause any test failures and did seem to resolve the performance problems I was able to replicate. Hoping it works for others as well.

### General checklist

### DIFF
--- a/packages/ruleset/src/rules/content-entry-contains-schema.js
+++ b/packages/ruleset/src/rules/content-entry-contains-schema.js
@@ -5,8 +5,8 @@ module.exports = {
   description: 'Content entries must specify a schema',
   formats: [oas3],
   given: [
-    '$.paths[*].[post,put,patch].requestBody.content[*]',
-    '$.paths[*].[get,post,put,patch,delete].[parameters,responses][*].content[*]'
+    '$.paths[*][post,put,patch].requestBody.content[*]',
+    '$.paths[*][get,post,put,patch,delete][parameters,responses][*].content[*]'
   ],
   severity: 'warn',
   resolved: true,
--- a/packages/ruleset/src/rules/optional-request-body.js
+++ b/packages/ruleset/src/rules/optional-request-body.js
@@ -6,7 +6,7 @@ module.exports = {
     'An optional requestBody with required properties should probably be required',
   message: '{{description}}',
   given:
-    "$.paths[*].[*].[?(@property === 'requestBody' && @.required !== true)].content.[*].schema",
+    "$.paths[*][*][?(@property === 'requestBody' && @.required !== true)].content[*].schema",
   severity: 'info',
   formats: [oas3],
   resolved: true,
--- a/packages/ruleset/src/rules/request-body-object.js
+++ b/packages/ruleset/src/rules/request-body-object.js
@@ -3,7 +3,7 @@ const { enumeration } = require('@stoplight/spectral-functions');
 module.exports = {
   description: 'All request bodies MUST be structured as an object',
   given:
-    '$.paths[*][*].requestBody.content.[?(@property ~= "^application\\\\/json(;.*)*$")].schema',
+    '$.paths[*][*].requestBody.content[?(@property ~= "^application\\\\/json(;.*)*$")].schema',
   severity: 'error',
   then: {
     field: 'type',
--- a/packages/ruleset/test/meta/rule-style.test.js
+++ b/packages/ruleset/test/meta/rule-style.test.js
@@ -1,0 +1,19 @@
+const rules = require('../../src/rules');
+
+// Test cases to enforce rule style
+describe.each(Object.entries(rules))(
+  'Rules conform to required style',
+  (name, rule) => {
+    it(`${name} rule does not contain implicit .[] recursion`, async () => {
+      const implicitRecursionRegex = /\.(?=\[)/;
+
+      if (Array.isArray(rule.given)) {
+        // `given` array
+        rule.given.map(g => expect(g).not.toMatch(implicitRecursionRegex));
+      } else {
+        // `given` string
+        expect(rule.given).not.toMatch(implicitRecursionRegex);
+      }
+    });
+  }
+);


dpopp07 (approved):
Looks good and at least on my machine, it solves the performance issue we've been seeing. My test case runs in ~3 seconds with this change as opposed to ~112 seconds without it. Pretty wild! Thanks for doing that investigation

pyrooka (approved):
Wow, that's a massive speed-up! Looks good.

padamstx (approved):
Changes look good.

We also have this same pattern ("].[") in the content-entry-contains-schema rule as well:
```
  given: [
    '$.paths[*].[post,put,patch].requestBody.content[*]',
    '$.paths[*].[get,post,put,patch,delete].[parameters,responses][*].content[*]'
  ],
```
Should that be changed too?  Is there a lesson to be learned from this going foward, such as "don't use [] next to ." or something?

hudlow:
@padamstx good catch. For whatever reason, I only checked for `.[*]` when I originally looked for other examples. I was thinking we might want to write a test that examines all rules and fails when it sees `.[` — is that an abuse of our testing system? I just feel like this subtlety could be easy to forget and I'd hate for it to result in another multi-week goose chase. FWIW, we may have been lucky in this case that I could trace the profiler output back to a particular line of code.

padamstx:
> I was thinking we might want to write a test that examines all rules and fails when it sees `.[` — is that an abuse of our testing system?
Any test that improves the quality of our software and our ability to prevent problems from being exposed to users is a good test in my book.
This would sort of be like a lint-check for our rule definitions.

padamstx:
I'm trying to understand more about any lessons learned here...

I think "." is typically used just as a path separator within a json path (e.g. `$.components.schemas[*]` would visit all named schemas in the API).  And "[]" is typically used to define a path with wildcards or multiple possible field names (e.g. `$.paths[*]` would visit all the "path item" objects in the API).   Is it just the case that using these two together is simply incorrect because it means something completely different then simply defining the path hierarchy?

One other thing to think about is the fact that spectral uses the "nimma" package for their json path engine (authored by one of the key spectral devs) and I wonder if the use of ".[" is simply not handled well by nimma or whether we are completely mis-using jsonpath.  In the end, it doesn't matter which is the case as we need to solve it either way, but would help inform us a little bit going forward perhaps.

hudlow:
@padamstx I went a bit further down the rabbit hole, but had to stop short of full comprehension. Here's what I learned:

1. I might be missing something, but the JSONPath RFC seems to forbid `.[` — you can see how [the dot selector ABNF](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base#section-3.4.2) requires a property name literal after the dot. That said, Nimma incorporates some nonstandard "JSONPath Plus" functionality, so it's no surprise it supports syntax that is not in the RFC.
2. Nimma also supports a "fallback" parser and Spectral configures [JSONPath Plus](https://github.com/JSONPath-Plus/JSONPath) itself as its fallback parser. This might explain some of the seemingly nondeterministic behavior we've seen — perhaps some of our environments resulted in the fallback being used and some didn't.
3. That said, both Nimma and JSONPath Plus seem to interpret `.[` within an expression as recursive descent. In the case of JSONPath Plus, this seems like it could be an accidental consequence of their expression [normalization logic](https://github.com/JSONPath-Plus/JSONPath/blob/c32644e635d793650ec2dcb8c934917ad0ca695c/src/jsonpath.js#L660) which converts both `.` and `[` to `;` before converting `;;` back to `..` (oof). Nimma seems far, far more rigorous and actually seems to deliberately interpret `.[` [as (recursive) descent](https://github.com/P0lip/nimma/blob/1c264ac8eebdc5669799d2658577f85d23781465/src/parser/parser.peg#L36) — refer to the [Peggy documentation](https://peggyjs.org/documentation.html#grammar-syntax-and-semantics-parsing-expression-types) for the syntax of this file and note that `&` seems to be a lookahead operator. Perhaps Nimma does this for compatibility with JSONPath Plus.

This does make me wish Nimma (and Spectral) supported a strict mode that threw an exception for any expression not fully compliant with the JSONPath RFC. I would consider this to be [virtuous intolerance](https://www.ietf.org/archive/id/draft-iab-protocol-maintenance-05.html#name-virtuous-intolerance) and it likely would have saved us a lot of time. (I am not sure how often we use nonstandard features of JSONPath Plus syntax, though.)

Unless someone else wants to dig deeper, or @P0lip is gracious enough to swoop in and enlighten us, this may be about as much closure as this bug is going to get.

P0lip:
> Perhaps Nimma does this for compatibility with JSONPath Plus.

Correct. I initially didn't support `.[` in the same way as JSONPath Plus, and it caused troubles for folks that already had existing JSON Path expressions that relied on that behavior.

That change landed relatively recently -> https://github.com/P0lip/nimma/commit/e7c98bf311721d1bbbfa0d91c3cb4f4e9d96f78c

> This does make me wish Nimma (and Spectral) supported a strict mode that threw an exception for any expression not fully compliant with the JSONPath RFC. I would consider this to be [virtuous intolerance](https://www.ietf.org/archive/id/draft-iab-protocol-maintenance-05.html#name-virtuous-intolerance) and it likely would have saved us a lot of time. (I am not sure how often we use nonstandard features of JSONPath Plus syntax, though.)

That's something to consider for sure.
Personally, I'd love to have such an option as well, since most folks aren't aware of the difference between `[` and `.[`. It looks quite subtle, while in practice it's the other way around.

> here. I guess the result of all this is "just don't use '.[' unless you know what it does" :)

💯

hudlow:
@P0lip Thanks so much for the info!

> That's something to consider for sure.
> Personally, I'd love to have such an option as well, since most folks aren't aware of the difference between `[` and `.[`. It looks quite subtle, while in practice it's the other way around.

Would you like me to open an issue on Nimma and/or Spectral? I wish I could say I had the bandwidth to take a stab at it, but that might not be realistic in the near future.

P0lip:
> Would you like me to open an issue on Nimma and/or Spectral? I wish I could say I had the bandwidth to take a stab at it, but that might not be realistic in the near future.

An issue in Spectral is fine. It's okay if you won't take a stab at it - it'd good to have that one in the Spectral issue tracker, so that we don't lose sight of it :)

PR #511, opened 2022-10-31
fix(collection-array-property): avoid bug that crashes nimma

dpopp07:
If a path object does not have a "get" operation, the code would previously pass `undefined` to a function that tried to access properties of the given operation object. This would result in an exception that would crash the underlying nimma parser.

This commit simply adds a check for truthyness before continuing with the function.


padamstx (approved):
LGTM!

PR #517, opened 2022-11-18
feat: prototype new -l/--loglevel option

padamstx:
## PR summary
This draft PR contains a prototype of a new feature that uses the "loglevel" package to provide a java.util.logging-like facility within the validator.
This prototype includes the following:
- A new -l/--loglevel command-line option that can be used to set the log level for one or more loggers (e.g. `-l root=info -l schema-description=debug` would set the root logger's level to 'info' and the `schema-description` logger's level to 'debug'.
- The `schema-description` and `property-description` rule functions were modified slightly to retrieve a logger named after their rule id, and use that logger to log a few debug messages.
- The overall idea is that there could be a "root" logger that could be used by the main portion of the validator for logging messages, and each rule would also have its own logger whose name is the same as the rule's rule id.  This will allow users (or developers of the validator) to turn on/off logging within specific rules or the main validator itself.

### General checklist


dpopp07:
I like the initial approach!

Some preliminary thoughts:
1. Might make sense to support environment variables for configuring log level as well. Environment variables can cause headaches but there definitely may be users that don't want to specify the log level each time they run the tool. Adding commands for storing persistent configuration may be another solution, though that comes with its own implications.
2. I don't want the API to be too complicated but it may be good to add support for defaulting to the root logger. e.g. if a user used `-l info`, the tool would know to set the root logger. Just trying to think about the simplest use case and make sure it's easy and intuitive for a user (not that my suggestion is more intuitive, I just want to consider if it would be)
3. Somewhat related - what if a user wanted to set one logging level for the _whole_ tool, including all rules? I suppose that would be a gross amount of output but a user wanted to do it, they shouldn't have to individually configure the level for each rule. I think it would be a valid use case for a level like "warning". We should think about how a user might do that.

packages/ruleset/src/functions/property-description.js
```diff
@@ -8,10 +9,15 @@ module.exports = function(schema, _opts, { path }) {
   return validateSubschemas(schema, path, propertyDescription);
 };
 
+const ruleId = 'property-description';
```
dpopp07, line 12: I wonder if this gets passed in somewhere (like the `_opts`). Not that it would make a huge difference but it could alleviate the burden of maintaining the rule name in two different places

padamstx:
>     2. I don't want the API to be too complicated but it may be good to add support for defaulting to the root logger. e.g. if a user used `-l info`, the tool would know to set the root logger. Just trying to think about the simplest use case and make sure it's easy and intuitive for a user (not that my suggestion is more intuitive, I just want to consider if it would be)
> 

This shouldn't be too hard to do.

>     3. Somewhat related - what if a user wanted to set one logging level for the _whole_ tool, including all rules? I suppose that would be a gross amount of output but a user wanted to do it, they shouldn't have to individually configure the level for each rule. I think it would be a valid use case for a level like "warning". We should think about how a user might do that.

Yeah, I've thought a little about the ability to use wild-cards to specify the logger names.  But I think any scheme that tries to support wild-cards would also require all the loggers to be known in advance. Without that, I think it will be difficult to support wild-cards since the loggers would normally be created only when the rule function is invoked the first time.  So we wouldn't know the set of loggers when we're processing command-line options.


need to make sure that all the loggers that might be needed know - up front - the names of all the loggers (essentially all the rule-id's plus 'root').  That way when we process the command-line args, we could rip through the loggers whose names match the wild-card and set the loglevel accordingly.  If we don't create all the loggers up front, then it would be difficult to support wild-cards since loggers whose names might match the wild-card won't be created until the rule function actually gets invoked.

packages/ruleset/src/functions/property-description.js
padamstx: From experimenting a little bit, I was able to find that the third argument passed to a rule function is essentially the entire "context" (this is where we retrieve the unresolved document when needed, etc.).  On this context object, there's a 'rule' field that contains info about the rule being invoked, and the 'name' field of that object contains the rule-id, as in this example:
```
  rule: Rule {
    name: 'schema-description',
    definition: {
      description: 'Schemas and schema properties should have a non-empty description',
      message: '{{error}}',
      given: [Array],
      severity: 'warn',
      formats: [Array],
      resolved: true,
      then: [Object]
    },
    owner: Ruleset {
      maybeDefinition: [Object],
      id: 1,
      formats: [Formats [Set]],
      definition: [Object],
      aliases: null,
      hasComplexAliases: false,
      extends: [Array],
      overrides: null,
      rules: [Object]
    },
    recommended: true,
    description: 'Schemas and schema properties should have a non-empty description',
    message: '{{error}}',
    documentationUrl: 'https://github.com/IBM/openapi-validator/blob/main/docs/ibm-cloud-rules.md#schema-description',
    resolved: true,
    formats: Formats(2) [Set] { [Function], [Function] }
  },

```
Notice that this Rule instance also has a field named 'documentationUrl' that has a very conveniently placed value that should point directly to its section in our reference docs :joy:.  I have no idea how that got there, but it's brilliant.  Weren't we just talking about something like this yesterday???

padamstx: Drats, I just double-checked that URL and it doesn't quite work because the tag value on the end should be prefixed with "rule-" (e.g. "rule-schema-description") :joy:

padamstx: Looks like we specify the URL in the ibm-oas.json file and spectral must be adding the `#<rule-id>` tag on the end.  I think that means we'll have to work out a way to avoid the "rule-" prefix on those tags inside our .md file.

pyrooka: I'm not sure if we can/want to rely on the filename or not, but if we can maybe something like this could work too:
```js
const ruleId = path.parse(__filename).name;
```

packages/validator/src/cli-validator/run-validator.js
```diff
@@ -34,6 +35,10 @@ const processInput = async function(program) {
     return Promise.reject(2);
   }
 
+  // Prototype:
+  // Set initial loglevel of the root logger to be 'info'.
+  log.setLevel('info');
```
pyrooka, line 40: Isn't "info" too verbose to use it as default? Maybe "warning" would be better.

pyrooka (commented):
Looks good as a prototype. Also, I agree with that 3 points that Dustin mentioned.

packages/validator/src/cli-validator/run-validator.js
padamstx: When we implement this feature for real, we would probably want to use the root logger wherever we're currently using `console.log` now, and I think we'd want the root logger to remain at the "info" log-level so we can in fact call logger.info() to log "informational" messages to the console (in addition to warning and error messages which would use logger.warn() and logger.error(), respectively).

pyrooka: Okay, that makes sense. Thanks!

padamstx:
I've updated the PR so that it now supports wild-cards (e.g. `-l *-description=debug`, `-l schema-*=debug`).  These expressions are really "globs" more than they are regular expressions (I used the minimatch package to do the matching).
One thing to note is that when you specify the -l options on the command line, those settings will apply to both of the following:
- any loggers that already existed BEFORE the command-line options were parsed (this is possible, depending on precisely when and how a rule function creates its logger)
- any loggers that are created after the command-line options were parsed

padamstx:
Closing this out because the logger feature has been merged into the 1.0-prerelease branch

PR #518, opened 2022-11-18
fix: remove buggy support for tolerating trailing commas

hudlow:
Signed-off-by: Dan Hudlow <dhudlow@us.ibm.com>
## PR summary
The feature to remove trailing commas (which are _not valid_ in JSON) mangled properties containing regular expressions such as `"pattern": "^[-0-9a-zA-Z_,]+$"` within an OpenAPI definition.

IMO, we should simply require valid JSON documents.

### General checklist

### DIFF
--- a/packages/validator/src/cli-validator/run-validator.js
+++ b/packages/validator/src/cli-validator/run-validator.js
@@ -168,16 +168,6 @@ const processInput = async function(program) {
 
       const fileExtension = ext.getFileExtension(validFile);
       if (fileExtension === 'json') {
-        // find and fix trailing commas
-        const match = originalFile.match(/,\s*[}\]]/m);
-        if (match) {
-          const chars = originalFile.substring(0, match.index);
-          const lineNum = (chars.match(/\n/g) || []).length + 1;
-          const msg = `Trailing comma on line ${lineNum} of file ${validFile}.`;
-          printError(chalk, chalk.red(msg));
-          exitCode = 1;
-          originalFile = originalFile.replace(/,(\s*[}\]])/gm, '$1');
-        }
         input = JSON.parse(originalFile);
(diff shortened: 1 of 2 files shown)


pyrooka (approved):
Looks good! I personally agree with the decision. I don't think it's the validator's responsibility to make the JSON valid, even in case if small issues like this.

packages/validator/src/cli-validator/run-validator.js
```diff
@@ -168,16 +168,6 @@ const processInput = async function(program) {
...
-        // find and fix trailing commas
-        const match = originalFile.match(/,\s*[}\]]/m);
-        if (match) {
-          const chars = originalFile.substring(0, match.index);
-          const lineNum = (chars.match(/\n/g) || []).length + 1;
-          const msg = `Trailing comma on line ${lineNum} of file ${validFile}.`;
-          printError(chalk, chalk.red(msg));
-          exitCode = 1;
-          originalFile = originalFile.replace(/,(\s*[}\]])/gm, '$1');
-        }
```
dpopp07, line 180: I agree that this is a good idea but I think this would be a good opportunity to utilize the logging feature that is in the works. IIRC, we added this code because users were struggling with low-detail error messages from the JSON parser that were time consuming to debug.

This would be a good candidate for a "debug" level log that says something like "JSON may have a trailing command" (intentionally vague because of the false positive you described). Perhaps only if the `JSON.parse` fails. Just something to give the user more insight into what their issue may be.

This certainly doesn't need to happen in this PR, I just wanted to capture it in a comment. I can open an internal issue linking this commit reminding us to re-add it once the logger is implemented. Or this PR can leave the console.log and nothing else.

What do you think?

dpopp07 (approved):
Looks good! Just left a comment about a logging consideration

PR #529, opened 2022-12-14
fix: move `validator` dependency to correct package within monorepo

### LINKED ISSUE #528: Validator dependency not declared
When running `lint-openai` with `yarn pnp`, you get the following error

```
Error: @ibm-cloud/openapi-ruleset tried to access validator, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

Looks like `openapi-ruleset/src/functions/security-scheme-attributes.js` imports `validator`, but the `validator` package isn't declared as a dependency of the `ruleset` package.

dpopp07:
The `validator` dependency is only used in the ruleset package but was still living in the `package.json` for the validator package. This was likely an oversight during the re-architecture into a monorepo. This resolves the issue by moving the dependency to the correct `package.json`.

Resolves #528

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12924,28 +12924,29 @@
     },
     "packages/ruleset": {
       "name": "@ibm-cloud/openapi-ruleset",
-      "version": "0.43.0",
+      "version": "0.45.0",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


padamstx (approved):
LGTM

PR #531, opened 2022-12-19
refactor: move select utility functions into new package

dpopp07:
Move out selected utilities and the "collections" module to a new folder, which the new NPM package is created from. The ruleset package now depends on this new utilities package.

This commit updates the ruleset package with the modified code to depend on the new package and it sets up the new package.


padamstx:
One question I have is... when you merge the PR into main and the resulting Travis build runs semantic-release, how will semrel know what the baseline should be for the new utilities package?  i.e. will you need to add a tag to the current main branch that essentially reflects version 0 of the new package?

dpopp07:
> will you need to add a tag to the current main branch that essentially reflects version 0 of the new package?

Yep! I manually published version 0.0.1 on NPM and I will create a new git tag for this commit to reflect that

PR #560, opened 2023-03-03
build: add pre-release config for semantic release

dpopp07:
Adding a "branch" to our semantic release configuration will allow for automated releases off of our pre-release branch, so that we can start publishing release candidates with each new change that we make.


padamstx (approved):
LGTM

PR #561, opened 2023-03-03
refactor: re-structure results format

dpopp07:
This commit moves towards relying on Spectral to give us all result data, including line numbers, so that we can soon remove our dependency on the AST algorithm to compute them. Additionally, the results object structure has been re-thought and changed in the code. It is based on and documented with a formal JSON schema. This structure is used as the official format for JSON output and is how the results are stored within the code.

BREAKING CHANGE: The structure of the JSON output is now different. See the documented JSON Schema for the new structure.


padamstx (approved):
LGTM

PR #562, opened 2023-03-03
build: use semantic release to control conditional deployment

dpopp07:
Right now, the Travis file dictates that the deploy step will only be run for the `main` branch. However, the only command we run during the deploy step is to invoke semantic release. Semantic release also dictates that it will only run on certain branches, so it should be sufficient for controlling the conditional deployment.

We want to deploy on multiple branches in order to publish release candidates from our pre-release branch, so this commit enables Travis to run the deploy step on each branch and lets semantic release control which branches are actually deployed from.


padamstx (approved):
LGTM

PR #565, opened 2023-03-06
refactor: optimize logic for summary-only feature

dpopp07:
Currently, we loop through all validations and check if "summary only" is enabled before we print them. It would be more efficient and readable to check once at the beginning, printing the summary and exiting if the feature is enabled.

This commit also includes some other, minor refactoring.

### DIFF
--- a/packages/validator/src/cli-validator/run-validator.js
+++ b/packages/validator/src/cli-validator/run-validator.js
@@ -266,13 +266,9 @@ async function runValidator(cliArgs, parseOptions = {}) {
       const numWarnings = results.warning.summary.total;
       if (numWarnings > limitsObject.warnings) {
         exitCode = 1;
-        // add the exceeded warnings limit as an error
-        results.error.results.push({
-          line: 0,
-          rule: 'warnings-limit',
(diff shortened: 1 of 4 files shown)


packages/validator/src/cli-validator/run-validator.js
```diff
@@ -266,13 +266,9 @@ async function runValidator(cliArgs, parseOptions = {}) {
       const numWarnings = results.warning.summary.total;
       if (numWarnings > limitsObject.warnings) {
         exitCode = 1;
-        // add the exceeded warnings limit as an error
-        results.error.results.push({
-          line: 0,
-          rule: 'warnings-limit',
-          path: [],
-          message: `Number of warnings (${numWarnings}) exceeds warnings limit (${limitsObject.warnings}).`
-        });
```
dpopp07, line 275: I wasn't a fan of the fact that we were adding a made-up rule here to signal the limit being exceeding (though it was almost certainly my idea a few years ago 🙂 )

I think this is more simple and direct

padamstx (approved):
LGTM

PR #567, opened 2023-03-15
refactor: apply config options to all output formats

dpopp07:
The configuration options, such as "errors only" and "summary only", need to apply to the JSON output, as well as the text output. This commit refactors the logic so that the configuration options are honored during the actual construction of the result object, so that it always represents the results according to the configuration.


docs/ibm-legacy-rules.md
dpopp07, line 1: All of these rules have been removed at this point

packages/validator/src/cli-validator/run-validator.js
```diff
@@ -223,26 +217,20 @@ async function runValidator(cliArgs, parseOptions = {}) {
...
+    }
+
+    // If the # of warnings exceeded the warnings limit, then this is an error.
+    const numWarnings = results.warning.summary.total;
+    const warningsLimit = context.config.limits.warnings;
+    if (warningsLimit >= 0 && numWarnings > warningsLimit) {
+      exitCode = 1;
+      logger.error(
+        `Number of warnings (${numWarnings}) exceeds warnings limit (${warningsLimit}).`
+      );
```
dpopp07, line 233: This can be independent of "errors only" now that warnings won't even be in the results for that case

packages/validator/src/cli-validator/utils/print-results.js
```diff
@@ -7,13 +7,12 @@ const each = require('lodash/each');
 const pad = require('pad');
 
 // this function prints all of the output
-module.exports = function print({ chalk, config, logger }, results) {
-  const { summaryOnly, errorsOnly } = config;
+module.exports = function print(context, results) {
+  const { chalk, config, logger } = context;
+  const types = ['error', 'warning', 'info', 'hint'];
```
dpopp07, line 12: A added a few changes like this based on new use of config options and what seemed best for moving args around... don't mind me 🙂

packages/validator/test/cli-validator/tests/option-handling.test.js
```diff
@@ -190,6 +189,13 @@ describe('cli tool - test option handling', function() {
       expect(outputObject.error.results[0]['message']).toEqual(
         'Every operation must have unique "operationId".'
       );
+
+      // json output should comply with written schema
+      const outputSchemaPath =
+        __dirname + '/../../../src/schemas/results-object.yaml';
+      const outputSchema = await readYaml(outputSchemaPath);
+      const results = validate(outputObject, outputSchema);
```
padamstx, line 197: Aha! so this is why you needed to make updates to results-object.yaml :joy:

padamstx (approved):
Looks great!

packages/validator/test/cli-validator/tests/option-handling.test.js
dpopp07: Haha yep, exposed all of the mistakes I made while writing it the first time 😂

PR #570, opened 2023-03-17
chore: update prettier version and adjust code

dpopp07:
Even though it's a perfectly compatible change, I thought the new version would be a good opportunity to update our linting rules which is something I've wanted to do for a while.

This commit bumps our `prettier` version to 2.x, which includes some beneficial defaults; primarily, enforcing trailing commas. I realize trailing commas may not be viewed as "beneficial" by all :) but I think they are valuable and the AirBnb JS style guide (which we try to use as our source of truth, not that we're following it perfectly) requires them. Happy to debate the change but thought I'd propose it because the lack of trailing commas has bugged me for a while haha.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,10 +8,13 @@ module.exports = {
   },
   'parser': 'babel-eslint',
   'rules': {
-    'prettier/prettier': ['error', {'singleQuote': true}],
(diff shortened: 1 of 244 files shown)


padamstx (approved):
Looks great!

PR #574, opened 2023-03-28
build: require node v16 as minimum version

dpopp07:
Node v14 will be sunset [very soon](https://endoflife.date/nodejs). We should take the v1 opportunity to bump our minimum version to v16.

BREAKING CHANGE: Node v16 is now the minimum supported version of Node for running this tool.


padamstx (approved):
LGTM

PR #578, opened 2023-03-31
docs: clean up individual changelogs

dpopp07:
TIL that the root changelog hasn't been updated since we switched to a monorepo, so I updated it to point to the individual changelogs. I also cleaned those up, as there were a few extraneous lines from the 1.0 release.


padamstx (approved):
Looks good!

PR #585, opened 2023-04-24
fix: include missing dependencies in ruleset package

dpopp07:
The `chalk` dependency was missing from the ruleset package, which would prevent users from using the ruleset as a standalone dependency. This commit adds it in.

Additionally, this commit makes the Spectral ruleset error handler a little smarter. When there are validation errors in the ruleset, Spectral returns an array of errors and we were losing the message, thus providing no info to the user. This adds a check to extract the actual error in these cases.


packages/validator/src/spectral/spectral-validator.js
```diff
@@ -167,6 +167,16 @@ module.exports = {
...
 
 function checkGetRulesetError(logger, error, file) {
+  // Spectral may report an "AggregateError", which could contain multiple errors
+  // and will hide the "message" fields within the individual errors. In this case,
+  // we should extract the first error and process with it instead.
+  if (error instanceof AggregateError && error.errors.length > 0) {
+    logger.debug(
+      'Spectral returned an `AggregateError`. The ruleset likely did not pass validation.'
+    );
+    error = error.errors[0];
```
padamstx, line 177: Should we be logging all the errors returned in the AggregateError instance?

padamstx (approved):
Changes look good, but I was just wondering if we should log each of the errors returned in an AggregateError rather than just the first one.  I guess if we logged each message, then we'd need to check each message to see which hints need to be logged along with it.  Up to you how we handle this, but I just think back to those situations where more info from Spectral might have helped debug problems.  I'm fine with merging the PR as-is.

PR #591, opened 2023-05-11
fix: edit logic to recursively account for nested composed schemas

dpopp07:
and add a test case to demonstrate gap

### DIFF
--- a/packages/ruleset/src/functions/required-property.js
+++ b/packages/ruleset/src/functions/required-property.js
@@ -4,9 +4,8 @@
  */
 
 const {
+  schemaHasProperty,
   validateSubschemas,
-  getPropertyNamesForSchema,
-  isObject,
 } = require('@ibm-cloud/openapi-ruleset-utilities');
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/functions/required-property.js
```diff
@@ -138,24 +105,28 @@ function checkRequiredProperties(schema, path) {
...
+  if (Array.isArray(required)) {
+    const errors = [];
     const localPath = [...path, 'required'];
     logger.debug(`${ruleId}: checking ${localPath.join('.')}`);
 
-    schema.required.forEach(function (name) {
+    required.forEach(name => {
       logger.debug(`${ruleId}: looking for property ${name}`);
-      if (!propNames.includes(name)) {
+      if (!schemaHasProperty(schema, name)) {
```
dpopp07, line 119: This is the main change - using `schemaHasProperty`, which checks recursively that at least one allOf element or all oneOf/anyOf elements contain the property

padamstx (approved):

PR #599, opened 2023-05-24
docs: remove link to external docker image

dpopp07:
There is a community-supported Docker image for running this tool. It is not known to be compromised in any way but it is against our security policies to link to an image controlled by an external agent, so this commit removes the section of the README linking to the external image.

We will likely look into hosting our own image in the future.

This also adds an NPM script for formatting the README's table of contents. The README previously instructed the user to run an NPM package; this way, they don't have to install anything.


padamstx (approved):
LGTM

PR #622, opened 2023-08-18
fix(imports): use correct imports for lodash modules

### LINKED ISSUE #534: Loadash warning
I'm getting this error/warning after running

```
npx lint-openapi src/public/openapi.yaml -e -r docs/ibm-lint-openapi-rules.yaml 
```

```
$ npx lint-openapi src/public/openapi.yaml -e -r docs/ibm-lint-openapi-rules.yaml 
'lodash/isPlainObject' is imported by lodash/isPlainObject?commonjs-external, but could not be resolved – treating it as an external dependency
'lodash/isEqual' is imported by lodash/isEqual?commonjs-external, but could not be resolved – treating it as an external dependency
'lodash/uniqWith' is imported by lodash/uniqWith?commonjs-external, but could not be resolved – treating it as an external dependency
'lodash/pickBy' is imported by lodash/pickBy?commonjs-external, but could not be resolved – treating it as an external dependency
```

I'm using v `0.97.3`

dpopp07:
The `lodash` packages supports installing individual modules via NPM but we use enough modules that we install the whole package. However, we were using the import style for individual modules, which didn't break because the path in `node_modules` was still correct but caused warnings in Spectral (when using custom YAML or JSON rulesets) because of the incorrect names.

This resolves those warnings, which are annoying to users.

Resolves #534

### DIFF
--- a/packages/ruleset/src/functions/array-attributes.js
+++ b/packages/ruleset/src/functions/array-attributes.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache2.0
  */
 
-const isPlainObject = require('lodash/isPlainObject');
+const { isPlainObject } = require('lodash');
 const {
   validateNestedSchemas,
   isArraySchema,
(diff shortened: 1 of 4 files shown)


padamstx (approved):
Looks good!

PR #624, opened 2023-08-21
fix(build): run all release steps for validator build

dpopp07:
The multi-sem-rel package we use merges the individual release configs into the base project's config. I thought that it deep-merged the arrays but it turns out that the validator package's `plugins` field was overriding the whole `plugins` field in the root config and so we were never actually going to release the package (because the NPM plugin would never be called).

This adds the rest of the plugins we call in the base config to the validator config, so things should return to normal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,10 @@
   "version": "0.0.0",
   "license": "Apache-2.0",
   "private": true,
-  "repository": "https://github.com/IBM/openapi-validator",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/IBM/openapi-validator.git"
+  },
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,7 +8,10 @@
   "version": "0.0.0",
   "license": "Apache-2.0",
   "private": true,
-  "repository": "https://github.com/IBM/openapi-validator",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/IBM/openapi-validator.git"
+  },
```
dpopp07, line 14: We were getting warnings in Travis that our `package.json` files were incorrectly formatted. Sure enough, NPM changed the structure a while back and we missed the memo. This commit updates to the structure recommended in their docs

padamstx (approved):
LGTM

PR #625, opened 2023-08-28
feat(ibm-no-operation-requestbody): add new spectral rule

padamstx:
## PR summary
This commit introduces a new spectral-style rule
'ibm-no-operation-requestbody' which will check to make sure that certain operations (delete, get, head, options) do not have a requestBody defined (as recommended by the API Handbook).

### General checklist

### DIFF
--- a/docs/ibm-cloud-rules.md
+++ b/docs/ibm-cloud-rules.md
@@ -58,6 +58,7 @@ which is delivered in the `@ibm-cloud/openapi-ruleset` NPM package.
   * [ibm-no-duplicate-description-with-ref-sibling](#ibm-no-duplicate-description-with-ref-sibling)
   * [ibm-no-if-modified-since-header](#ibm-no-if-modified-since-header)
   * [ibm-no-if-unmodified-since-header](#ibm-no-if-unmodified-since-header)
+  * [ibm-no-operation-requestbody](#ibm-no-operation-requestbody)
   * [ibm-no-optional-properties-in-required-body](#ibm-no-optional-properties-in-required-body)
   * [ibm-no-space-in-example-name](#ibm-no-space-in-example-name)
   * [ibm-openapi-tags-used](#ibm-openapi-tags-used)
@@ -252,7 +253,9 @@ for any resources (paths) that support the <code>If-Match</code> and/or <code>If
 <tr>
 <td><a href="#ibm-no-body-for-delete">ibm-no-body-for-delete</a></td>
 <td>warn</td>
-<td>DELETE operations should not contain a requestBody.</td>
+<td>[Deprecated] DELETE operations should not contain a requestBody.  This rule has been deprecated; please use
+the <code>ibm-no-operation-requestbody</code> rule instead.
+</td>
 <td>oas3</td>
 </tr>
 <tr>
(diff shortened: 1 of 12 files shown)


docs/ibm-cloud-rules.md
```diff
@@ -2980,6 +2991,103 @@ paths:
...
+</pre>
+</td>
+</tr>
+<tr>
+<td valign=top><b>Compliant example:</b></td>
+<td>
+<pre>
+paths:
+  '/v1/things/search':
+    post:     &lt;&lt;&lt; GET changed to POST
```
dpopp07, line 3070: Perhaps this should just be a GET without a request body? Not a big deal, I get what you are demonstrating here, just seemed odd to me that the operation switched

packages/ruleset/src/functions/delete-body.js
```diff
@@ -12,6 +12,22 @@ module.exports = function (operation, _opts, context) {
...
+    // but spectral will still invoke the rule despite the severity setting.
+    // If a user enables this rule in their custom ruleset, we'll want to log a warning
+    // to alert them to the deprecation status.
+    if (context.rule && context.rule.definition) {
+      const severity = context.rule.definition.severity;
+      if (
+        (typeof severity === 'number' && severity >= 0) ||
+        (typeof severity === 'string' && severity !== 'off')
+      ) {
+        logger.warn(
```
dpopp07, line 26: Will this log more than once if the rule is enabled and there are multiple operations?

packages/ruleset/src/functions/index.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
   mergePatchProperties: require('./merge-patch-properties'),
   operationIdCasingConvention: require('./operationid-casing-convention'),
   operationIdNamingConvention: require('./operationid-naming-convention'),
+  operationRequestBody: require('./operation-requestbody'),
```
dpopp07, line 27: I know this is picky 😅 but to be consistent with my comment from last time - it really helps me when debugging rules and digging through code to have the rule/function files match the rule ID (i.e. `no-operation-requestbody`) when possible. If it's a hassle, I don't mind leaving it but my guess is it'll be less intuitive down the road than it seems now

dpopp07 (commented):
Looks good! A couple minor things but the logic itself looks great

docs/ibm-cloud-rules.md
padamstx: haha, yeah I guess I was looking at it from the opposite angle... i.e. using a GET where we should really use a POST instead :)
I'll change it

packages/ruleset/src/functions/delete-body.js
padamstx: No, this is inside the "if (!logger)" above, which we'll enter only the first time we are invoked.  I initially started setting up a separate flag to make sure we log the warning only once, then realized I could piggyback off the if block that was already there to initialize the logger.

dpopp07: Ah, of course - I figured you wouldn't have implemented it in a way that would log multiple messages but I was struggling to understand how you did it (totally missed that context 🙂)

packages/ruleset/src/functions/index.js
padamstx: actually, I had initially was going to use "ibm-operation-requestbody" as the rule name, but then changed to "ibm-no-operation-requestbody" and forgot to rename the files after that.  will fix

padamstx: done

padamstx:
Ok, just pushed changes to address review comments

dpopp07 (approved):
Looks good! 👍

PR #627, opened 2023-09-07
feat(new-rule): ibm-schema-naming-convention

dpopp07:
Adds a new rule that enforces the schema naming guidelines in the API Handbook.

Specifically, the new rule checks for correctly named collection schemas, resource collection element schemas, creation/replacement schemas, and patch schemas against the name of the associated canonical schema.

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/resource-response-consistency.js
+++ b/packages/ruleset/src/functions/resource-response-consistency.js
@@ -8,6 +8,7 @@ const { isObject } = require('@ibm-cloud/openapi-ruleset-utilities');
 const {
   getResourceSpecificSiblingPath,
   getResponseCodes,
+  getSuccessResponseSchemaForOperation,
   isCreateOperation,
   isJsonMimeType,
   isOperationOfType,
@@ -167,9 +168,9 @@ function resourceResponseConsistency(operation, path, apidef) {
 
 function getCanonicalSchema(path, apidef) {
   const resourceSpecificPath = isOperationOfType('post', path)
-    ? getResourceSpecificSiblingPath(path, apidef)
+    ? getResourceSpecificSiblingPath(path.at(-2), apidef)
     : // This is a PUT or PATCH and should already be on the path we need
-      path[path.length - 2].toString().trim();
+      path.at(-2).trim();
 
   logger.debug(
     `${ruleId}: calculated resource-specific path to be "${resourceSpecificPath}"`
@@ -189,62 +190,11 @@ function getCanonicalSchema(path, apidef) {
     );
     return;
   }
-  if (!resourceGetOperation.responses) {
-    logger.debug(
-      `${ruleId}: no responses defined on GET operation at path "${resourceSpecificPath}"`
-    );
-    return;
-  }
 
-  const [, getOpSuccessCodes] = getResponseCodes(
-    resourceGetOperation.responses
+  const { schemaObject } = getSuccessResponseSchemaForOperation(
+    resourceGetOperation,
+    path
   );
 
-  logger.debug(
-    `${ruleId}: corresponding GET operation has the following status codes: ${getOpSuccessCodes.join(
-      ', '
-    )}`
-  );
-
-  if (
-    !getOpSuccessCodes ||
-    !getOpSuccessCodes.length ||
-    !getOpSuccessCodes.includes('200')
-  ) {
-    logger.debug(
-      `${ruleId}: no 200 success code found in responses defined on GET operation at path "${resourceSpecificPath}"`
-    );
-    return;
-  }
-
-  const successResponse = resourceGetOperation.responses['200'];
-  if (!successResponse.content || !isObject(successResponse.content)) {
-    return;
-  }
-
-  // Find the first content object determined to be JSON -
-  // we are only interested in JSON content.
-  const jsonMimeType = Object.keys(successResponse.content).find(mimeType =>
-    isJsonMimeType(mimeType)
-  );
-  if (!jsonMimeType) {
-    logger.debug(
-      `${ruleId}: no JSON content found on 200 response defined on GET operation at path "${resourceSpecificPath}"`
-    );
-    return;
-  }
-
-  const jsonContent = successResponse.content[jsonMimeType];
-  if (!isObject(jsonContent) || !jsonContent.schema) {
-    logger.debug(
-      `${ruleId}: no JSON content schema found in 200 response defined on GET operation at path "${resourceSpecificPath}"`
-    );
-    return;
-  }
-
-  logger.debug(
-    `${ruleId}: found schema in ${jsonMimeType} content object in 200 response defined on GET operation at path "${resourceSpecificPath}"`
-  );
-
-  return jsonContent.schema;
+  return schemaObject;
 }
--- a/packages/ruleset/src/functions/schema-naming-convention.js
+++ b/packages/ruleset/src/functions/schema-naming-convention.js
@@ -1,0 +1,474 @@
+/**
+ * Copyright 2023 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const { schemaHasProperty } = require('@ibm-cloud/openapi-ruleset-utilities');
+
+const {
+  LoggerFactory,
+  getRequestBodySchemaForOperation,
+  getResourceSpecificSiblingPath,
+  getSuccessResponseSchemaForOperation,
+} = require('../utils');
+
+let ruleId;
+let logger;
+
+/**
+ * The implementation for this rule makes assumptions that are dependent on the
+ * presence of the following other rules:
+ *
+ * - ibm-request-and-response-content: all request bodies and relevant success
+ *   responses define content objects
+ *
+ * - ibm-content-contains-schema: the content objects define schemas
+ *
+ * - ibm-avoid-inline-schemas: all schemas are named (defined with references)
+ *
+ * - ibm-collection-array-property: the presence and correct name of the
+ *   property in a collection schema that holds the resource list
+ *
+ * - (yet to be developed): schema names use upper camel case
+ */
+
+module.exports = function schemaNames(apidef, options, context) {
+  if (!logger) {
+    ruleId = context.rule.name;
+    logger = LoggerFactory.getInstance().getLogger(ruleId);
+  }
+  return checkSchemaNames(apidef, context.documentInventory.graph.nodes);
+};
+
+/**
+ * This function checks for most of the API Handbook's schema naming conventions.
+ * For each pair of resource-oriented paths, it determines the name of the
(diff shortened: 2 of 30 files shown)


packages/ruleset/src/functions/resource-response-consistency.js
```diff
@@ -8,6 +8,7 @@ const { isObject } = require('@ibm-cloud/openapi-ruleset-utilities');
 const {
   getResourceSpecificSiblingPath,
   getResponseCodes,
+  getSuccessResponseSchemaForOperation,
```
dpopp07, line 11: I needed a stretch of code that already existed in this function so I moved it out to a utility

packages/ruleset/src/functions/resource-response-consistency.js
```diff
@@ -167,9 +168,9 @@ function resourceResponseConsistency(operation, path, apidef) {
 
 function getCanonicalSchema(path, apidef) {
   const resourceSpecificPath = isOperationOfType('post', path)
-    ? getResourceSpecificSiblingPath(path, apidef)
+    ? getResourceSpecificSiblingPath(path.at(-2), apidef)
```
dpopp07, line 171: I also adjusted this utility to be more flexible for use in the new rule

packages/ruleset/test/array-attributes.test.js
```diff
@@ -497,13 +497,11 @@ describe(`Spectral rule: ${ruleId}`, () => {
       };
 
       const results = await testRule(ruleId, rule, testDocument);
-      expect(results).toHaveLength(6);
+      expect(results).toHaveLength(4);
```
dpopp07, line 500: I had to make adjustments to the "root" testing document to adhere to the API Handbook naming guidance - this resulted in a few of the rules catching less problems since they only happen in the canonical schema, for example

packages/ruleset/test/utils/root-document.js
```diff
@@ -669,13 +669,73 @@ module.exports = {
             maxLength: 1024,
           },
         },
+        example: {
+          id: 'acb123',
+          name: 'The Two Towers',
+          director: 'Peter Jackson',
+          running_time: 179,
```
dpopp07, line 676: Some of these other changes are to address Spectral-core rules that flagged things about this document when I exported it to run the whole validator against it. Just house cleaning changes

packages/ruleset/src/functions/schema-naming-convention.js
```diff
@@ -0,0 +1,481 @@
...
+ *   responses define content objects
+ * 
+ * - ibm-content-contains-schema: the content objects define schemas
+ * 
+ * - ibm-avoid-inline-schemas: all schemas are named (defined with references)
+ * 
+ * - ibm-collection-array-property: the presence and correct name of the
+ *   property in a collection schema that holds the resource list
+ * 
+ * - (yet to be developed): schema names use upper camel case
```
dpopp07, line 36: I'm working on this rule now - I'm going to push it to this same `api-symmetry` branch before I merge either of them, since they are inter-dependent

packages/ruleset/src/functions/schema-naming-convention.js
```diff
@@ -0,0 +1,474 @@
...
+ *                                       been sanitzed for our purposes. it maps referenced
+ *                                       schema locations to their reference location.
+ * @returns {string} - the name of the referenced schema at a given path, or undefined
+ */
+function getSchemaNameAtPath(path, pathToReferencesMap) {
+  if (!path && typeof path !== 'string') {
+    return;
+  }
+
+  // resolve path to reference - add comments explaining
```
padamstx, line 395: Does this line still belong?  It just looks like notes to yourself during development is all :)

packages/ruleset/src/functions/schema-naming-convention.js
```diff
@@ -0,0 +1,474 @@
...
+  // Build the path, replacing each path that resolves to a reference with the
+  // referenced path in order to match the expected format in the
+  // pathToReferencesMap (which comes from graph nodes that Spectral gives us).
+  // See the function documentation above for more info.
+  let pathBuilder = '';
+  for (const pathSegment of path.split('.')) {
+    pathBuilder += `${pathSegment}`;
```
padamstx, line 402: Another way to do this would be add a period to the string immediately before adding the next path segment (as long as pathBuilder already has SOMETHING in it):
```suggestion
    if (pathBuilder) {
      pathBuilder += '.';
    }
    pathBuilder += `${pathSegment}`;
```
then remove the `pathBuilder += '.';` line below

padamstx: that way you don't need to remove anything later on.

packages/ruleset/src/functions/schema-naming-convention.js
```diff
@@ -0,0 +1,474 @@
...
+ * ThingPrototype.properties.data').
+ *
+ * @param {string} path - the resolved JSON path to a schema, as a dot-separated string
+ * @param {object} pathToReferencesMap - the graph nodes map from Spectral, which has
+ *                                       been sanitzed for our purposes. it maps referenced
+ *                                       schema locations to their reference location.
+ * @returns {string} - the name of the referenced schema at a given path, or undefined
+ */
+function getSchemaNameAtPath(path, pathToReferencesMap) {
+  if (!path && typeof path !== 'string') {
```
padamstx, line 391: Should this actually be:
```suggestion
  if (!path || typeof path !== 'string') {
```
?

packages/ruleset/src/functions/schema-naming-convention.js
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * Takes a path to a referenced schema (as a string) and extracts the last element,
+ * which will be the name of the schema.
+ *
+ * @param {string} reference - the ref value as a dot-separated string (e.g. 'components.schemas.Thing')
+ * @returns {string} - the name of the schema (e.g. 'Thing')
+ */
+function getSchemaNameFromReference(reference) {
+  if (!reference && typeof reference !== 'string') {
```
padamstx, line 434: Should this also use || instead of &&?
```suggestion
  if (!reference || typeof reference !== 'string') {
```

padamstx (commented):
Overall looks great.  Just had a few questions to consider

packages/ruleset/src/utils/get-resource-specific-sibling-path.js
```diff
@@ -6,30 +6,27 @@
...
- * Given an operation on a generic resource path (e.g. /foo) to an operation,
- * look for a specific resource path that is a sibling (e.g. /foo/{id}); if
- * found, return the sibling path as a string.
- *
- * Note: the given path MUST be to an operation object.
+ * Given a "generic" resource path string (e.g. '/foo'), look for a
+ * "specific" resource path that is a sibling (e.g. /foo/{id}); if
+ * found, return the sibling path string.
  *
  * @param {*} path the array of path segments indicating the "location" of an operation within the API definition
```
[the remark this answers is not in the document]
dpopp07: Ah no, good catch!

packages/ruleset/src/utils/schema-finding-utils.js
```diff
@@ -0,0 +1,136 @@
...
+  isJsonPatchMimeType,
+  isMergePatchMimeType,
+} = require('./mimetype-utils');
+
+/**
+ * Takes an operation object/path combo and finds the JSON success
+ * response (200) schema, if it exists. We're only interested in
+ * JSON content and this is to be used for exclusively GET operations
+ * (individual "gets" or "lists") so we only look at 200 status codes.
+ *
```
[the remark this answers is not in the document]
dpopp07: You bring up a good point - it does look like the Handbook is not even so strict everywhere and we don't have rules verifying that behavior. It may be best to be more flexible. Though, in the case of determining a canonical schema, I'm not sure what other status code would make sense.

I can make this more flexible, or perhaps the answer is to add a new rule that verifies the existence of 200 responses for "get" and "lists" on resource-oriented paths. Or perhaps both 🙂 What do you think?

packages/ruleset/src/functions/schema-naming-convention.js
dpopp07: Haha that's exactly what that is - good catch!

packages/ruleset/src/functions/schema-naming-convention.js
dpopp07: Good call - I knew there was a better way to handle that but never ended up revisiting it. I'll make that change

packages/ruleset/src/functions/schema-naming-convention.js
dpopp07: Yes 🙂

packages/ruleset/src/functions/schema-naming-convention.js
dpopp07: Yes again 🙂

packages/ruleset/test/schema-naming-convention.test.js
```diff
@@ -0,0 +1,860 @@
...
+            },
+          },
+        },
+      };
+
+      const results = await testRule(ruleId, rule, testDocument);
+      expect(results).toHaveLength(0);
+    });
+
+    it('Prototype schema in PUT request body is correctly named but the path is not resource-oriented', async () => {
```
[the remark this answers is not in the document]
dpopp07: You're right, it should be. I definitely made sure you were paying attention on this one 😂

packages/ruleset/src/utils/schema-finding-utils.js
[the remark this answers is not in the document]
padamstx: I think for this rule it might be good to be more tolerant of "incorrect" success status codes...  perhaps just look for the first response that matches against `2\d\d` and defines a schema?
Separately, we might also want to have a rule that makes sure get operations use the correct status codes, although I think we might have something in that area already IIRC.

packages/ruleset/test/schema-naming-convention.test.js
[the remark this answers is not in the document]
padamstx: LOL

packages/ruleset/src/utils/schema-finding-utils.js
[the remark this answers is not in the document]
dpopp07: This is in a utility so it'll apply to more rules than this current one, but I agree it should be more flexible. I can make that change.

We do have a rule in this area but it's not strict on GET operations - I believe it's fine with any success code. That may be correct - e.g. I think it's acceptable to use a 204 for a GET if the resource has no body representation. So perhaps we just make this more flexible and leave the other rules as is

dpopp07:
All comments have been addressed

padamstx (approved):
LGTM

PR #628, opened 2023-09-08
feat(new-rule): ibm-schema-casing-convention

dpopp07:
This adds a new rule to verify that schema names (i.e. the keys in the 'schemas' object in the 'components' object) are upper camel case, as specified in the API Handbook.

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/schema-casing-convention.js
+++ b/packages/ruleset/src/functions/schema-casing-convention.js
@@ -1,0 +1,50 @@
+/**
+ * Copyright 2023 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const { isObject } = require('@ibm-cloud/openapi-ruleset-utilities');
+const { casing } = require('@stoplight/spectral-functions');
+const { LoggerFactory } = require('../utils');
+
+let casingConfig;
+let ruleId;
+let logger;
+
+module.exports = function (components, options, context) {
+  // Save this rule's "functionOptions" value since we need
+  // to pass it on to Spectral's "casing" function.
+  casingConfig = options;
+
+  if (!logger) {
+    ruleId = context.rule.name;
+    logger = LoggerFactory.getInstance().getLogger(ruleId);
+  }
+
+  return schemaCaseConvention(components, context.path);
+};
+
+function schemaCaseConvention(components, path) {
+  if (!components.schemas && !isObject(components.schemas)) {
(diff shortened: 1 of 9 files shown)


packages/ruleset/src/functions/schema-casing-convention.js
```diff
@@ -0,0 +1,50 @@
...
+  if (!logger) {
+    ruleId = context.rule.name;
+    logger = LoggerFactory.getInstance().getLogger(ruleId);
+  }
+
+  return schemaCaseConvention(components, context.path);
+};
+
+function schemaCaseConvention(components, path) {
+  if (!components.schemas && !isObject(components.schemas)) {
```
padamstx, line 28:
```suggestion
  if (!components.schemas || !isObject(components.schemas)) {
```
Should this be an OR instead of an AND?

padamstx (changes requested):
Overall, this looks good although I think I found one little && vs || gotcha :)

Also... in addition to checking the keys within `components.schemas`, I was wondering if we should consider checking the `name` and/or `title` fields belonging to inline object schemas as well.  I haven't thought about all the ramifications of that, but it just came to mind while reviewing is all.

packages/ruleset/src/functions/schema-casing-convention.js
dpopp07: I don't know why I keep doing that 😅 I will fix!

dpopp07:
> I was wondering if we should consider checking the name and/or title fields

That's a good question, needed to take some time to think it through. I'm not sure if there is a `name` field for schemas, but from what I read in the JSONSchema spec, `title` seems to be a field to declare the purpose of the schema, not give it a name. I can't remember if we use it for SDK Gen or not but even if we do, tooling isn't as much in scope IMO since we adjust casing per language anyways.

Another thing is, we do have a rule to check for schemas not being defined inline. For the exceptions to that rule, we probably don't need names anyways and so I don't think they're important to check with this rule. That said - I will mark this rule dependent on that rule. Something I'm trying to get in the habit of is view the rules as atomic but still inter-connected and allow rules to make assumptions based on what we check for in other rules. I want to keep track of what those assumptions are in the comments and I missed that for that rule, so I'm glad you reminded me!

In terms of implementation, I think I would advocate for keeping the check limited to `components.schemas`. Let me know how you feel about that

padamstx:
>  Let me know how you feel about that

That's fine.

PR #636, opened 2023-10-01
fix(ibm-schema-naming-convention): allow canonical schema for create/replace

dpopp07:
The Handbook states:

"There is an important exception to this guidance if the nature of an API is to provide clients with full control over entire resources, with no system-defined or immutable aspects. For such APIs, the canonical schema itself SHOULD be used in requests to create or replace a resource."

This commit allows for the name of the canonical schema to be used for resource-oriented POSTs and PUTs.


padamstx (approved):
LGTM

PR #639, opened 2023-10-11
fix(api-symmetry): handle nested reference schemas

dpopp07:
If reference schemas were nested in composed properties on the canonical schema, the code would not resolve them the way it would for non-nested reference schemas.

This moves the resolving code to the constraint checker, so that every schema and applicator sub-schema will be resolved if it is a reference.

### DIFF
--- a/packages/ruleset/src/functions/api-symmetry.js
+++ b/packages/ruleset/src/functions/api-symmetry.js
@@ -145,60 +145,43 @@ function checkForGraphFragmentPattern(
   schemaFinder
 ) {
   let variantOmitsProperties = false;
-  let isValidGraphFragment = true;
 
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/functions/api-symmetry.js
```diff
@@ -145,60 +145,43 @@ function checkForGraphFragmentPattern(
   schemaFinder
 ) {
   let variantOmitsProperties = false;
-  let isValidGraphFragment = true;
```
dpopp07, line 148: This didn't necessarily cause problems but it did make debugging a lot harder, since once this got set, every call to `isGraphFragment` would return false. Now, we use a local variable which is probably a more clean and appropriate overall.

pyrooka (approved):
As far as I can tell, it looks good!

PR #641, opened 2023-11-10
feat: display actual ruleset version with --version option

dpopp07:
Before this commit, the --version option would print the version of the validator tool and the version of the IBM Cloud OpenAPI Ruleset that is run by default if the user doesn't provide their own ruleset.

However, if the user did use their own, locally installed version of our ruleset, the version included in the output of the tool would not actually reflect the version being used. Additionally, users could install new versions of the validator tool but if they didn't update their locally installed instance of the ruleset, they would not get the updated behavior of the rules (which are often the reasons for bumping the version of the tool).

This commit adds logic to print the actually used (whether the default or locally installed) ruleset version and additionally will print a warning if the user has upgraded the version of their tool without upgrading the version of their locally installed ruleset.

This all required a little bit of re-organizing of some of the code.

### DIFF
--- a/packages/validator/src/cli-validator/utils/print-versions.js
+++ b/packages/validator/src/cli-validator/utils/print-versions.js
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2023 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const getDefaultRulesetVersion = require('./get-default-ruleset-version.js');
+const getLocalRulesetVersion = require('./get-local-ruleset-version');
+const getVersionString = require('./get-version-string');
+const checkRulesetVersion = require('./check-ruleset-version');
+const { findSpectralRuleset } = require('../../spectral/utils');
+
+module.exports = printVersions;
+
+/**
+ * Prints the "version" strings upon request. It always
+ * includes the semantic version of the validator tool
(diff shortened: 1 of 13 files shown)


packages/validator/src/cli-validator/utils/print-versions.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+  } else {
+    // If the user isn't using a custom, local ruleset
+    // we will use our default ruleset.
+    versionInfo.ruleset = `ruleset: ${getDefaultRulesetVersion()} (default)`;
+  }
+
+  if (!versionInfo.ruleset) {
+    // Add a note explaining why a ruleset version is not inlcuded in the output.
```
padamstx, line 88: Usually, I defer to @pyrooka on these types of matters, but I noticed he's not tagged as a reviewer so I feel compelled to point out this mis-spelled word :joy:
```suggestion
    // Add a note explaining why a ruleset version is not included in the output.
```

padamstx (approved):
LGTM, just one typo that I could find :)

packages/validator/src/cli-validator/utils/print-versions.js
pyrooka: I am so proud! 😂

PR #642, opened 2023-12-20
refactor: expose cli tool utilities under a single export

dpopp07:
This is a small, mostly superficial change I've been wanting to make for a while, so I took some time to do it while things are slow 🙂 

To better organize the private utility functions for the CLI validator, this commit wraps their exposure within a single object exported from an index.js file. This simplifies the import syntax in our other files.


padamstx (approved):
LGTM!

PR #645, opened 2024-01-29
fix(ibm-resource-response-consistency): expand support to bulk operations

dpopp07:
## PR summary

This commit adds support for recognizing bulk operations and recommending the collection schema be returned instead of the canonical schema. It also:
- Updates the message to be more clear
- Updates the path to be more aligned with the operation
- Adds an "info" log pointing to the specific operation in case it is unclear

Previously, the rule would flag a bulk operation that didn't return the collection schema but it did so unintentionally (and with a bad heuristic) and the message would indicate that the canonical schema should be returned, which was false (often, the canonical schema *was* being returned).

Also, the warning would be tied to a specific schema and it was difficult to parse the debug logs to determine which operation was responsible for the warning.

### General checklist

### DIFF
--- a/docs/ibm-cloud-rules.md
+++ b/docs/ibm-cloud-rules.md
@@ -5233,7 +5233,7 @@ paths:
 </tr>
 <tr>
 <td valign=top><b>Description:</b></td>
-<td>Synchronous responses for create-style POST or update (PUT/PATCH) operations on a given resource should return the canonical schema for the resource. As in, the same schema as the "GET" request for the same resource.</td>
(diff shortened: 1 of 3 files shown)


docs/ibm-cloud-rules.md
```diff
@@ -5226,7 +5226,7 @@ paths:
 </tr>
 <tr>
 <td valign=top><b>Description:</b></td>
-<td>Synchronous responses for create-style POST or update (PUT/PATCH) operations on a given resource should return the canonical schema for the resource. As in, the same schema as the "GET" request for the same resource.</td>
+<td>Synchronous responses for create-style POST or update (PUT/PATCH) operations on a given resource should return the same schema as the "GET" request for the same resource. Requests that operate on a single resource should return the canonical schema. Requests that perform bulk operations, like a PUT that performs a bulk replace, should return the collection schema.</td>
```
padamstx, line 5229: We discussed re-wording this one...

padamstx (approved):
Overall looks great. Just had a couple comments about docs and comments, but I'll approve now to avoid a re-review.

PR #646, opened 2024-01-31
fix(resource-response-consistency): avoid destructuring property of undefined values

dpopp07:
Addresses a bug reported in which a property is attempted to be deconstructed from an object type. When the `getCanonicalSchema` function was converted to return an object, a return statement was left unconverted and was still returning undefined. This caused the bug upon attempted deconstruction of the return value.

Also, I updated the project copyright string for the new year.


padamstx (approved):
LGTM

PR #648, opened 2024-02-23
fix: adjust header parameter casing convention check

dpopp07:
This allows us to support standard header patterns that would previously be flagged by the validator. The `casing` function options provided by Spectral are not sophsiticated enough to support the complexity of header parameter conventions (which use kebab-separated pascal case with capitalized abbreviations).

The new regular expression is a little more relaxed than the old convention but it also more accurate.

I made this change in a configuration-compatible way by expanding the config support to accept function arguments for either the casing or pattern built-in Spectral functions. This is slightly hacky but gives both us and users more control without changing any existing behavior. We might want to make this change for our other casing convention checks but for now, this is sufficient to solve our issues. The way we support configuration of Spectral functions is complicated, especially when we decide we don't want to use those functions 🙂 

I'm also happy to make this a "hidden feature" and take it out of the document and just use it as a convenience pattern for us to control casing through regular expressions without changing our configuration API.

### DIFF
--- a/docs/ibm-cloud-rules.md
+++ b/docs/ibm-cloud-rules.md
@@ -4047,8 +4047,8 @@ n/a
 <tr>
 <td valign=top><b>Description:</b></td>
 <td>This rule verifies that each parameter name complies with the casing convention associated with that parameter's type.
-For example, the default casing convention for query parameters is snake-case (e.g. `my-query-param`), the default for path parameters
-is snake-case (e.g. `my_path_param`), and the default casing convention for header params is kebab-separated pascal-case with provision for capitalized abbreviations (e.g. `IBM-CustomHeader-Name`).
+For example, the default casing convention for query parameters is snake-case (e.g. <code>my_query_param</code>), the default for path parameters
+is snake-case (e.g. <code>my_path_param</code>), and the default casing convention for header params is kebab-separated pascal-case with provision for capitalized abbreviations (e.g. <code>IBM-CustomHeader-Name</code>).
 These default casing conventions constitute the default configuration for the rule, although the rule's configuration can be modified to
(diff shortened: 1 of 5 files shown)


docs/ibm-cloud-rules.md
```diff
@@ -4048,7 +4048,7 @@ n/a
 <td valign=top><b>Description:</b></td>
 <td>This rule verifies that each parameter name complies with the casing convention associated with that parameter's type.
 For example, the default casing convention for query parameters is snake-case (e.g. `my-query-param`), the default for path parameters
```
padamstx, line 4050:
```suggestion
For example, the default casing convention for query parameters is snake-case (e.g. <code>my_query_param</code>), the default for path parameters
```
I realize you didn't change this, but my-query-param isn't snake case, is it? :)

docs/ibm-cloud-rules.md
```diff
@@ -4048,7 +4048,7 @@ n/a
...
-is snake-case (e.g. `my_path_param`), and the default casing convention for header params is pascal-case (e.g. `X-My-Custom-Header`).
+is snake-case (e.g. `my_path_param`), and the default casing convention for header params is kebab-separated pascal-case with provision for capitalized abbreviations (e.g. `IBM-CustomHeader-Name`).
```
padamstx, line 4051:
```suggestion
is snake-case (e.g. <code>my_path_param</code>), and the default casing convention for header params is kebab-separated pascal-case with provision for capitalized abbreviations (e.g. <code>IBM-CustomHeader-Name</code>).
```

padamstx (approved):
LGTM, just had a couple comments about the docs.  I'll approve to now to avoid a re-review.

PR #657, opened 2024-04-03
fix(ibm-api-symmetry): improve validation logic to avoid false positives

padamstx:
## PR summary
This PR fixes the ibm-api-symmetry rule to avoid false positives.

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/api-symmetry.js
+++ b/packages/ruleset/src/functions/api-symmetry.js
@@ -215,15 +215,15 @@ function checkForGraphFragmentPattern(
             canonical,
             canonicalPath,
             schemaFinder,
-            s =>
+            (schema, path) =>
               // Note that these first two conditions are guaranteed to be met at
               // least once by the first call to `canonicalSchemaMeetsConstraint`
-              'properties' in s &&
-              isObject(s.properties[name]) &&
+              'properties' in schema &&
+              isObject(schema.properties[name]) &&
               isGraphFragment(
                 prop,
-                s.properties[name],
-                [...canonicalPath, 'properties', name],
+                schema.properties[name],
+                [...path, 'properties', name],
                 false
               )
           )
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/functions/api-symmetry.js
```diff
@@ -253,15 +268,21 @@ function checkForGraphFragmentPattern(
...
+              previousResult &&
+              isGraphFragment(v, canonical, canonicalPath, true),
+            true
+          )
+        ) {
+          logger.info(
+            `${ruleId}: variant schema applicator '${applicator}' is not a graph fragment of the canonical schema`
+          );
+          result = false;
+        }
```
dpopp07, line 285: I'm not sure I understand this change - isn't it logically equivalent but with a log statement added? 🙂 

Let me know if I'm missing something

packages/ruleset/src/functions/api-symmetry.js
```diff
@@ -280,34 +301,58 @@ function checkForGraphFragmentPattern(
...
-          let allAreIncluded = true;
+          let omitted = false;
           for (const [name, prop] of Object.entries(c.properties)) {
-            const included = schemaHasConstraint(
-              variant,
-              s =>
+            logger.debug(`${ruleId}: checking canonical property '${name}'`);
+
+            const included = schemaHasLooseConstraint(variant, s => {
+              return (
```
dpopp07, line 315: This `return` is probably unnecessary - you could get rid of the brackets and use pure arrow notation, which returns implicitly.

packages/ruleset/src/functions/api-symmetry.js
```diff
@@ -319,9 +364,13 @@ function checkForGraphFragmentPattern(
...
   );
 
+  logger.debug(
+    `${ruleId}: isGraphFragment() returned [${variantIsGraphFragment},${variantOmitsProperties}]`
+  );
+
   if (!variantOmitsProperties) {
     logger.info(
-      `${ruleId}: schema does not omit any properties in the canonical schema - it is not a proper graph fragment`
+      `${ruleId}: variant schema properties are not a proper subset of the canonical schema properties`
```
dpopp07, line 373: This is a matter of taste but I still think it helps make it clear to say something about properties not being omitted rather than just indicating it isn't a subset. e.g. it also wouldn't be a proper subset if anything was added, but this check is specifically about not omitting properties and the user should know that's what happened.

I'd be fine with something like this, even:

```suggestion
      `${ruleId}: variant schema properties are not a proper subset of the canonical schema properties (no properties omitted)`
```

dpopp07 (changes requested):
Looks good but I left a couple comments!

Also, perhaps there should be a unit test covering the case that prompted the fix. Let me know what you think.

dpopp07 (approved):
Looks good! As discussed, I'll target your branch with #658 and merge it in here before we merge this one.

PR #664, opened 2024-05-08
Modify tests to work with Node v20

### LINKED ISSUE #662: Fix test failures with Node 20
The project says it supports Node 16+, but does not run a build on Node 20 (which is the current LTS version) and has minor test failures on Node 20.

dpopp07:
## PR summary

Version 20 of Node changes the behavior of error messages. Where the
tests relied on equality with specific error messages, they have been
updated to match against slightly more generic regular expressions to
support running the tests.

Also, v20 is added to the Travis builds as it is the current LTS version.

Resolves #662

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: node_js
 node_js:
 - 16
 - 18
+- 20
 
 cache:
   npm: false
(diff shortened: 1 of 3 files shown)


padamstx (approved):
LGTM

PR #665, opened 2024-05-20
build: avoid CVE via .cveignore file

padamstx:


scripts/create-binaries.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
pyrooka, line 1: Why is this change? I feel it's somewhat specific to the Travis' Ubuntu image...:)

pyrooka (approved):
LGTM!

scripts/create-binaries.sh
padamstx: I made it consistent with the other build script (deploy-container-image.sh).  My understanding is that using `#!/usr/bin/env bash` is "better" in some respects as it doesn't depend on a hard-coded location for the `bash` executable.

PR #666, opened 2024-05-21
feat(ibm-no-superfluous-allof): add new rule and documentation

padamstx:
## PR summary
This commit adds the new "ibm-no-superfluous-allof" rule which warns about the use of a superfluous allOf, which is a schema that contains only a single-element allOf.

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/no-superfluous-allof.js
+++ b/packages/ruleset/src/functions/no-superfluous-allof.js
@@ -1,0 +1,38 @@
+/**
+ * Copyright 2017 - 2024 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const { validateSubschemas } = require('@ibm-cloud/openapi-ruleset-utilities');
+const { LoggerFactory } = require('../utils');
+
+let ruleId;
+let logger;
+
+module.exports = function (schema, _opts, context) {
+  if (!logger) {
+    ruleId = context.rule.name;
+    logger = LoggerFactory.getInstance().getLogger(ruleId);
+  }
+
+  return validateSubschemas(schema, context.path, checkForSuperfluousAllOf);
+};
+
+function checkForSuperfluousAllOf(schema, path) {
+  // We're interested only in schemas that contain ONLY a single-element allOf list.
+  if (
+    Array.isArray(schema.allOf) &&
+    schema.allOf.length === 1 &&
+    Object.keys(schema).length === 1
+  ) {
+    return [
+      {
+        // Use rule description for error message.
+        message: '',
+        path,
+      },
+    ];
+  }
+
+  return [];
+}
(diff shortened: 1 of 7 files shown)


pyrooka (approved):
Looks good to me!

packages/ruleset/src/functions/no-superfluous-allof.js
```diff
@@ -0,0 +1,37 @@
...
+function checkForSuperfluousAllOf(schema, path) {
+  // We're interested only in schemas that contain ONLY a single-element allOf list.
+  if (
+    Array.isArray(schema.allOf) &&
+    schema.allOf.length === 1 &&
+    Object.keys(schema).length === 1
+  ) {
+    return [
+      {
+        message: 'Avoid schemas with a superfluous allOf attribute',
```
dpopp07, line 30: Picky, optional suggestion to slightly reword since it's not the _schema_ that needs to be avoided but the `allOf`.

```suggestion
        message: 'Avoid using superfluous allOf attributes in schemas',
```

packages/ruleset/test/no-superfluous-allof.test.js
```diff
@@ -0,0 +1,115 @@
...
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const { noSuperfluousAllOf } = require('../src/rules');
+const { makeCopy, rootDocument, testRule, severityCodes } = require('./utils');
+
+const rule = noSuperfluousAllOf;
+const ruleId = 'ibm-no-superfluous-allof';
+const expectedSeverity = severityCodes.warning;
+const expectedMsg = 'Avoid schemas containing only a single-element allOf';
```
dpopp07, line 12: Hmm I'm surprised that this message doesn't match the message provided in the function. Do you know why that message isn't being used?

dpopp07 (approved):
Looks good! Left one wording suggestion and one question about the test.

packages/ruleset/test/no-superfluous-allof.test.js
padamstx: This was happening because the rule defn didn't specify the "message" field.  I apparently copy/pasted from the wrong existing rule (one that supplied the error message in the rule itself rather than the rule function) :joy:

packages/ruleset/src/functions/no-superfluous-allof.js
padamstx: After thinking about this a little, I think the error message contained in the rule definition itself (`Avoid schemas containing only a single-element allOf`) might capture the essence of what we're trying to say.  It's not ONLY the single-element allOf that is the "problem"... it's the fact that a schema essentially contains nothing else BUT a single-element allOf. A single-element allOf is perfectly fine within a ref-sibling pattern, for example.  The reason that an allOf might be "superfluous" is that it contains only a single element AND the containing schema doesn't contain anything else.

Having said all that, I think an argument can be made that the path that we report in the error entry should probably reflect the schema that contains the allOf, and not the allOf itself. I.e. it's the otherwise empty schema that renders the single-element allOf as "superfluous".

So I'll make these changes and then we can take another look at the PR and see if that looks ok.

padamstx:
I've pushed some changes to address review comments.  Please take another look.

PR #669, opened 2024-06-18
feat: add new 'ibm-no-ref-in-example' rule

padamstx:
## PR summary
This commit introduces the new 'ibm-no-ref-in-example' rule which will check to make sure that "$ref" is not used in "example" fields.  Note that $ref is valid in an "examples" field, but not in an "example" field.

### General checklist

### DIFF
--- a/docs/ibm-cloud-rules.md
+++ b/docs/ibm-cloud-rules.md
@@ -63,7 +63,9 @@ which is delivered in the `@ibm-cloud/openapi-ruleset` NPM package.
   * [ibm-no-nullable-properties](#ibm-no-nullable-properties)
   * [ibm-no-operation-requestbody](#ibm-no-operation-requestbody)
   * [ibm-no-optional-properties-in-required-body](#ibm-no-optional-properties-in-required-body)
+  * [ibm-no-ref-in-example](#ibm-no-ref-in-example)
   * [ibm-no-space-in-example-name](#ibm-no-space-in-example-name)
+  * [ibm-no-superfluous-allof](#ibm-no-superfluous-allof)
   * [ibm-no-unsupported-keywords](#ibm-no-unsupported-keywords)
   * [ibm-openapi-tags-used](#ibm-openapi-tags-used)
   * [ibm-operation-responses](#ibm-operation-responses)
@@ -348,6 +350,12 @@ which is not allowed.</td>
 <td>info</td>
 <td>If a requestBody schema contains properties that are defined as required, then the requestBody itself
 should probably be required instead of optional.</td>
+<td>oas3</td>
+</tr>
+<tr>
+<td><a href="#ibm-no-ref-in-example">ibm-no-ref-in-example</a></td>
+<td>info</td>
+<td>Makes sure that <code>$ref</code> is not used within an <code>example</code> field.</td>
 <td>oas3</td>
 </tr>
 <tr>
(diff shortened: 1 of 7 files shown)


packages/ruleset/test/no-ref-in-example.test.js
```diff
@@ -0,0 +1,202 @@
...
+      testDocument.paths['/v1/drinks'].get.parameters[0] = {
+        in: 'query',
+        name: 'start',
+        content: {
+          'application/json': {
+            schema: {
+              type: 'string',
+            },
+            example: {
+              $ref: '#/components/schemas/Error',
```
HenryGessau, line 182: Should we also have a few test cases with more deeply nested ref's?
E..g.:
```yaml
example:
  library:
    name: my-library
    address:
      $ref: '#/components/schemas/Address/example'
    books:
    - $ref: '#/components/schemas/Book/example'
```

docs/ibm-cloud-rules.md
```diff
@@ -63,7 +63,9 @@ which is delivered in the `@ibm-cloud/openapi-ruleset` NPM package.
   * [ibm-no-nullable-properties](#ibm-no-nullable-properties)
   * [ibm-no-operation-requestbody](#ibm-no-operation-requestbody)
   * [ibm-no-optional-properties-in-required-body](#ibm-no-optional-properties-in-required-body)
+  * [ibm-no-ref-in-example](#ibm-no-ref-in-example)
   * [ibm-no-space-in-example-name](#ibm-no-space-in-example-name)
+  * [ibm-no-superfluous-allof](#ibm-no-superfluous-allof)
```
dpopp07, line 68: Nice catch!

packages/ruleset/src/rules/no-ref-in-example.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright 2017 - 2024 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const { oas3 } = require('@stoplight/spectral-formats');
+// const { falsy } = require('@stoplight/spectral-functions');
```
dpopp07, line 7: Can be removed

packages/ruleset/test/no-ref-in-example.test.js
dpopp07: This is tricky because I don't think that's _technically_ invalid. It's just probably not what the user wants.

For example, the above _could_ be a representation of this valid JSON object:
```json
"library": {
  "name": "my-library",
  "address": {
    "$ref": "#/components/schemas/Address/example"
  },
  "books": [{ "$ref": "#/components/schemas/Book/example" }]
}
```

where `"$ref"` is just an arbitrary key with no meaning. I know that seems rare and maybe we could warn to remind the user its _likely_ wrong but it's hard to tell the difference between unintentional and intentional use.

dpopp07 (approved):
Looks good! 👍

packages/ruleset/src/rules/no-ref-in-example.js
padamstx: fixed

packages/ruleset/test/no-ref-in-example.test.js
padamstx: @HenryGessau I can understand the reason for your comment, and this scenario certainly has some subtleties associated with it.  But, I agree with Dustin here in that an argument could be made that the example object you have posted is probably valid.  I think the rule should be focused on looking for a top-level "$ref" property within the "example" object, and avoid checking for nested uses of "$ref" since those might be valid.

meem: FWIW, our hacky check found several places in the VPC spec where spec authors have done exactly what Henry's asking about: https://github.ibm.com/riaas/api-spec/pull/6260/files

padamstx: Ok, I'm convinced :joy:
I'll expand the scope of the rule so that it flags a `$ref` property within the `example` field at any level of nesting.

padamstx:
@dpopp07 @HenryGessau @meem
I've updated the rule function so that it performs the search for a `$ref` field recursively, and I've updated the tests accordingly.

packages/ruleset/test/no-ref-in-example.test.js
padamstx: fixed

dpopp07 (approved):
Updated code looks good! 👍

PR #672, opened 2024-06-26
fix: allow property names to be $ref without crashing

dpopp07:
The rules checking for valid ref patterns and for circular references both assumed that any $ref values found in the entire document would be reference values, and therefore strings.

However, it's valid to define the field with the literal name "$ref" in other places - e.g. as a property name. The ref-related rules would crash if it encountered these cases because of the assumption that the values would be strings. This commit adds a simple check to enforce that the rules are, in fact, dealing with a string value.

### General checklist


pyrooka (approved):
Nice little improvement! :)

PR #675, opened 2024-07-11
docs: add programmatic usage section to readme

### LINKED ISSUE #649: Update documentation to reflect latest recommendations
Hello, first of all thanks for open sourcing this project!

As I was searching through some older issues I came across [this recommendation](https://github.com/IBM/openapi-validator/issues/514#issuecomment-1332714100):

> Our recommendation moving forward is to use [Spectral's Node API](https://meta.stoplight.io/docs/spectral/eb68e7afd463e-spectral-in-java-script) and import our ruleset as a dependency to get consistent behavior with the CLI tool. That's what the CLI tool is doing behind the scenes.

To be honest this is the first time I have heard about Spectral and therefore it was not immediately clear to me how to hook this API validation ruleset up with the Spectral NPM package. I was wondering if you could provide a simple update to the current README to reflect your latest recommendation.

Many thanks :)

dpopp07:
Resolves #649

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,7 +663,7 @@ module.exports = {
 </table>
 
 ### Programmatic Usage
-While the validator does not expose an API for usage within a Node program, you can achieve programmatic behavior
+While the validator does not expose an API for usage within a Node.js program, you can achieve programmatic behavior
 consistent with the CLI by using the open-source tool [Spectral's Node API](https://meta.stoplight.io/docs/spectral/eb68e7afd463e-spectral-in-java-script)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -661,6 +662,23 @@ module.exports = {
 </tr>
 </table>
 
+### Programmatic Usage
+While the validator does not expose an API for usage within a Node program, you can achieve programmatic behavior
```
padamstx, line 666:
```suggestion
While the validator does not expose an API for usage within a Node.js program, you can achieve programmatic behavior
```
Just a nit, but in semi-official docs, it should probably be spelled out :)

padamstx (approved):
One small suggestion, but approving now to avoid a re-review

PR #678, opened 2024-08-27
build(deps): force new version of sub-dependency micromatch

dpopp07:
These are only development dependencies, so I'm not triggering a new release.


padamstx (approved):
LGTM

PR #679, opened 2024-08-27
feat(ibm-pattern-properties): enforce use of anchors for property patterns

dpopp07:
This commit adds an additional check to the rule dealing with `pattern-properties`. It requires the use of starting and ending anchors (^ and $, respectively) for the regular expression it defines as its single element, much like we require for the `pattern` field of string schemas.


padamstx (approved):
LGTM

PR #681, opened 2024-09-04
fix(ibm-accept-and-return-models): exempt non-object schemas

dpopp07:
When checking that request/response bodies are defined as models, we don't need to check schemas that aren't objects. These schemas should be objects, but there are other rules that will catch that and it's confusing for the validator to tell users they need the "properties" field on a non-object schema.


pyrooka (approved):
LGTM!

PR #685, opened 2024-09-24
fix: make schema path de-resolution utility more robust

dpopp07:
Before, the check used faulty logic to determine if we were on the last iteration through a loop. This could lead to unintended behavior in rare cases. This clarifies the logic to only execute the last time through the loop and for no other reason (the potential reason being if the last path segment matched an earlier path segment).

### DIFF
--- a/packages/ruleset/src/utils/get-schema-name-at-path.js
+++ b/packages/ruleset/src/utils/get-schema-name-at-path.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 IBM Corporation.
+ * Copyright 2023 - 2024 IBM Corporation.
  * SPDX-License-Identifier: Apache2.0
  */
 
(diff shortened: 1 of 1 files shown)


packages/ruleset/src/utils/get-schema-name-at-path.js
```diff
@@ -42,17 +42,22 @@ function getSchemaNameAtPath(path, pathToReferencesMap) {
   // pathToReferencesMap (which comes from graph nodes that Spectral gives us).
   // See the function documentation above for more info.
   let pathBuilder = '';
-  for (const pathSegment of path.split('.')) {
+  const pathSegments = path.split('.');
+
+  for (let i = 0; i < pathSegments.length; i++) {
```
dpopp07, line 47: The only real change is that now we use a numerically-indexed for-loop so that we know for sure when we're going through it for the last time.

pyrooka (approved):
Looks good! 👍

padamstx (approved):
Only thing I could find was that the copyright year should be updated.

dpopp07:
> Only thing I could find was that the copyright year should be updated.

Good catch! Fixed

PR #691, opened 2024-11-04
fix: bump jsonpath-plus to avoid vulnerability

padamstx:


dpopp07 (approved):
Looks good! 👍 

It seems that [Spectral is releasing this fix soon]. I'm down to make this change as a tactical fix and once they release a new version, we can update our version of their tool that pulls this in and remove the override.

PR #694, opened 2024-11-05
fix: edit code to process composed schemas in context

dpopp07:
This is the result of my experiment to try and get composed schemas, like `allOf`, to be processed within the context of their top-level required field.

The code could potentially be cleaned up but it seems to work alright (all of the tests pass). Let me know what you think or if this gives you any ideas!

### DIFF
--- a/packages/ruleset/src/functions/required-array-properties-in-response.js
+++ b/packages/ruleset/src/functions/required-array-properties-in-response.js
@@ -41,6 +41,25 @@ function checkForOptionalArrays(schema, path) {
       `${ruleId}: examining object schema at location: ${path.join('.')}`
     );
 
+    // If "schema" is an allOf element, then we need to make an educated
+    // guess as to whether or not we should complain about an optional array property.
+    // This will be our heuristic:
+    // If "schema" is an allOf element AND...
+    // 1. DOES NOT INCLUDE the "required" field, then we'll assume that the allOf element
+    //    IS NOT fully-defined and we'll avoid returning errors for any optional array properties
+    //    within "schema".
+    // 2. DOES INCLUDE the "required" field, then we'll assume that the allOf element is
+    //    fully-defined and go ahead and return an error for any optional array properties.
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/functions/required-array-properties-in-response.js
```diff
@@ -36,33 +31,60 @@ module.exports = function (schema, _opts, context) {
...
+    const errors = [];
+
+    // We use 'validateNestedSchemas' at the top level to skip any composed
+    // schema elements. They may be relying on the 'required' field for the
+    // whole composed schema, and we lose that context if we process them in
+    // isolation. Here, we use special logic to recursively look at composed
+    // schemas as a whole, checking both the "top-level" required list and the
+    // required list for each element that we find an array property in.
+    validateComposedSchemas(schema, path, (s, p) => {
+      for (const applicatorType of ['allOf', 'oneOf', 'anyOf']) {
```
padamstx, line 48: I'm not 100% sure that we want to treat oneOf/anyOf with the exact same behavior (wrt the required field) as allOf. Because of the semantics of allOf, it makes sense to me to combine the allOf element's required field with the "contextual" required field. But the slightly different semantics of oneOf/anyOf cause me to view elements in those lists as more self-contained schemas. Perhaps for those, we really only want to look at the local required field within the oneOf/anyOf element itself (i.e. perhaps only add the contextual required list for an allOf element).

dpopp07: That's a good consideration. I'm not 100% sure either, but I think that if a schema that includes a `oneOf` also includes a top-level `required` field, it might make sense to apply that to all of the `oneOf` elements. An object validating against that whole schema would need to have the property (additionally, I would expect it to need to be defined in all of the `oneOf` elements for that reason).

FWIW, I just did a generator test and if I define a property within each `oneOf` element and then make that property required at the top-level, the generated code treats the property as required for the parent schema (which is the actual response schema) but optional for the subschemas.

If we decide not to account for those, it should be easy to modify my code.

PR #698, opened 2024-11-14
check for schemas that are dictionaries of dictionaries

dpopp07:
## PR summary

Two changes:

### fix(ibm-well-defined-dictionaries): flag dictionaries with dictionary-type values

The API Handbook states that the values for dictionaries should not themselves
be dictionaries (no dictionaries of dictionaries). Previously, this rule would
not catch that scenario. Now, it will.

Additionally, more robust logic is used to check for the presence of ambiguous
dictionaries. The old logic would miss flagging valid violations in complex,
composed schemas without the fix.

### feat: add utility for checking composite schemas with a looser constraint

The existing `schemaHasConstraint` function will only determine a schema meets
the given constraint when every element of a `oneOf` or `anyOf` list. This is
desired behavior in the majority of cases. However, in some cases, it is
helpful to note when a schema includes a `oneOf` or `anyOf` list with only a
single element meeting the constraint, instead of all of them. The new utility
function, `schemaLooselyHasConstraint`, delivers this functionality.

### General checklist

### DIFF
--- a/packages/ruleset/src/functions/api-symmetry.js
+++ b/packages/ruleset/src/functions/api-symmetry.js
@@ -9,6 +9,7 @@ const {
   isObjectSchema,
   isArraySchema,
   schemaHasConstraint,
+  schemaLooselyHasConstraint,
 } = require('@ibm-cloud/openapi-ruleset-utilities');
 
 const {
@@ -310,7 +311,7 @@ function checkForGraphFragmentPattern(
           for (const [name, prop] of Object.entries(c.properties)) {
             logger.debug(`${ruleId}: checking canonical property '${name}'`);
 
-            const included = schemaHasLooseConstraint(
+            const included = schemaLooselyHasConstraint(
               variant,
               s =>
                 'properties' in s &&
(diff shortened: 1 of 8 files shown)


packages/ruleset/src/functions/api-symmetry.js
```diff
@@ -310,7 +311,7 @@ function checkForGraphFragmentPattern(
           for (const [name, prop] of Object.entries(c.properties)) {
             logger.debug(`${ruleId}: checking canonical property '${name}'`);
 
-            const included = schemaHasLooseConstraint(
+            const included = schemaLooselyHasConstraint(
```
dpopp07, line 314: I just moved the logic from this file to a new utility since we now need it for more than one rule. I also took the opportunity to change the name, since I wasn't a fan of the old one 😛 I'm open to other name suggestions as well!

packages/ruleset/test/rules/well-defined-dictionaries.test.js
```diff
@@ -12,7 +12,7 @@ const {
 } = require('../test-utils');
 
 const rule = wellDefinedDictionaries;
-const ruleId = 'ibm-valid-path-segments';
+const ruleId = 'ibm-well-defined-dictionaries';
```
dpopp07, line 15: Never noticed this typo before!

packages/ruleset/test/rules/well-defined-dictionaries.test.js
```diff
@@ -39,17 +39,25 @@ describe(`Spectral rule: ${ruleId}`, () => {
       expect(results).toHaveLength(0);
     });
 
-    it('Includes a dictionary that nests a well-defined dictionary with scalar values', async () => {
+    it('Includes a well-defined dictionary with composed scalar values', async () => {
       const testDocument = makeCopy(rootDocument);
 
       testDocument.components.schemas.Movie.properties.metadata = {
         description: 'a dictionary mapping keys to string values',
```
padamstx, line 46:
```suggestion
        description: 'a dictionary mapping keys to integer values',
```

padamstx (approved):
LGTM

packages/ruleset/test/rules/well-defined-dictionaries.test.js
dpopp07: Whoops, nice catch!

PR #700, opened 2024-11-26
feat(markdown-report): add a feature to generate a markdown report containing validator results

dpopp07:
## PR summary

This commit adds the `-m/--markdown-report` option to the validator CLI, which will cause the validator to produce a Markdown file containing all results produced by the validator (including rule violations and impact score data).

### General checklist


packages/validator/src/cli-validator/run-validator.js
```diff
@@ -234,12 +238,40 @@ async function runValidator(cliArgs, parseOptions = {}) {
...
+      impactScore: {
+        ...impactScoreInformation,
+      },
+    };
+
+    // If the user requested a markdown report of the results, print that here.
+    // Note that we need to do this before we check if the "summaryOnly" option
+    // was provided, because that will filter the results and we want to print
+    // all of the results in the report.
+    let markdownReportLocation = false;
```
pyrooka, line 272: Beware, picky comment: I know JS is loosely typed and this code works fine, but it might be better to use a value that's "closer" to strings, since the name of the variable also implies that it will hold the a path to the file. With my somewhat limited JS knowledge, I would consider using `undefined`, `null` or maybe and empty string - instead of a boolean.

pyrooka (approved):
Looks great! I ran some tests in my local env, and the rendered markdown documents are pretty good. Left one picky comment, otherwise I found zero flaws in the logic.

packages/validator/test/test-utils/get-captured-text.js
```diff
@@ -13,11 +13,11 @@ module.exports.getCapturedTextWithColor = callsToLog =>
...
-    // the validator only ever uses the first arg in consolg.log
-    const output = preserveColors ? args[0] : stripAnsi(args[0]);
-
     // the tests expect `console.log()` to be interpreted as a newline
     // but the mock captures the info as `undefined`
-    return output === undefined ? '\n' : output;
+    const output = args[0] === undefined ? '\n' : args[0];
+
+    // the validator only ever uses the first arg in consolg.log
+    return preserveColors ? output : stripAnsi(output);
```
dpopp07, line 21: I reversed the order of this because the `stripAnsi` function will throw an error if it receives an `undefined` value.

packages/validator/src/cli-validator/run-validator.js
dpopp07: Fixed!

PR #706, opened 2024-12-19
feat(ibm-use-date-based-format): introduce new validation rule

dpopp07:
## PR summary

This commit introduces the new 'ibm-use-date-based-format' rule,
which will heuristically verify that schemas, with either a name
or an example value indicating a date-based logical type, be
strings and use either "date" or "date-time" as the format.

### General checklist
- [x] `.secrets.baseline` has been updated as needed
- [x] `npm run generate-utilities-docs` has been run if any files in `packages/utilities/src` have been updated

- [x] Added scoring rubric entry for new rule (packages/validator/src/scoring-tool/rubric.js)

### DIFF
--- a/packages/ruleset/src/functions/use-date-based-format.js
+++ b/packages/ruleset/src/functions/use-date-based-format.js
@@ -1,0 +1,312 @@
+/**
+ * Copyright 2024 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+const {
+  getExamplesForSchema,
+  isDateSchema,
+  isDateTimeSchema,
+  isIntegerSchema,
+  isObject,
+  isStringSchema,
+  validateNestedSchemas,
+} = require('@ibm-cloud/openapi-ruleset-utilities');
+
+const {
+  LoggerFactory,
+  isDateBasedName,
+  isDateBasedValue,
+  isParamContentSchema,
+  isParamSchema,
+  isRequestBodySchema,
+  isResponseSchema,
+  isSchemaProperty,
+} = require('../utils');
+
+let ruleId;
+let logger;
+
+/**
+ * The implementation for this rule makes assumptions that are dependent on the
+ * presence of the following other rules:
+ *
+ * - oas3-valid-media-example
+ * - oas3-valid-schema-example
+ *
+ * These rules verify that the correct, specific format (date vs date-time) is
+ * used for schemas based on their example value. So, we aren't as specific
+ * with that check in this rule - we recommend either "date" or "date-time".
+ */
+
+module.exports = function (schema, _opts, context) {
+  if (!logger) {
+    ruleId = context.rule.name;
+    logger = LoggerFactory.getInstance().getLogger(ruleId);
+  }
+
+  return checkForDateBasedFormat(
+    schema,
+    context.path,
+    context.documentInventory.resolved
+  );
+};
+
+/**
+ * This function implements a rule that enforces date-based schemas use either
+ * the "date" or "date-time" format, so that they're accurately documented as
+ * date-based logical types. We use a heuristic based on either the name of a
+ * schema (derived from the property name, if the schema is a property schema)
+ * or the example value provided for a given schema or schema property.
+ *
+ * The logic here recursively checks all schemas for the presence of unmarked
+ * date-based schemas. As it traverses the schemas, it compiles a list of
+ * potentially-relevant example values. This way, if an object schema defines
+ * its own example, which includes a value for a nested property that should
+ * be identified by the rule, we can track down the value once we reach the
+ * schema for said property. The logic will also gather any relevant parameter
+ * or media type examples that may be defined outside of the schema path.
+ *
+ * @param {object} s the schema to check
+ * @param {array} p the array of path segments indicating the "location" of the schema within the API definition
+ * @param {object} apidef the resolved API definition
+ * @returns an array containing the violations found or [] if no violations
+ */
+function checkForDateBasedFormat(s, p, apidef) {
+  // Map connecting a list of examples for a schema to its logical path.
+  const examples = {};
+
+  // Check for any examples outside of the schema path - they may be in
+  // request bodies, response bodies, or parameters. Store these separately.
+  const indirectExamples = checkForIndirectExamples(p, apidef);
+
+  return validateNestedSchemas(s, p, (schema, path, logicalPath) => {
+    logger.debug(`${ruleId}: checking schema at location: ${path.join('.')}`);
+    logger.debug(
+      `${ruleId}: logical schema path is : ${logicalPathForLogger(logicalPath)}`
+    );
+
+    // Use a composition-aware utility to gather any examples relevant to this
+    // schema, including those defined on applicator schemas in oneOf, etc.
+    const schemaExamples = getExamplesForSchema(schema);
+    logger.debug(`${ruleId}: ${schemaExamples.length} examples found`);
+
+    // Examples have already been stored on the parent - as we go through the
+    // schemas, check for the presence of an example value for the current
+    // property within the parent's example.
+    const parentalExamples =
+      // If the logical path is empty, there are no parents, but there may be
+      // indirect examples to add in the ":" branch.
+      logicalPath.length > 0
+        ? // Look at the example values for the logical parent schema, if any.
+          // For successive, nested properties, this will end up propagating
+          // examples through the recursive descent so that example values
+          // separated from a property by multiple degrees of nesting will
+          // still be preserved.
+          examples[logicalPath.slice(0, -1).join('.')]
+            .map(e => {
+              // Check the parental example values for
+              // the presence of the current property.
+              const prop = logicalPath.at(-1);
+
+              // Check for sentinel indicating an array.
+              if (prop === '[]' && Array.isArray(e)) {
+                return e;
+              }
+
+              // Check for sentinel indicating a dictionary.
+              if (prop === '*' && isObject(e)) {
+                return Object.values(e);
+              }
+
+              // Standard model path. Wrap value in an array to match list and
+              // dictionary behavior - it will be flattened out later.
+              return [e[prop]];
+            })
+
+            // Each example may map to multiple examples - flatten the result
+            // of the mapping to include all relevant examples in the list.
+            .flat()
+
+            // We are not guaranteed to find a value - filter
+            // out any values that are not defined.
+            .filter(e => e !== undefined)
+        : // Add indirect examples to the map - note that they will necessarily
+          // be indexed with the empty string key, like primary schemas.
+          indirectExamples;
+
+    logger.debug(
+      `${ruleId}: ${parentalExamples.length} examples found in logical parent`
+    );
+
+    // Index the examples with the stringified logical path.
+    // Note that the unconditional assignment is intentional - there may be
+    // existing entries for this same logical path (e.g. for the same
+    // nested property within a different oneOf sibling) but we want to
+    // override them, always. Otherwise, the behavior would depend on the
+    // order the schemas are checked in (the logic may look at more examples
+    // for one instance of a property than another, arbitrarily) and we don't
+    // make a guarantee that order will be stable in `validateNestedSchemas`.
+    examples[logicalPath.join('.')] = [...schemaExamples, ...parentalExamples];
+
+    // Perform the validation using the first value example value found for the
+    // schema at this logical path.
+    const exampleValue = examples[logicalPath.join('.')].find(
+      e => e !== undefined
+    );
+
+    return performValidation(schema, path, exampleValue);
+  });
+}
+
+// This function performs the actual checks against a schema to determine if
+// it should be a "date" or "date-time" schema, but isn't defined as one.
+// It is wrapped in the outer function for the gathering of examples, etc. but
+// this function implements the checks: 1) see if the name of a property
+// indicates that it is a date-based schema and 2) see if the example value for
+// a schema indicates that it is a date-based schema.
+function performValidation(schema, path, exampleValue) {
+  // If this is already a date or date-time schema, no need to check if it should be.
+  if (isDateSchema(schema) || isDateTimeSchema(schema)) {
+    logger.debug(
+      `${ruleId}: skipping date-based schema at location: ${path.join('.')}`
+    );
+
+    return [];
+  }
+
+  // Check if this is a schema property
+  if (isSchemaProperty(path)) {
+    logger.debug(`${ruleId}: detected named property at "${path.join('.')}"`);
+
+    // Check for a name that would indicate the property should be date-based
+    if (isDateBasedName(path.at(-1))) {
+      logger.debug(
+        `${ruleId}: property name at "${path.join('.')}" is date-based`
+      );
+
+      // We only assume a property could be a date-time value if it's a string or integer
+      if (isStringSchema(schema) || isIntegerSchema(schema)) {
+        logger.debug(
+          `${ruleId}: date-based property name at "${path.join(
+            '.'
+          )}" is a string or integer`
+        );
+
+        // If the schema is determined to be a date-time schema by the name alone,
+        // we can return - no need to look for an example value.
+        return [
+          {
+            message:
+              'According to its name, this property should use type "string" and format "date" or "date-time"',
+            path,
+          },
+        ];
+      }
+    }
+  }
+
+  // Check example values for string schemas.
+  if (isStringSchema(schema)) {
+    if (exampleValue !== undefined) {
+      logger.debug(`${ruleId}: example value found: ${exampleValue}`);
+
+      if (isDateBasedValue(exampleValue)) {
+        return [
+          {
+            message:
+              'According to its example value, this schema should use type "string" and format "date" or "date-time"',
+            path,
+          },
+        ];
+      }
+    } else {
+      logger.debug(`${ruleId}: no example value found`);
+    }
+  }
+
+  return [];
+}
+
+// This function takes an object, as well as a path to a specific value, and
+// parses the object, looking for the value at that path. If it finds one,
+// the value will be returned. If not, the function will return `undefined`.
+function getObjectValueAtPath(obj, pathToValue) {
+  return pathToValue.reduce((value, field) => value?.[field], obj);
+}
+
+// "Indirect" examples are those coming from request bodies, response bodies, and parameters.
+function checkForIndirectExamples(path, apidef) {
+  logger.debug(
+    `${ruleId}: checking indirect examples for schema at location: ${path.join(
+      '.'
+    )}`
+  );
+
+  // Parameter and Media Type objects have the same format when it comes
+  // to examples, so we can treat all of these scenarios the same way.
+  if (
+    isRequestBodySchema(path) ||
+    isResponseSchema(path) ||
+    isParamSchema(path) ||
+    isParamContentSchema(path)
+  ) {
+    // Example fields would be siblings of the schema we're looking at, so we need to look in the API
+    // for the path, minus the last value (which is "schema").
+    const examples = getOpenApiExamples(
+      getObjectValueAtPath(apidef, path.slice(0, -1))
+    );
+
+    // Check for the special case of looking at a content schema for a parameter that
+    // itself defines an example (pull the last three values off the path to check).
+    if (isParamContentSchema(path)) {
+      examples.push(
+        ...getOpenApiExamples(getObjectValueAtPath(apidef, path.slice(0, -3)))
+      );
+    }
+
+    logger.debug(
+      `${ruleId}: ${
+        examples.length
+      } indirect examples found for schema at location: ${path.join('.')}`
+    );
+
+    // Put the examples in the format the downstream algorithm for this rule needs.
+    return examples;
+  }
+
+  return [];
+}
+
+// OpenAPI defines its own example structure, separate from schema examples,
+// on Parameter and Media Type objects. Use this function to parse those
+// structures and return any relevant examples. The argument may be either a
+// Parameter or Media Type object and will return a list.
+function getOpenApiExamples(artifact) {
+  if (!isObject(artifact)) {
+    return [];
+  }
+
+  // The `example` and `examples` fields are mutually exclusive.
+  if (artifact.example) {
+    return [artifact.example];
+  }
+
+  // This will be a map, potentially containing multiple examples. Return all of them.
+  if (artifact.examples) {
+    return Object.values(artifact.examples).map(
+      exampleObject => exampleObject.value
+    );
+  }
+
+  return [];
+}
+
+// Format the logical path in a way that makes sense when the array is empty.
+function logicalPathForLogger(logicalPath) {
+  if (!logicalPath.length) {
+    return `'' (primary schema)`;
+  }
+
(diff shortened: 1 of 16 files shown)


dpopp07:
Whoops, forgot to update the docs - I will push that change soon! Everything else is ready to review

packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -0,0 +1,106 @@
...
+    new RegExp('.*timestamp.*'),
+  ];
+
+  if (dateBasedNamePatterns.some(regex => regex.test(name))) {
+    return true;
+  }
+
+  // Check for direct equality with certain terms.
+  const dateBasedNames = [
+    'created',
```
padamstx, line 52: You could just fold these into the array of regexp's above (e.g. `new RegExp('^created$')` with proper anchoring to ensure the direct equality match.  That way, you have just one set of regex's to check.

packages/ruleset/test/rules/use-date-based-format.test.js
```diff
@@ -0,0 +1,4082 @@
...
+            expect(results[i].code).toBe(ruleId);
+            expect(results[i].message).toBe(expectedExampleMsg);
+            expect(results[i].severity).toBe(expectedSeverity);
+            expect(results[i].path.join('.')).toBe(expectedPaths[i]);
+          }
+        });
+      });
+    });
+  });
+});
```
padamstx, line 4082: Wow, this is a long testcase! :)

padamstx (commented):
PR looks great!
I just had one small suggestion for `date-based-utils.js`.

packages/ruleset/src/utils/date-based-utils.js
dpopp07: Good idea! 👍

dpopp07: Fixed

padamstx (approved):
LGTM!

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,359 @@
...
+ * @param {object} s the schema to check
+ * @param {array} p the array of path segments indicating the "location" of the schema within the API definition
+ * @param {object} apidef the resolved API definition
+ * @returns an array containing the violations found or [] if no violations
+ */
+function checkForDateBasedFormat(s, p, apidef) {
+  const examples = [];
+  const propertyPath = [];
+
+  return validateSubschemas(s, p, (schema, path) => {
```
hudlow, line 81: I probably should have made this clearer in the docs for `validateSubschemas()` but it _usually_ isn't what you want. What I said was this:

> This function is useful when a certain **syntactic practice** is prescribed or proscribed within **every simple schema** composed and/or nested by a schema.

Compare to the docs for `validateNestedSchemas()`:

> This function is useful when a certain **semantic quality** is required for every **composite schema** nested by a schema. The `validate` function passed to this function must itself be "composition-aware." Usually this means it will use other utilities (such as `getSchemaType()` or `schemaHasConstraint()`) to determine the semantic qualities of a composite schema.

What am I getting at here? Well, because `validateSubschemas()` treats all referenced schemas (whether they are schemas nested in properties or composed by applicators) the same, it's difficult and confusing to relate your validation logic to an outside context because the context changes between calls of the `validate` function in _irregular_ ways. We're not traversing either the logical property tree _or_ the logical composition tree. We're traversing both simultaneously!

For this reason, it really is best to limit the use of `validateSubschemas()` to validation logic whether the context _simply doesn't matter_: or, as stated in the docs, where you only care about validating a syntactic practice within a "simple schema" and you don't care about how it logically nests or composes into anything else.

This descent confusion is the cause of the bug noted in my three-part comment.

What you need to do here instead is use `validateNestedSchemas()`. You could start by calling `getPropertyNamesForSchema()` and checking the property names. Then things get a little trickier.

I would recommend factoring out the core logic from `getPropertyNamesForSchema()` into `collectFromCompositeSchema(schema, collector: (composedSchema) => collectedItems`. Then `getPropertyNamesForSchema(schema)` can be simplified into `collectFromCompositeSchema(schema, isObject(schema.properties) ? Object.keys(schema.properties) : []`. Then you can add `getExamplesForSchema(schema)` also using `collectFromCompositeSchema()`.

You can still index examples as you go so that nested schemas can refer to examples in their parents, but matching logical paths to schema paths is going to be tricky. I think I would recommend updating `validateNestedSchemas()` to track a `logicalPath` in addition to a `path`. It should be fairly straightforward, though you'll have to decide whether to break compatibility by inserting it into the middle of the current parameter list.

One subtle aspect of what I'm proposing here is that because we'd be flattening examples across logical paths, a specific example value might not apply to a particular instance of a schema — but given this is just a heuristic and a given logical path certainly ought not to have a mixture of fundamental types (\*cough\*), I think it should work fine in this case.

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,359 @@
...
+      getObjectValueAtPath(apidef, path.slice(0, -1))
+    );
+
+    // Check for the special case of looking at a content schema for a parameter that
+    // itself defines an example (pull the last three values off the path to check).
+    if (isParamContentSchema(path)) {
+      examples.push(
+        ...getOpenApiExamples(getObjectValueAtPath(apidef, path.slice(0, -3)))
+      );
+    }
```
hudlow, line 307: I don't think you need to do this way down here — these sorts of examples only exist at the top level where the rule is called for each schema-usage location, right?

packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -0,0 +1,101 @@
...
+ * time property. It returns `true` if the name matches one of the expressions.
+ *
+ * @param {string} name the name of a schema property
+ * @returns a boolean value indicating that the property seems to be date-based
+ */
+function isDateBasedName(name) {
+  // Check for matching against certain patterns.
+  const dateBasedNamePatterns = [
+    // The name `created`.
+    new RegExp('^created$'),
```
hudlow, line 18: FWIW, I prefer native RegExp literal synax.

```suggestion
    /^created$/,
```

packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -0,0 +1,101 @@
...
+ * @returns a boolean value indicating that the value seems to be date-based
+ */
+function isDateBasedValue(value) {
+  const regularExpressions = [
+    // Includes abbreviated month name.
+    new RegExp('\\b(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\\b'),
+
+    // Includes full month name.
+    new RegExp(
+      '\\b(January|February|March|April|May|June|July|August|September|October|November|December)\\b'
```
hudlow, line 79: With native RegExp literal syntax, you don't have to have double-`\`.

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,359 @@
...
+      }
+    }
+
+    // Use a slice (a copy) of the `propertyPath` array so that the
+    // invoked function can modify it without modifying the original.
+    return performValidation(
+      schema,
+      path,
+      apidef,
+      propertyPath.slice(),
```
hudlow, line 133: Personally, I think it is generally a bad practice to modify a parameter, so this should be done within the function. I also slightly prefer the clarity of `[... propertyPath]`

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,359 @@
...
+  // Check example values for string schemas.
+  if (isStringSchema(schema)) {
+    // If this is a property, we need to include its name in the path.
+    if (isSchemaProperty(path)) {
+      propertyPath.push(path.at(-1));
+    }
+
+    // Either use the schema example directly or search the list of examples
+    // for an example object that contains a value for this property.
+    const exampleValue = schema.example || findExample(propertyPath, examples);
```
hudlow, line 184: This isn't composition-aware, so it will miss examples in schemas like this:
```
{
  "type": "string",
  "allOf": [
    {
      "example": "2024-12-19"
    }
  ]
}
```

hudlow (changes requested):

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: That's right - this function is the first thing that's called when we look at a schema. If it's the top level, a schema in a schema-usage location, this will end up populating the list with any relevant examples. Other times, we'll call it but it will end up being skipped. I'm not sure I understand what you mean by "way down here" 🙂 is there another place we should check for these examples?

packages/ruleset/test/utils/date-based-utils.test.js
```diff
@@ -0,0 +1,87 @@
...
+      expect(isDateBasedName('expires')).toBe(true);
+      expect(isDateBasedName('start_time')).toBe(true);
+      expect(isDateBasedName('start_date')).toBe(true);
+      expect(isDateBasedName('end_time')).toBe(true);
+      expect(isDateBasedName('end_date')).toBe(true);
+      expect(isDateBasedName('create_time')).toBe(true);
+      expect(isDateBasedName('create_date')).toBe(true);
+      expect(isDateBasedName('created_time')).toBe(true);
+      expect(isDateBasedName('created_date')).toBe(true);
+      expect(isDateBasedName('modify_time')).toBe(true);
```
dpopp07, line 67: Something about this (and other things like `running_time` for a movie, or `hang_time` for a football) feels like it could conceivably be a duration, which makes me reluctant to flag these for integer types. That's why [I currently don't check](https://github.com/IBM/openapi-validator/blob/dp/date-time-rule/packages/ruleset/src/utils/date-based-utils.js#L53-L55) for these. What do you think @hudlow ?

packages/ruleset/src/functions/use-date-based-format.js
hudlow: My point was just that it's unnecessary and a bit wasteful to be checking if it's a request body, response, or parameter schema in the actual recursion.

dpopp07: Oh I see - yes, that can be optimized. So, this feedback is more on the location of the _invocation_ of this function - I'll update that.

packages/ruleset/src/utils/date-based-utils.js
dpopp07: I do, too - I initially made this change for a linter problem that seems to have been caused by something else (because I can't reproduce it). So, I'll go back to this 👍

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: Good call - will update 👍

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,401 @@
...
+ * @param {object} apidef the resolved API definition
+ * @returns an array containing the violations found or [] if no violations
+ */
+function checkForDateBasedFormat(s, p, apidef) {
+  // Map connecting a list of examples for a schema to its logical path.
+  const examples = {};
+
+  // Check for any examples outside of the schema path - they may be in
+  // request bodies, response bodies, or parameters. Store these separately.
+  const indirectExamples = checkForIndirectExamples(p, apidef);
```
hudlow, line 81: Personally, I think it's better to always prioritize examples closer to the schema definition because then the example cited in the rule output will be easier to find when looking at the schema it pertains to.

With that being said, I think `indirectExamples` should probably simply be prepended to `examples` for primary schemas.

packages/utilities/src/utils/validate-subschemas.js
```diff
@@ -27,11 +31,26 @@ const validateNestedSchemas = require('./validate-nested-schemas');
...
+ * - `subschema`: the composed or nested schema
+ * - `path`: the array of path segments to locate the subschema within the resolved document
+ * - `logicalPath`: the array of path segments to locate an instance of `subschema` within an
+ *    instance of `schema` (the schema for which `validateSubschemas()` was originally called.)
+ *    Within the logical path, an arbitrary array item is represented by `[]` and an arbitrary
+ *    dictionary property is represented by `*`.
+ *
+ * The provided `validate()` function is guaranteed to be called:
+ * - for a schema before any of its composed schemas
+ * - for a schema before any of its nested schemas
```
hudlow, line 44: We made need something like this since we may be depending on this behavior.

```suggestion
 * - for a schema before any of its nested schemas
 * - more recently for a schema's parent in the nesting or composition tree than for any of its parents siblings in those trees
```

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: > With that being said, I think indirectExamples should probably simply be prepended to examples for primary schemas.

This is how I implemented it initially when updating the logic. However, I adjusted it for two reasons.

1. The OpenAPI specification states that 

> if referencing a schema which contains an example, the example(s) value SHALL override the example provided by the schema.

so I thought this logic would keep things closer to the specification (not that it really matters for this heuristic).

2. Separating it out made it much easier to log when an example value was found on an indirect example vs a primary schema example.

> Personally, I think it's better to always prioritize examples closer to the schema definition because then the example cited in the rule output will be easier to find when looking at the schema it pertains to.

I do agree that it makes it easier for the user. Let me know what you think about the above context.

packages/utilities/src/utils/validate-subschemas.js
dpopp07: That's a change to make in #709, right? That's one more guarantee that you can add a test for (I believe it should be accurate).

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,401 @@
...
+    );
+
+    // Use a composition-aware utility to gather any examples relevant to this
+    // schema, including those defined on applicator schemas in oneOf, etc.
+    const schemaExamples = getExamplesForSchema(schema);
+    logger.debug(`${ruleId}: ${schemaExamples.length} examples found`);
+
+    if (schemaExamples.length) {
+      // Index the examples with the stringified logical path.
+      examples[logicalPath.join('.')] = schemaExamples;
```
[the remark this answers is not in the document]
dpopp07: Agreed, good thought!

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,401 @@
...
+    logger.debug(
+      `${ruleId}: logical schema path is : ${logicalPathForLogger(logicalPath)}`
+    );
+
+    // Use a composition-aware utility to gather any examples relevant to this
+    // schema, including those defined on applicator schemas in oneOf, etc.
+    const schemaExamples = getExamplesForSchema(schema);
+    logger.debug(`${ruleId}: ${schemaExamples.length} examples found`);
+
+    if (schemaExamples.length) {
```
hudlow, line 94: I'd get rid of this check — what's wrong with an empty array being in your map?

packages/ruleset/src/functions/use-date-based-format.js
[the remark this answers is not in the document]
dpopp07: Done

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: Good catch - at one point, I was intending this to be an optimization but it isn't one. This check is performed in `findValueInExamples` anyways, so this is redundant. Removing...

packages/ruleset/src/functions/use-date-based-format.js
hudlow, line 103: Speaking of prepending indirect examples, I think if prepend something like this...

```js
logicalPath.length > 0 ? examples[logicalPath.slice(0, -1).join('.')].map(e => e[logicalPath.at(-1)]).filter(e => e !== undefined) : indirectExamples
```

... then you can probably get rid of most of the logic in `findExample()`, `findValueInExamples()`, and `getObjectValueAtPath()`. I.e., if we eagerly gather available example values from parent examples, we don't have to do so much back-solving to find them later.

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,401 @@
...
+
+  // Look at each level of hierarchy of the example path, INCLUDING examples on
+  // primary schemas, which will have an example path of an empty array. For
+  // that reason, we can't loop through the example path elements, as we would
+  // miss the iteration with no elements in the array. Using a simple for-loop
+  // that executes one more time than there are segments in the example path,
+  // we execute this logic for the empty path scenario as well, and thereby
+  // properly handle primary schemas. The length of the example path is used to
+  // initialize the counter, rather than in the conditional, since the array
+  // will change length as we progress.
```
hudlow, line 232: Per the comment above, you probably don't need this at all, but as a matter of edification, I think `do-while` is much clearer here:

```
do {
  ...
} while (examplePath.pop() !== undefined);
```

packages/utilities/src/utils/validate-subschemas.js
hudlow: Yes.

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: Oh, 100%. I feel silly haha - I had a `do-while` loop initially but ran into the issue of executing the code for the empty array, and didn't even consider popping the array item within the conditional itself! I knew there was a better way to do this 😅

packages/ruleset/src/functions/use-date-based-format.js
hudlow: (Well, you still need to handle `[]` and `*` sentinels, obviously.)

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  return undefined;
+}
+
+// This function takes an object, as well as a path to a specific value, and
+// recursively parses the object looking for the value at that path. If it
+// finds one, the value will be returned. If not, the function will return
+// `undefined`.
+function getObjectValueAtPath(obj, pathToValue) {
```
hudlow, line 299: Recursion is probably overkill here — a loop should be fine.

dpopp07: I'm surprised you didn't suggest `reduce` here 🙂 

I felt the recursion was cleaner than when I had the implementation as a loop, but I'll consider adjusting it.

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: This ended up working well! I added this (extending it with comments for readability) and was able to remove a lot of code, including both `findExample()` and `findValueInExamples()`. Note that I needed to keep `getObjectValueAtPath`, as it's used to parse the API definition for indirect examples, but I was able to simplify it to one line now that I don't need to support the sentinel values in it.

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: `reduce` ended up working best, fixed

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: The one drawback to this is losing specificity in the logs presented to the user but I don't think that's a big deal in this case.

packages/ruleset/test/utils/date-based-utils.test.js
hudlow: I'm fine with not flagging everything with a `_time` suffix, but I think it would be good to flag it with the common prefixes that indicate date-time values.

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: Agreed, thanks for the further justification. I've re-implemented it this way.

packages/ruleset/test/utils/date-based-utils.test.js
dpopp07: Good call, I'll add some more literal values to the function.

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,305 @@
...
+    // Note that the unconditional assignment is intentional - there may be
+    // existing entries for this same logical path (e.g. for the same
+    // nested property within a different oneOf sibling) but we want to
+    // override them, always. Otherwise, the behavior would depend on the
+    // order the schemas are checked in (the logic may look at more examples
+    // for one instance of a property than another, arbitrarily) and we don't
+    // make a guarantee that order will be stable in `validateNestedSchemas`.
+    examples[logicalPath.join('.')] = [...schemaExamples, ...parentalExamples];
+
+    return performValidation(schema, path, logicalPath, examples);
```
hudlow, line 147: No biggie, but there's technically no need to pass `logicalPath` and the full `examples` map in here, right? Because we're pre-collecting the parental examples, `performValidation()` is only ever going to use `examples[logicalPath.join('.')]`.

Sorry, this popped into my head over the weekend and then I forgot about it.

hudlow (approved):

packages/ruleset/src/functions/use-date-based-format.js
dpopp07: Oh good catch - that's right! I'll refactor.

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,305 @@
...
+                return Object.values(e)[0];
+              }
+
+              // Standard model path.
+              return e[prop];
+            })
+
+            // We are not guaranteed to find a value - filter
+            // out any values that are not defined.
+            .filter(e => e !== undefined)
```
hudlow, line 128:
```suggestion
          examples[logicalPath.slice(0, -1).join('.')]
            .map(e => {
              // Check the parental example values for
              // the presence of the current property.
              const prop = logicalPath.at(-1);

              // Check for sentinel indicating an array.
              if (prop === '[]' && Array.isArray(e)) {
                return e;
              }

              // Check for sentinel indicating a dictionary.
              if (prop === '*' && isObject(e)) {
                return Object.values(e);
              }

              // Standard model path.
              return [e[prop]];
            })
            
            // each example may map to multiple examples, so
            // flatten the result of the mapping.
            .flat()

            // We are not guaranteed to find a value - filter
            // out any values that are not defined.
            .filter(e => e !== undefined)
```

packages/ruleset/src/functions/use-date-based-format.js
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  return [];
+}
+
+// This function takes an object, as well as a path to a specific value, and
+// parses the object, looking for the value at that path. If it finds one,
+// the value will be returned. If not, the function will return `undefined`.
+function getObjectValueAtPath(obj, pathToValue) {
+  return pathToValue.reduce((value, field) => value[field], obj);
```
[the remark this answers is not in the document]
dpopp07: Agreed - I always forget about optional chaining in JS! There are probably a lot of places in the validator that would benefit from it, at least syntactically. I'll make this change before merging.

PR #708, opened 2024-12-20
feat: add `collectFromComposedSchemas()` and `getExamplesForSchema()` to utilities

hudlow:
## PR summary
Adds a generic `collectFromComposedSchemas()` function that can be used to build functions like the also-added `getExamplesForSchema()`, and refactors `getPropertyNamesForSchema()` to use `collectFromComposedSchemas()` as well.

### General checklist
- [x] `.secrets.baseline` has been updated as needed
- [x] `npm run generate-utilities-docs` has been run if any files in `packages/utilities/src` have been updated

### DIFF
--- a/packages/utilities/src/utils/collect-from-composed-schemas.js
+++ b/packages/utilities/src/utils/collect-from-composed-schemas.js
@@ -1,13 +1,13 @@
 /**
- * Copyright 2017 - 2023 IBM Corporation.
+ * Copyright 2024 IBM Corporation.
  * SPDX-License-Identifier: Apache2.0
  */
 
 /**
  * Returns an array of items collected by the provided `collector(schema) => item[]` function for a
- * simple or composite schema, and deduplicate primitives in the result. The collector function is
+ * simple or composite schema, and deduplicates primitives in the result. The collector function is
  * not run for `null` or `undefined` schemas.
- * @param {object} schema simple or composite OpenAPI 3.0 schema object
+ * @param {object} schema simple or composite OpenAPI 3.x schema object
  * @param {Function} collector a `(schema) => item[]` function to collect items from each simple schema
  * @param {boolean} includeSelf collect from the provided schema in addition to its composed schemas (defaults to `true`)
  * @param {boolean} includeNot collect from schemas composed with `not` (defaults to `false`)
--- a/packages/utilities/src/utils/collect-from-composed-schemas.js
+++ b/packages/utilities/src/utils/collect-from-composed-schemas.js
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2017 - 2023 IBM Corporation.
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+/**
+ * Returns an array of items collected by the provided `collector(schema) => item[]` function for a
+ * simple or composite schema, and deduplicate primitives in the result. The collector function is
+ * not run for `null` or `undefined` schemas.
+ * @param {object} schema simple or composite OpenAPI 3.0 schema object
+ * @param {Function} collector a `(schema) => item[]` function to collect items from each simple schema
+ * @param {boolean} includeSelf collect from the provided schema in addition to its composed schemas (defaults to `true`)
+ * @param {boolean} includeNot collect from schemas composed with `not` (defaults to `false`)
+ * @returns {Array} collected items
+ */
+function collectFromComposedSchemas(
+  schema,
+  collector,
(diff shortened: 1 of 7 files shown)


packages/utilities/src/utils/collect-from-composed-schemas.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright 2017 - 2023 IBM Corporation.
```
dpopp07, line 2:
```suggestion
 * Copyright 2024 IBM Corporation.
```

(Not the only instance of this)

packages/utilities/src/utils/collect-from-composed-schemas.js
```diff
@@ -0,0 +1,56 @@
...
+ * SPDX-License-Identifier: Apache2.0
+ */
+
+/**
+ * Returns an array of items collected by the provided `collector(schema) => item[]` function for a
+ * simple or composite schema, and deduplicate primitives in the result. The collector function is
```
dpopp07, line 8:
```suggestion
 * simple or composite schema, and deduplicates primitives in the result. The collector function is
```

packages/utilities/src/utils/get-examples-for-schema.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Returns an array of examples for a simple or composite schema. For each composed schema, if
+ * `schema.examples` is present (and an array), `schema.example` is ignored.
+ * @param {object} schema simple or composite OpenAPI 3.0 schema object
+ * @returns {Array} examples
+ */
+function getExamplesForSchema(schema) {
+  return collectFromComposedSchemas(schema, s => {
+    if (Array.isArray(s.examples)) {
```
dpopp07, line 19: I'm under the impression that `examples` is only supported on Parameter and Media Type objects, not on schemas generally, and that it's a map. I thought only `example` was supported on schemas - does the spec allow for a list of examples as well?

packages/utilities/test/get-examples-for-schema.test.js
```diff
@@ -0,0 +1,154 @@
...
+  it('should return `[]` for empty object', async () => {
+    expect(getExamplesForSchema({})).toEqual([]);
+  });
+
+  it('should return examples for simple schema', async () => {
+    expect(
+      getExamplesForSchema({
+        examples: ['three', 'two', 'one'],
+      }).sort()
+    ).toEqual(['three', 'two', 'one'].sort());
```
dpopp07, line 26: I wouldn't think that `sort()` would be necessary for a list of primitives - am I missing something?

packages/utilities/test/get-examples-for-schema.test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('should return single example for simple schema', async () => {
+    expect(
+      getExamplesForSchema({
+        example: 'one',
+      }).sort()
```
dpopp07, line 33: Or for lists of size one 🙂

dpopp07 (commented):
First round - mostly just typos and questions

packages/utilities/src/utils/get-examples-for-schema.js
hudlow: OpenAPI 3.1 [deprecates `example`](https://spec.openapis.org/oas/v3.1.0#fixed-fields-19) in favor of ["native" JSON Schema `examples`](https://json-schema.org/draft/2020-12/json-schema-validation#name-examples).

packages/utilities/test/get-examples-for-schema.test.js
hudlow: Technically, there are no guarantees on the order examples are returned in.

packages/utilities/src/utils/get-examples-for-schema.js
dpopp07: Wow, I'm not sure how I got mixed up there - I thought `example` was the native field from JSON Schema. I probably need to update some things in my new rule based on that knowledge, as well. Thanks for linking that!

dpopp07 (approved):
Looks great! 👍 

The only other thing I wondered about is if we'd want to return a map with the examples, instead of a list, with the actual paths attached somehow - but that would require some rework of the logic and I'm not sure it's worth it. Rules utilizing this should already be able to know the logical path, which would be more useful. So, I think this is ready to merge as-is.

PR #710, opened 2025-01-07
build: migrate ci/cd to github actions

dpopp07:
Migrates the CI/CD build (including build, test, and publish steps) from Travis to GitHub Actions.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,102 @@
+# This workflow will build and unit test the project.
+# If the workflow is running on the "main" branch, then
+# semantic-release is also run to create a new release (if
+# warranted by the new commits being built).
+
+name: Build/Test/Publish
+
+on: 
+  push:
+    branches: ['**']
```
hudlow, line 10: Personally, I don't think I'd mind limiting this to `main` — doesn't seem like a huge burden to create a draft PR if you want a build to be triggered.

padamstx: Yeah, I don't see a problem with specifying `main` here either.  For the validator project, I think the PR and branch builds are the same anyway.  At the very least, github will probably let us know if something goes wrong :joy:

.github/workflows/build.yaml
```diff
@@ -0,0 +1,102 @@
...
+    # Only run this job on the main branch and only for our max version of Go.
+    if: "github.ref_name == 'main' && github.event_name != 'pull_request'"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: ['20']
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
padamstx, line 86:
```suggestion
        uses: actions/checkout@v4
        with:
          persist-credentials: false
```

padamstx (approved):
Looks good, except the small change I mentioned in the checkout step of the create-release job.  I'll approve now to avoid another review cycle.

.github/workflows/build.yaml
dpopp07: Done

.github/workflows/build.yaml
dpopp07: Done

PR #715, opened 2025-01-22
fix(ibm-api-symmetry): handle dictionaries and top-level schemas

### LINKED ISSUE #699: Unexpected ibm-api-symmetry warning with dictionary
I am getting validation warning `ibm-api-symmetry` with an openapi spec that includes a dictionary following https://swagger.io/docs/specification/v3_0/data-models/dictionaries/ :

```
$ lint-openapi error_example.yaml --log-level ibm-api-symmetry=info
IBM OpenAPI Validator (validator: 1.20.0), @Copyright IBM Corporation 2017, 2024.

Validation Results for error_example.yaml:

[INFO] ibm-api-symmetry: checking variant schema DevicePrototype against canonical schema Device
[INFO] ibm-api-symmetry: schema defines 'additionalProperties' - it is not a proper graph fragment
[INFO] ibm-api-symmetry: nested object property input_parameters is not a graph fragment of canonical property input_parameters
[INFO] ibm-api-symmetry: nested object property device_config is not a graph fragment of canonical property device_config
[INFO] ibm-api-symmetry: nested object property details is not a graph fragment of canonical property details
[INFO] ibm-api-symmetry: variant schema DevicePrototype is not a graph fragment of canonical schema Device
Warnings:

  Message :   Variant schema should be a graph fragment of the canonical schema
  Rule    :   ibm-api-symmetry
  Path    :   components.schemas.DevicePrototype
  Line    :   120

Summary:

  Total number of errors   : 0
  Total number of warnings : 1

  Warnings:
   1 (100%) : Variant schema should be a graph fragment of the canonical schema

```

I am using this version:  `validator: 1.20.0; ruleset: 1.19.0 (default)`

Here is the yaml file that generates the warning: [error_example.txt](https://github.com/user-attachments/files/17798808/error_example.txt)

Is this a legitimate warning?

I guess it is not because if I define `input_parameters` in `DeviceConfiguration` as a string instead of a dictionary, no warning is generated, so I guess it has to do with the dictionary definition. In addition, if I comment the path `/devices/{device_id}` (lines 14 to 36 in the file) no warning is generated either.

Is there anyway to skip the validation for this `input_parameters`?

Thanks for your help.

dpopp07:
This rule was subject to both false positives, as it flagged any schema with a
dictionary property (even when inappropriate), and false negatives, as it did
not check top-level schemas, only schema properties. This fix resolves both
issues and cleans up the rule logic on the whole.

Additionally, changes are made to the rule's logging. The code
to determine if one schema is a graph fragment of the other
uses a depth-first algorithm that prints a log, if it determines the
schema to violate the graph fragment pattern, with the reason behind
the violation. Due to the depth-first nature of the algorithm, the
logs are currently printed in depth first order, which is not as
coherent for the user to read.

This change introduces a stack to collect the logs during processing
and print them in reverse order afterwards to give the user a
better sense of what happened during the processing.

Resolves #699

### DIFF
--- a/packages/ruleset/src/functions/api-symmetry.js
+++ b/packages/ruleset/src/functions/api-symmetry.js
@@ -22,6 +22,10 @@ const {
 let ruleId;
 let logger;
 
+// The graph fragment check is depth-first. Use this stack to
+// print the relevant info logs in a sane order.
+const infoLogStack = [];
+
 /**
  * The implementation for this rule makes assumptions that are dependent on the
  * presence of the following other rules:
@@ -175,7 +179,7 @@ function checkForGraphFragmentPattern(
         schema => getSchemaType(variant) === getSchemaType(schema)
       )
     ) {
-      logger.info(
+      infoLogStack.push(
         `${ruleId}: variant and canonical schemas are different types`
       );
       result = false;
@@ -199,7 +203,7 @@ function checkForGraphFragmentPattern(
           )
       )
     ) {
-      logger.info(
+      infoLogStack.push(
         `${ruleId}: variant is array with schema that is not a graph fragment of canonical items schema`
       );
       result = false;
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM

PR #716, opened 2025-01-23
fix: ingore violations from invalid locations of an api definition

dpopp07:
## PR summary

Some extensions in an API definition, like those providing SDK-related metadata, are not valid places to check for rule violations.

Since some built-in, Spectral rules use broad JSON paths for their rules, it is possible for them to report false positives within extension objects, even when the extension objects contain information that is completely unrelated to the rule (like the typed-enum rule).

To avoid this in a generic way, this change filters out any rule violations that occur within our internal, SDK-related metadata extension and provides for any potential locations that may be added in the future.

### General checklist
- [ ] `.secrets.baseline` has been updated as needed
- [ ] `npm run generate-utilities-docs` has been run if any files in `packages/utilities/src` have been updated

### DIFF
--- a/packages/validator/src/spectral/index.js
+++ b/packages/validator/src/spectral/index.js
@@ -71,6 +71,12 @@ function convertResults(spectralResults, { config, logger }) {
       continue;
     }
 
+    // Ingore any rule violations that occur in sections of the spec that are
+    // not valid locations for API validation.
+    if (invalidLocation(r.path)) {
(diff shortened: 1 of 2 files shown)


dpopp07:
We may want to do something like this at some point, but right now the only rule tripped up by the metadata is Spectral's `typed-enum` rule, which we can patch.

Closing this in favor of #719

PR #717, opened 2025-01-23
fix(ibm-use-date-based-format): tighten heuristic for flagging date-time values

dpopp07:
## PR summary

The regular expressions used to determine if a value should be considered a date-time value are too loose and allow false positives to sneak in. It's a tricky problem, but this commit makes an attempt to tighten them by dismissing the possibility of a date-time value if any letters are found that are not relevant to a date-time value.

### General checklist
- [ ] `.secrets.baseline` has been updated as needed
- [ ] `npm run generate-utilities-docs` has been run if any files in `packages/utilities/src` have been updated

### DIFF
--- a/packages/ruleset/src/utils/date-based-utils.js
+++ b/packages/ruleset/src/utils/date-based-utils.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 IBM Corporation.
+ * Copyright 2024-2025 IBM Corporation.
  * SPDX-License-Identifier: Apache2.0
  */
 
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -77,12 +77,40 @@ function isDateBasedName(name) {
...
  */
 function isDateBasedValue(value) {
-  const regularExpressions = [
-    // Includes abbreviated month name.
-    /^\b(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\b/,
+  // The full and abbreviated values for months will be used in the generic
+  // string check below, as well as in the group of date-based expressions in
+  // the primary check below that.
+  const months =
+    /\b(Jan(uary)*|Feb(ruary)*|Mar(ch)*|Apr(il)*|May|Jun(e)*|Jul(y)*|Aug(ust)*|Sep(tember)*|Oct(ober)*|Nov(ember)*|Dec(ember)*)\b/g;
```
padamstx, line 84:
```suggestion
    /\b(Jan(uary)?|Feb(ruary)?|Mar(ch)?|Apr(il)?|May|Jun(e)?|Jul(y)?|Aug(ust)?|Sep(tember)?|Oct(ober)?|Nov(ember)?|Dec(ember)?)\b/g;
```
I think we need to use `?` instead of `*` so that we match zero or one instance of `uary` (for example). As it is, we'll match on `Januaryuaryuary`.

packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -77,12 +77,40 @@ function isDateBasedName(name) {
...
+  // In an effort to avoid false positives with strings that may contain
+  // date-time values, but are not themselves date-time values, check for
+  // the presence of any lowercase letters that are not included in the
+  // names of months or days.
+  if (typeof value === 'string') {
+    // "Day" strings are not enough on their own to determine date-time values,
+    // but we need to remove them from any values before we check for any
+    // letters that are not relevant to a date-time value.
+    const days =
+      /\b(Mon(day)*|Tue(sday)*|Wed(nesday)*|Thu(rsday)*|Fri(day)*|Sat(urday)*|Sun(day)*)\b/g;
```
padamstx, line 95: Same here... `*` --> `?`

padamstx (changes requested):
I think we need to use `?` instead of `*` in the new regex's

packages/ruleset/src/utils/date-based-utils.js
dpopp07: Great catch! 👍

dpopp07: Fixed and added test

padamstx (approved):
LGTM

PR #719, opened 2025-01-27
fix(typed-enum): modify spectral rule to look only at schemas

dpopp07:
This Spectral rule checks any object that has a type and enum field, regardless of whether it is in a valid schema location or not. This fix modifies the given paths to run the same functionality against schema objects only.

This will resolve a number of recently observed false positives.


padamstx (approved):
LGTM

PR #721, opened 2025-01-31
Purge "qualitative" for consistency

hudlow:
I apparently couldn't be consistent myself, so I think whatever semantic benefit I would argue for "qualitative", it cannot overcome the friction.

Untitled


dpopp07 (approved):
Looks good! 👍

PR #723, opened 2025-02-03
fix(ibm-use-date-based-format): avoid false positives for mac addresses

dpopp07:
This adjustment to the date-based value heuristic prevents a time from being detected if the word boundary is a colon character. This prevents colon-separated values that might otherwise look like dates, like MAC addresses, from being detected.

### DIFF
--- a/packages/ruleset/src/utils/date-based-utils.js
+++ b/packages/ruleset/src/utils/date-based-utils.js
@@ -122,7 +122,9 @@ function isDateBasedValue(value) {
     /\b(0?[1-9]|1[012])[./-]([012]?[1-9]|3[01])[./-]\d{4}\b/,
 
     // Includes time in the format (T)tt:tt:tt (where t can be s/m/h/etc.)
-    /(\b|T)\d\d:\d\d:\d\d\b/,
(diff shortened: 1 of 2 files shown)


packages/ruleset/src/utils/date-based-utils.js
```diff
@@ -122,7 +122,9 @@ function isDateBasedValue(value) {
     /\b(0?[1-9]|1[012])[./-]([012]?[1-9]|3[01])[./-]\d{4}\b/,
 
     // Includes time in the format (T)tt:tt:tt (where t can be s/m/h/etc.)
-    /(\b|T)\d\d:\d\d:\d\d\b/,
+    // This this case, don't consider a colon character a word break, to avoid
```
padamstx, line 125: Should this be "In this case"???

padamstx (approved):
LGTM, except I spotted a typo in a comment that probably makes sense to fix :).  I'll approve now to avoid a re-review cycle.

packages/ruleset/src/utils/date-based-utils.js
dpopp07: Yes, good catch 🙂

PR #724, opened 2025-02-04
fix(ibm-use-date-based-format): handle null example values

dpopp07:
Previously, if the example value given for a property was an explicit 'null', this rule would crash the validator.

This commit filters null examples from the logic so that they won't cause any trouble.


padamstx (approved):
LGTM

PR #727, opened 2025-02-07
fix(ibm-major-version-in-path): skip path checks if there are no paths

dpopp07:
This rule checks both the `servers` entries and the path strings in an API
definition. If an API definition contains no path strings in the `paths`
object, this rule would return a false positive. This adds logic to skip
the path string checks if there are none to check.

I also renamed a file to better match the rule name.


padamstx (approved):
LGTM

PR #730, opened 2025-02-12
build: fix exe build script

padamstx:
This commit fixes the create-binaries.sh script
by adding the "-t" (targets) option to explicitly
specify the node versions for which the executables should be built.
Because we haven't successfully published the executables in about 9 months (!), I also modified the semantic-release configuration so that we avoid the steps to build and publish the executables.  If we need to add this back later, it should be relatively easy since the create-binaries.sh script is now fixed.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,5 +1,5 @@
 {
-  "branches": "main",
+  "branches": ["main"],
   "debug": true,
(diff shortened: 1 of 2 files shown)


dpopp07 (approved):
Looks good! 👍

PR #751, opened 2025-06-04
fix(build): bump tar-fs to avoid CVE-2025-48387

padamstx:
I got lucky...  all I had to do is run `npm install` in my local sandbox to re-load dependencies, and the tar-fs dependency got bumped to the fixed version without needing to do anything else.


pyrooka (approved):
LGTM!

PR #760, opened 2025-08-27
fix: fix dependency issue with inflected

diatrcz:
Adds inflected dependency to the ruleset package as well, not just the validator.


pyrooka (approved):
LGTM!

PR #777, opened 2026-01-23
fix: bump lodash version to avoid CVEs

Andris28:
Bump lodash version from 4.17.21 to 4.17.23 to avoid CVEs.
ref: [vulnerability](https://github.com/advisories/GHSA-xxjr-mmjv-4gpg)


pyrooka (approved):
Looks good, thanks for the quick fix!

PR #778, opened 2026-01-26
fix: bump jsonpath-plus to avoid CVEs

Andris28:
Bump jsonpath-plus from 10.2.0 to 10.3.0 to avoid CVEs.


diatrcz (approved):
looks great to me! 😄

pyrooka (approved):
LGTM!

PR #807, opened 2026-05-07
fix(deps): lock dependencies

diatrcz:
Locks dependencies to avoid vulnerabilities.


pyrooka (approved):
Thanks for making this tedious work! Looks good!

PR #816, opened 2026-06-18
build: use dependency cooldown logic

Andris28:
In this commit we add dependency cooldown logic
that enables us to update more frequently and worry
less about compromised packages.


diatrcz (approved):
lgtm!

diatrcz (approved):
lgtm!

pyrooka (approved):
Looks good!

PR #827, opened 2026-07-03
build: add workflow for github pages

diatrcz:
Add github pages configuration through Github Actions.


_config.yml
pyrooka, line 1: I think this file should be in the project root as that's the source directory for Jekyll in the GH Action workflow.

pyrooka (commented):
Overall this looks pretty good, but the deployed pages still doesn't look 100% correct. I see some markdown elements rendered there: https://ibm.github.io/openapi-validator/docs/ibm-cloud-rules.html
Can we fix it somehow?

.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,51 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
```
pyrooka: This line should be removed I think. It's not just a sample anymore. :)

.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
```
pyrooka: Since `cancel-in-progress` is set to `true` this comment contradicts with the actual config, right?

docs/ibm-cloud-rules.md
```diff
@@ -779,8 +778,7 @@ npm install @ibm-cloud/openapi-ruleset
 If you would simply like to modify a rule's severity or disable a rule altogether,
 follow the instructions in this section.
 
-> [!CAUTION]
-> Disabling certain rules may lead to false negatives for problems identified by other rules.
+> **Caution:** Disabling certain rules may lead to false negatives for problems identified by other rules.
```
pyrooka: Visually the older version is better, but I guess it's necessary to change this for Jekyll, right?

pyrooka (changes requested):
Few comments to address, otherwise it looks good. I like the result a lot!

docs/ibm-cloud-rules.md
diatrcz: It was an issue in the beginning making the table under it break, but since then i fixed most of the broken html parsing in the markdown file. I can check how it renders now and change it back if it doesn't break it.

pyrooka (approved):
Looks good!

